PR #141, opened 2019-08-06
PickList Component

pr3tori4n:
**Related Issue:** #9

## Summary

This introduces a new Picker component. It's a type of menu where you have a list of selections to pick from. It supports multi-select behavior as an option. Each selection is represented by a `picker-row` and the rows can be sub-grouped in a picker using `picker-group`'s.

Currently, I'm doing sort of a POC with this component to see what the implementation would be like in using slotted content versus passing in JSON data. Currently, it's functioning with one, the other, or both. We can simplify it if we decide we don't need to support one of these variations.

NOTE: in a follow up PR, I'll be introducing a calcite-filter component, as well as additions to these components integrating that filter.

### Tests
Just remembered. Need to add tests still. ^_^


src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,262 @@
...
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  setupDragAndDrop(): void {
+    const sortGroups = [this.el, ...Array.from(this.el.querySelectorAll("calcite-picker-group"))];
+    sortGroups.forEach((sortGroup) => {
+      Sortable.create(sortGroup, {
+        group: "whateva",
```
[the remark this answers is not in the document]
pr3tori4n: cool - know where that snippet is? I'll re-use it.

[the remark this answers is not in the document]
driskull: they use `guid` here: https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tab/calcite-tab.tsx#L14

src/components/calcite-picker/calcite-picker-group/calcite-picker-group.tsx
```diff
@@ -0,0 +1,27 @@
+import { Component, Host, Prop, h } from "@stencil/core";
+
+@Component({
+  tag: "calcite-picker-group",
+  shadow: true
+})
+export class CalcitePickerGroup {
```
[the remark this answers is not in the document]
pr3tori4n: It seems the suggestion is each component should have its own directory. It doesn't comment on whether to have all of those directories on a single level versus nesting them. https://stenciljs.com/docs/style-guide#file-structure
I'm ok with moving these all to a common level, but I find nesting second-class components into their first-class cousins lends a bit of a lay-up for understanding which is the main component and which are supporting components.

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.scss
```diff
@@ -0,0 +1,31 @@
+:host {
+  display: flex;
+  align-items: center;
+  padding: 12px;
+  background-color: white;
+  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.75);
+  margin-bottom: 12px;
+  cursor: pointer;
+}
+:host([hidden]) {
```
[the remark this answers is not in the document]
pr3tori4n: Adding newlines

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,150 @@
...
+  shadow: true
+})
+export class CalcitePickerRow {
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Prop({ reflect: true }) textHeading: string;
```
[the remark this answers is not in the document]
pr3tori4n: jsdoc you mean? Or you mean add some notes in the README?

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,143 @@
...
+      return null;
+    }
+    if (this.icon === "grip") {
+      return (
+        <span class="handle">
+          <CalciteIcon size="24" path={handleVertical24} />
+        </span>
+      );
+    } else {
+      /* tslint:disable */
```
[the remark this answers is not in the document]
pr3tori4n: So as far as moving forward - do you guys agree with my judgement that this is more readable than the alternative? If so can we explore the tslint rule soonish? Otherwise I'll have to change it, then change it back.

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,150 @@
...
+        </span>
+      );
+    }
+  }
+
+  renderSecondaryAction() {
+    return this.editing ? (
+      <calcite-action
+        onClick={() => {
+          this.deleteRow();
```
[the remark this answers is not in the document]
pr3tori4n: how about using bind()? I'm not sure what the syntax would be for a public method done that way.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,262 @@
...
+  @State() editing = false;
+
+  @Watch("editing")
+  editingChangeHandler() {
+    this.slottedRows.forEach((item) => {
+      item.toggleAttribute("editing", this.editing);
+    });
+  }
+
+  deletedRows = new Set();
```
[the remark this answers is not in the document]
pr3tori4n: non-issue now that we're removing this funcitonality.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Element() el: HTMLElement;
+
+  sortable = null;
```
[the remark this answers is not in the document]
pr3tori4n: not being used, deleting.

src/components/calcite-picker/calcite-picker-group/calcite-picker-group.tsx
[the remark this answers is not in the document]
driskull: But in their example there aren't any nested components.

```
├── card
│   ├── card.ios.css
│   ├── card.md.css
│   ├── card.css
│   ├── card.tsx
│   └── test
│       └── basic
│           ├── e2e.js
│           └── index.html
├── card-content
│   ├── card-content.ios.css
│   ├── card-content.md.css
│   ├── card-content.css
│   └── card-content.tsx
├── card-title
│   ├── card-title.ios.css
│   ├── card-title.md.css
│   ├── card-title.css
```

I think it makes sense to keep them all at root since they are all `HTMLElement`s.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  @Prop({ reflect: true }) dragEnabled = false; /* ignored unless mode is configuration */
+
+  @Prop({ reflect: true }) editEnabled = false; /* ignored unless mode is configuration */
+
+  @Prop({ reflect: false }) data = [];
+
+  @State() selectedValues = new Set();
+
+  @Watch("selectedValues")
```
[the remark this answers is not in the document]
pr3tori4n: removed.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  getIconType() {
+    let type = null;
```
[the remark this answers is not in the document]
pr3tori4n: taken care of in the test PR.
https://github.com/Esri/calcite-app-components/pull/166/files#diff-79b6a9f0da30d1c1767967c29a737473R210

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+  @Watch("editing")
+  editingChangeHandler() {
+    this.slottedRows.forEach((item) => {
+      item.toggleAttribute("editing", this.editing);
+    });
+  }
+
+  deletedRows = new Set();
+  slottedRows: any;
```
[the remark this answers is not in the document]
pr3tori4n: removed.

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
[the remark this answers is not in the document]
pr3tori4n: Addressed in the Test PR.
https://github.com/Esri/calcite-app-components/pull/166/files#diff-a4da798c4e7b1713f6a12ef185306ea9L139

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,277 @@
...
+    this.deletedRows.forEach((row: HTMLCalcitePickerRowElement) => {
+      row.toggleAttribute("hidden", false);
+    });
+    this.deletedRows = new Set();
+    this.editing = false;
+  }
+
+  confirmDelete() {
+    let selectedChanged = false;
+    this.deletedRows.forEach((row: HTMLCalcitePickerRowElement) => {
```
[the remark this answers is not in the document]
pr3tori4n: removing in another PR.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,277 @@
...
+      if (this.selectedValues.has(row.value)) {
+        this.selectedValues.delete(row.value);
+        selectedChanged = true;
+      }
+      row.remove();
+    });
+    if (selectedChanged) {
+      this.pickerSelectionChange.emit(this.selectedValues);
+    }
+    this.deletedRows = new Set();
```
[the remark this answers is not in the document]
pr3tori4n: removing this block in another PR.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,277 @@
...
+    }
+    if (!this.multiple && selected) {
+      this.rows.forEach((item) => {
+        if (item !== row) {
+          this.deselectRow(item);
+        }
+      });
+    }
+    this.lastSelectedRow = row;
+    this.pickerSelectionChange.emit(this.selectedValues);
```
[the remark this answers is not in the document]
pr3tori4n: (calcite)pickerSelectionChange isn't an internal event. But I've updated the others.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,277 @@
...
+    this.editing = false;
+  }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Public Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method() async getSelectedRows() {
```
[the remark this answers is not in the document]
pr3tori4n: Right now, the only way to get the selected rows is to listen to a change event on the picker. I think being able to get the selected rows without requiring a trigger to change is a nice convenience addition. If you guys think it will cause more harm than good, we can remove it.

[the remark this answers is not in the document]
driskull: Can’t they just query the dom for rows?

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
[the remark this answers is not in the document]
jcfranco: JSDoc.

src/components/calcite-picker/calcite-picker.tsx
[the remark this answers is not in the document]
pr3tori4n: > Can’t they just query the dom for rows?

Not unless they want to have a code split for one type of data vs another. The component is retruning an object containing all selected rows keyed by value (might move to a map). querySelectorAll returns a nodeList.

I don't think it's a question of why it should be there. It's a question of - is there a good reason to remove it. I don't see any problems in keeping it.

pr3tori4n:
@jcfranco @driskull I've addressed your comments. Can you take a second look.

src/components/calcite-picker-row/calcite-picker-row.scss
```diff
@@ -0,0 +1,36 @@
...
+  margin-bottom: 12px;
+  cursor: pointer;
+}
+
+:host([hidden]) {
+  display: none;
+}
+
+:host(.highlight) {
+  border-left: 3px solid transparent;
```
driskull: I think we need RTL styles for this

src/components/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,152 @@
+import { Component, Element, Event, EventEmitter, Host, Method, Prop, h } from "@stencil/core";
```
driskull: was the PR to rename this file to `item` instead of `row` seperate?

src/components/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,152 @@
...
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Prop({ reflect: true }) editing = false;
+
+  @Prop({ reflect: true }) icon: "square" | "circle" | "grip" | null = null;
+
+  @Prop() metadata: object;
```
driskull: What if we had these as seperate properties like `tags`, `category`? Then we wouldn't have to accept an object from a Property

src/components/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  @Element() el: HTMLElement;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  @Event() calcitePickerRowToggled: EventEmitter;
```
driskull: should be `calcitePickerRowToggle` and `calcitePickerRowDelete`

src/components/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Public Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method() async toggle(shiftPressed: boolean) {
+    this.selected = !this.selected;
+    this.calcitePickerRowToggled.emit({
```
driskull: Would it be better to `@Watch selected`? because if a user sets the selected Property without the toggle method, the event wont be fired

src/components/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -0,0 +1,152 @@
...
+  iconClickHandler(e) {
+    this.toggle(e.shiftKey);
+  }
+
+  secondaryActionContainerClickHandler(e) {
+    e.stopPropagation();
+  }
+
+  deleteRow() {
+    this.calcitePickerRowDeleted.emit({ row: this.el, value: this.value });
```
driskull: I thought we were removing the delete stuff?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * When true, the rows will be sortable via drag and drop.
+   * Only applies when mode is configuration
+   */
+  @Prop({ reflect: true }) dragEnabled = false;
```
driskull: can we rename to `draggable`? for dom consistency

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  @Prop({ reflect: true }) editEnabled = false; /* ignored unless mode is configuration */
```
driskull: can we rename to `editable`? for dom consistency

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+  @Watch("editing")
+  editingChangeHandler() {
+    this.rows.forEach((item) => {
+      this.editing ? item.setAttribute("editing", "") : item.removeAttribute("editing");
+    });
+  }
+
+  deletedRows = new Set();
+
+  rows: any;
```
driskull: should these be typed?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  lastSelectedRow = null;
```
driskull: should this be typed?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  @Event() calcitePickerSelectionChange: EventEmitter;
+
+  @Event() calcitePickerRowsDeleted: EventEmitter;
```
driskull: should be named `calcitePickerItemDelete` if delete stuff is staying

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+
+  @Listen("calcitePickerRowToggled") calcitePickerRowToggledHandler(event): void {
+    event.stopPropagation(); // private event
+    const { row, selected, shiftPressed } = event.detail;
+    if (selected) {
+      if (this.multiple && shiftPressed && this.lastSelectedRow) {
+        const start = this.rows.indexOf(this.lastSelectedRow);
```
driskull: suggestion: add a const for `this.rows`

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+  }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Public Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method() async getSelectedRows(): Promise<Set<object>> {
+    return this.selectedValues;
```
driskull: I'm confused on why this method is needed if its only just returning `this.selectedValues`? Shouldn't `selectedValues` just be a property then?

Wouldn't it be more DOM-like for the user to just get the selected rows by doing this: `component.querySelectorAll("calcite-picker-item[selected=true]")`?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  getIconType() {
+    let type = null;
+    if (this.mode === "configuration" && this.dragEnabled) {
+      type = "grip";
+    } else if (this.mode === "selection" && this.multiple) {
```
driskull: suggestion, have a const for `this.multiple`

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+        <CalciteIcon size={action.icon.size} path={action.icon.path} />
+      </calcite-action>
+    ) : null;
+  }
+
+  renderEditButton() {
+    return this.editEnabled ? (
+      !this.editing ? (
+        <section>
+          <calcite-action onClick={() => this.startEdit()}>
```
driskull: Add some text to the action for accessibility. may apply to other places

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+            <CalciteIcon size="16" path={pencil16} />
+          </calcite-action>
+        </section>
+      ) : (
+        <section>
+          <button onClick={() => this.cancelDelete()}>Cancel</button>
```
driskull: text needs to be prop

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,291 @@
...
+          <button onClick={() => this.confirmDelete()}>OK</button>
```
driskull: text needs to be prop

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: yes

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: It seems like we'd be losing more than we'd be gaining. It's very limiting if we do it that way. It also forces whoever is using this component to massage the data to match a the format we specify. Accepting an object property is fine, especially in the context of a JSX ecosystem. On the other hand, the loss in flexibility and the added demand on the consumer seem like notable negatives.

Is there a gain to this I'm not seeing?

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: separate PR.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: draggable is a native prop. I had that originally, but I couldn't use it. I forget the error I was getting.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: this is getting removed.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: not staying.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: It's generally accepted for components that properties shouldn't mutate. State mutates. This method exposes a way to get that state.

The data schema for querySelectorAll won't match what the event handler provides. This method will get back the same data exactly.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: edit is going away.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: edit is going away.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: edit is going away.

src/components/calcite-picker-row/calcite-picker-row.scss
pr3tori4n: @asangma Would we also swap the positioning of the icon and secondary icon?

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: Is that a legit use case? If a user is setting selected, then they already have a list of items. Would they want the event to fire in that scenario - giving them the updated internal list if they have a master list they are using to set the state?

src/components/calcite-picker-row/calcite-picker-row.tsx
driskull: What we gain is the ability to set the property just using a string rather than an object. 

Right now, if you want to set the property without JSX you have to use an object which means you have to have access to the node. 

I'm not sure this is valid: `metadata='{"category":"dinosaur", "tags":["carnivore"]}'`

I think you would have to have the node and set `node.metadata = json;`

src/components/calcite-picker/calcite-picker.tsx
driskull: Ok. @jcfranco is this ok? I'm just not seeing why we need this method. IMO we should go through dom queries to get the "data" which would just be dom elements.

src/components/calcite-picker/calcite-picker.tsx
driskull: gotcha 👍

src/components/calcite-picker-row/calcite-picker-row.tsx
driskull: It seems there is a disconnect between `toggle` and `selected`. They don't do the same thing but they are kind of acting together. If we want to watch the `selected` attribute and make that the source of truth then the event should probably be named `calcitePickerSelectedChange` or something similar. 

It seems like the `toggle` method was introduced just to handle which key was pressed but I think that logic belongs in the Picker widget itself and not on the `row/item`

src/components/calcite-picker-row/calcite-picker-row.tsx
driskull: the other issue is if you want to change the category or the tags only, you have to set the whole object along with the tags. So you have to maintain a copy of the object.

src/components/calcite-picker/calcite-picker.tsx
driskull: I think that the question is what is the source of truth? Is it the `row/item` nodes that have a selected attribute? In this code, it seems like the Picker is maintaining the state of what is selected, so there could be some kind of disconnect there. If we just relied on the row/item nodes being the source of truth for what is selected then this property wouldn't be necessary. right?

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: Fair points. there's give and take both ways. I think this is another scenario where the end-user needs should drive our decision.
@asangma @kerryrobinson can you find out what the needs are for the mapmaker from Kevin and/or Adelheide and let us know?

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: @asangma @kerryrobinson can you guys help define the requirements? Do our dev users need the ability to toggle the selected attribute, or will a public method suffice? There's currently a technical challenge with watching the property and being able to support the shift-click functionality. We'll discuss internally if there's a way to get **click**, **shift+click** as well as **property change** behavior all working together, but if there's not a simple solution, are we ok to proceed with having **click**, **shift+click** and **public method** as ways of interacting with the picker rows?

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: I think this is another scenario where the end-user needs should drive our decision.
@asangma and @kerryrobinson can you do some research here and find out from Kevin, Adelheide and/or Kumar how they will be using this, and what information they are going to need from the picker? Will they be able to get by with querying dom nodes? Will they need an event listener to detect changes to selections in the picker? Will they need a method that can get them that data? Some questions to keep in mind is whether or not selection order is important - they won't be able to determine that if querying dom nodes, but the picker can track that state and surface it.

src/components/calcite-picker-row/calcite-picker-row.scss
asangma: Shouldn't flex handle that?

driskull: yeah it should

src/components/calcite-picker-item/calcite-picker-item.scss
```diff
@@ -0,0 +1,36 @@
...
+:host([hidden]) {
+  display: none;
+}
+
+:host(.highlight) {
+  border-left: 3px solid transparent;
+}
+
+:host(.highlight[selected]) {
+  border-left-color: blue;
```
driskull: his would also need RTL support.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,219 @@
...
+   */
+  @Prop({ reflect: true }) textHeading: string;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @State() selectedValues = {};
```
jcfranco: Please type. Does `Record<string, HTMLCalcitePickerItem>` look correct?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  items: any;
```
jcfranco: Please type.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  lastSelectedItem = null;
```
jcfranco: Please type.

src/components/calcite-picker/calcite-picker.tsx
jcfranco: I think putting this off until @asangma or @kat10140 follow up is good. API-wise, it does make sense to me to expose props/methods for this. I'd expect a property for this since it aligns with `HTMLSelectElement#selectedOptions` (read-only) and Stencil also recommends using props over methods.

Whether the item has a selected attribute or not for me it depends on the target API. Would we want something similar to how `HTMLSelectElement` + `HTMLOptionElement` work together or not?

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+export class CalcitePickerItem {
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Prop({ reflect: true }) editing = false;
+
+  @Prop() metadata: object;
```
jcfranco: This is going away (along w/ `editing`), right?

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  @Element() el: HTMLElement;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  @Event() calcitePickerItemToggled: EventEmitter;
```
jcfranco: Rename event to `calcitePickerItemToggle` for consistency.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  iconClickHandler(e) {
```
jcfranco: Please type method arguments.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+    this.toggle(e.shiftKey);
+  }
+
+  secondaryActionContainerClickHandler(e) {
```
jcfranco: Nitpick: `event` vs `e`.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+  // --------------------------------------------------------------------------
+
+  renderIcon() {
+    const { icon } = this;
+    if (!icon) {
+      return null;
+    }
+    if (icon === ICON_TYPES.grip) {
+      return (
+        <span class="handle">
```
jcfranco: Move classes to CSS object.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,137 @@
...
+      );
+    }
+  }
+
+  render() {
+    const { icon } = this;
+    return (
+      <Host
+        class={icon !== ICON_TYPES.square && icon !== ICON_TYPES.circle ? CSS.highlight : null}
+        onClick={this.iconClickHandler.bind(this)}
```
jcfranco: Bind the method in its declaration to avoid creating a new function on each render.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,219 @@
...
+  // --------------------------------------------------------------------------
+
+  @Event() calcitePickerSelectionChange: EventEmitter;
+
+  @Listen("calcitePickerItemToggled") calcitePickerItemToggledHandler(event) {
+    event.stopPropagation(); // private event
+    const { item, value, selected, shiftPressed } = event.detail;
+    if (selected) {
+      if (this.multiple && shiftPressed && this.lastSelectedItem) {
+        const start = this.items.indexOf(this.lastSelectedItem);
```
jcfranco: Since `this.items` is used a few times, could you use a local variable for readability? May also apply to other properties.

pr3tori4n:
@jcfranco can you take a second look. Your comments have been addressed.

src/components/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: Found a solution. Watching the selected prop has been added.

pr3tori4n:
@driskull I've addressed your feedback. Please review.

src/components/calcite-picker-item/calcite-picker-item.tsx
jcfranco: Both approaches end up creating a new function on each render call. You can see this in the output code:

```js
render() {
  // ...
    h("div", {
      onClick: this.iconClickHandler.bind(this) // new fn
    },
  // ...
}
```

The way to avoid this is to bind it in the handler declaration:

```tsx
// gets bound and initialized in the constructor (after compilation)
iconClickHandler = (event: MouseEvent):void => { /* °˖✧◝(⁰▿⁰)◜✧˖° */ } 
```

which will end up something like this:

```js
 constructor() {
    this.iconClickHandler = (event) => { /* °˖✧◝(⁰▿⁰)◜✧˖° */ } 
    // ...
  }

 render() {
  // ...
    h("div", { 
      // ...
      onClick: this.iconClickHandler // handler already bound
    },
  // ...
}
```

src/components/calcite-picker-item/calcite-picker-item.tsx
jcfranco: This is still pending.

src/components/calcite-picker-item/calcite-picker-item.tsx
jcfranco: > it could be dropped from this PR and added to the filter if that's clearer.

👍 to keep the PR focused.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+    this.calcitePickerSelectionChange.emit(selectedValues);
+  }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  setupItems(): void {
```
jcfranco: Nitpick: `setUpItems` (verb vs noun). It'd also be consistent w/ `setUpDragAndDrop`.

src/components/utils/guid.ts
```diff
@@ -0,0 +1,14 @@
+function generateId(counts: number[]): string {
```
jcfranco: Does this need a test/spec file? Could be handled in a follow-up issue to add tests for utils.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  iconClickHandler(event: MouseEvent) {
+    this.isSelected = !this.isSelected;
+    this.emitToggleEvent(event.shiftKey);
```
driskull: It's still a little odd that the `shiftKey` is captured here since it seems like more of a concern of the `calcite-picker` whether or not the shiftKey matters.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+    this.items = Array.from(this.el.querySelectorAll("calcite-picker-item"));
+    this.items.forEach((item) => {
```
driskull: could the onclick for shift key handling be setup here instead?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -0,0 +1,226 @@
...
+    });
+    if (this.dragEnabled && this.mode === "configuration") {
+      this.setUpDragAndDrop();
+    }
+  }
+
+  setUpDragAndDrop(): void {
+    const sortGroups = [this.el, ...Array.from(this.el.querySelectorAll("calcite-picker-group"))];
+    sortGroups.forEach((sortGroup) => {
+      Sortable.create(sortGroup, {
```
driskull: should `Sortable` be destroyed when the component is?

src/components/calcite-picker-item/calcite-picker-item.tsx
pr3tori4n: ah - yea didn't realize .bind created a new function. kk Changing.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: I looked into that. the entire click handler could be moved here (shift and all). But that didn't solve any problems, it just moved them. The state/prop split is what solved the problems and provided a solution for introducing watching the "selected" prop. I don't see any reason to move the click since it won't be solving anything.

src/components/calcite-picker-item/calcite-picker-item.scss
```diff
@@ -0,0 +1,44 @@
...
+  border-left: 3px solid transparent;
+  &[dir="rtl"] {
+    border-left: none;
+    border-right: 3px solid transparent;
+  }
+}
+
+:host(.highlight[selected]) {
+  border-left-color: blue;
+  &[dir="rtl"] {
```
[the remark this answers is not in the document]
pr3tori4n: You're right. There is.

src/components/calcite-picker-item/calcite-picker-item.tsx
pr3tori4n: doesn't hurt anything.

src/components/utils/guid.ts
pr3tori4n: copied and pasted this from calcite repo as an interim solution - need to find out how both repos can properly share these utils - so we should only need to add a test once. Assuming this eventually goes in another repo - we can add the test there.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,180 @@
...
+        </span>
+      );
+    }
+  }
+
+  render() {
+    const { icon } = this;
+    return (
+      <Host
+        dir={this.dir}
```
driskull: In other components, we've been using a utility class to set the RTL CSS. The problem with this is that if a `dir` changes on a parent, this will be set on the host.

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,180 @@
...
+        class={icon !== ICON_TYPES.square && icon !== ICON_TYPES.circle ? CSS.highlight : null}
```
driskull: Should we be setting classes on the host element? @jcfranco

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,180 @@
...
+        onClick={this.pickerClickHandler}
+        selected={this.isSelected}
```
driskull: Why is this set on here? Shouldn't the `selected` prop already handle it?

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,178 @@
...
+  secondaryActionContainerClickHandler(event: MouseEvent) {
+    event.stopPropagation();
+  }
+
+  emitChangeEvent(shiftPressed = false) {
+    this.calcitePickerItemSelectedChange.emit({
+      item: this.el,
+      value: this.value,
+      selected: this.isSelected,
+      shiftPressed
```
driskull: It doesn't hurt anything to have the shiftKey here, but if the click handler was on the `PIcker` it would be unnecessary to pass it via event.

driskull: The Picker is what needs to know if a shift key was pressed, so it should be listening to the click event instead of relying on that event information to be passed via a child component.

pr3tori4n: This is a stylistic thing that has no impact beyond it being different. Closing out this thread, please open a new issue if you want to discuss further.

src/components/calcite-picker-item/calcite-picker-item.tsx
pr3tori4n: The prop is no longer mutable, so we need to have this here if we want to reflect the state back to the dom. If we DON'T want to reflect the state back to the dom, this can be removed.

src/components/calcite-picker-item/calcite-picker-item.tsx
pr3tori4n: ok I'll do that.

pr3tori4n: worth noting that if dir changes on a parent - this component will likely need to re-render somehow anyway since a parent change wouldn't trigger a re-render.

src/components/calcite-picker-item/calcite-picker-item.tsx
driskull: ok. makes sense.

src/components/calcite-picker-item/calcite-picker-item.tsx
driskull: Its an architectural difference. But if we decided that control clicking should also do the same as shift, it would require a change on the custom event. It would be cleaner if the event information didn't have to be passed to the picker.

src/components/calcite-picker-item/calcite-picker-item.tsx
driskull: IMO, I don't think we should be setting classes on the host element. These should probably be in the shadowDOM.

pr3tori4n: I remembered reading that somewhere.
https://developers.google.com/web/fundamentals/web-components/best-practices#do-not-self-apply-classes

The highlight will go away once we split these - so that will work itself out. 

But the dir I changed to use a class per other feedback - so we can go back to using an attribute for it. I'd like to avoid adding another wrapping element if its redundant. We could use something other than dir if there's concern that will subvert changes to the parent tree, but even then there will be weirdness if this component detects a dir on load. Let's chat about how best to handle this.

PR #142, opened 2019-08-06
Filter Component

pr3tori4n:
**Related Issue:** #9 

## Summary

This introduces a generic filter component. Internally, it uses an text input for interaction. It accepts a dataset as one of it's properties (an array of objects). Typing in the text-input triggers a search for any string value in the dataset, looking for a partial match. Any objects with a partially matching string will be considered a match, all others will be filtered out.

The component fires an event `calciteFilterChange` and passes back a subset of the initial data set passed in with the filtered results.

This is pretty generic in that it doesn't have opinions on the schema of objects in the data-set passed in. And it just reports the results of the filtering, allowing a higher order component to act on the filtered items.

NOTE, this branch's base is the picker branch, so the PR is going into that branch until the picker is merged into master.


src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -220,6 +223,25 @@ export class CalcitePicker {
     this.editing = false;
   }
 
+  handleFilter(filteredData) {
+    const values = filteredData.map((item) => item.value);
+    this.rows.forEach((row) => {
+      row.toggleAttribute("hidden", values.indexOf(row.value) === -1);
```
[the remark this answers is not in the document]
pr3tori4n: might need to flesh this out further with an enhancement - will we want to pre-bake certain states for non-matches in the picker and have a prop to configure which to use?
@asangma can you open an issue for this if this is in fact an added requirement?

Fixed toggleAttribute.

package.json
```diff
@@ -38,6 +38,8 @@
     "@esri/calcite-base": "github:Esri/calcite-base#3d18d24",
     "@esri/calcite-colors": "1.2.1",
     "@esri/calcite-ui-icons": "2.5.0",
+    "calcite-fonts": "github:ArcGIS/calcite-fonts#v1.1.2",
```
driskull: Do we need calcite fonts for this?

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Prop() textLabel: string;
+
+  @Prop() textPlaceholder: string;
+
+  @Prop() data: object[];
```
driskull: I forgot, but do we want to handle exact vs like matches?

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Element() el: HTMLElement;
```
driskull: This should be `HTMLCacliteFilterElement` right?

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render() {
+    return (
+      <Host>
+        <label>
```
driskull: Should we support a label/input as a slot or just an input as a slot?

src/components/calcite-picker-group/calcite-picker-group.tsx
```diff
@@ -0,0 +1,34 @@
+import { Component, Host, Prop, h } from "@stencil/core";
```
driskull: This file needs to be removed

src/components/calcite-picker-group/calcite-picker-row.e2e.ts
```diff
@@ -0,0 +1,25 @@
+import { newE2EPage } from "@stencil/core/testing";
```
driskull: This file needs to be removed

src/components/calcite-picker-group/resources.ts
```diff
@@ -0,0 +1,3 @@
+export const CSS = {
```
driskull: This file needs to be removed

src/components/calcite-picker-item/calcite-picker-item.scss
```diff
@@ -0,0 +1,44 @@
+:host {
```
driskull: This file needs to be removed

src/components/calcite-picker-item/calcite-picker-item.tsx
```diff
@@ -0,0 +1,185 @@
+import {
```
driskull: This file needs to be removed

src/components/calcite-filter/calcite-filter.tsx
pr3tori4n: partial matching. The exact matching is an app feature that I provided a demo to Adelheide for.

src/components/calcite-filter/calcite-filter.tsx
pr3tori4n: not sure. what's the use case you have in mind?

driskull: We can tackle it as an enhancement if a good use case arises. Things like handling validation or setting input types.

src/components/calcite-filter/calcite-filter.tsx
driskull: ok

pr3tori4n:
removed the obsolete files.

package.json
```diff
@@ -38,6 +38,8 @@
     "@esri/calcite-base": "github:Esri/calcite-base#3d18d24",
     "@esri/calcite-colors": "1.2.1",
     "@esri/calcite-ui-icons": "2.5.0",
+    "lodash-es": "4.17.15",
+    "popper.js": "1.15.0",
```
driskull: popper should be gone.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -199,6 +202,30 @@ export class CalcitePickList {
     });
   }
 
+  handleFilter = (event) => {
+    const filteredData = event.detail;
+    const values = filteredData.map((item) => item.value);
+    this.items.forEach((item) => {
+      if (values.indexOf(item.value) === -1) {
+        item.setAttribute("hidden", "");
```
driskull: Is the filter component only for hiding or is it for matching elements too? If its for matching/highlighting then we probably shouldn't hide but set a class on it.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -199,6 +202,30 @@ export class CalcitePickList {
...
+      }
+    });
+  };
+
+  getItemData() {
+    const result = [];
+    this.items.forEach((item) => {
+      const obj = {};
+      Array.from(item.attributes).forEach((attr: any) => {
+        obj[attr.name] = attr.value;
```
driskull: is the `name` part of a schema somewhere?

src/demos/calcite-pick-list.html
```diff
@@ -10,21 +10,21 @@
     <calcite-demo-nav page-id="calcite-pick-list"></calcite-demo-nav>
     <section class="example-container">
       <calcite-pick-list id="one" text-heading="Selection Pick List - Multi" multiple>
-        <calcite-pick-list-item text-heading="T. Rex" text-description="arm strength impaired" value="trex">
+        <calcite-pick-list-item text-heading="T. Rex" text-description="arm strength impaired" value="trex"  metadata='{"category":"dinosaur", "tags":["carnivore"]}'>
```
driskull: I think we should update the samples to set the metaData on the node rather than through the attribute since that isn't valid because its an object. We don't want authors doing this.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -63,6 +63,8 @@ export class CalcitePickList {
 
   @State() selectedValues: Map<string, HTMLCalcitePickListItemElement> = new Map();
 
+  @State() dataForFilter: object[] = [];
```
driskull: If we expect the filter to be a reusable component, then any component that uses the filter will have to do this `getItemData` method and `dataForFilter` state. 

It might make it much easier on the developer if this was handled by the Filter component itself.

For example:

```
CalciteFilter {
  referenceElement: HTMLElement; // The element used to query for filter items
  dataAttr: string; //  The attribute used to store the metadata. EG: 'metadata' 
  selector: string; // The selector to query for filter items. EG: "calcite-picker-item"
  matchedClass: string; // The class applied to matches. EG: "calcite-filter-matched"
  unmatchedClass: string; //  The class applied to non matches. EG: "calcite-filter-unmatched"
}
```

src/components/calcite-filter/calcite-filter.tsx
jcfranco: For maintainability, can you order props alphabetically? We could formalize it in https://github.com/Esri/calcite-app-components/issues/172.

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  @Prop() data: object[];
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Element() el: HTMLCacliteFilterElement;
```
jcfranco: Typo.

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,93 @@
...
+  shadow: true
+})
+export class CalciteFilter {
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Prop() textLabel: string;
```
jcfranco: Can you add some internal doc for this component?

src/components/calcite-pick-list/calcite-pick-list.tsx
jcfranco: @driskull Can we wait for more use cases to arise before extracting components?

src/components/calcite-pick-list/resources.ts
```diff
@@ -7,3 +7,6 @@ export enum ICON_TYPES {
   square = "square",
   grip = "grip"
 }
+export const TEXT = {
+  filterPlaceholder: "filter results"
```
jcfranco: Should `filter` be capitalized?

src/demos/calcite-pick-list.html
jcfranco: Agreed. I thought we had decided to not support complex data through attributes.

src/components/calcite-pick-list/calcite-pick-list.tsx
jcfranco: @driskull https://developer.mozilla.org/en-US/docs/Web/API/Attr

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -199,6 +202,30 @@ export class CalcitePickList {
...
+        item.removeAttribute("hidden");
+      }
+    });
+  };
+
+  getItemData() {
+    const result = [];
+    this.items.forEach((item) => {
+      const obj = {};
+      Array.from(item.attributes).forEach((attr: any) => {
```
jcfranco: Are you seeing any warnings/errors by not typing `attr`? TS should infer the `Attr` type.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -199,6 +202,30 @@ export class CalcitePickList {
...
+      } else {
+        item.removeAttribute("hidden");
+      }
+    });
+  };
+
+  getItemData() {
+    const result = [];
+    this.items.forEach((item) => {
+      const obj = {};
```
jcfranco: Let's provide the type whenever possible. `obj` can be typed as `Record<string, string>`. You can use this type `result` and `getItemData`.

Also, is there a more expressive name than `obj`?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -50,6 +50,8 @@ export class CalcitePickListItem {
 
   @Prop({ reflect: true }) icon: ICON_TYPES | null = null;
 
+  @Prop() metadata: object;
```
jcfranco: Can you add JSDoc to all public API?

src/components/calcite-pick-list/calcite-pick-list.tsx
driskull: oh. duh :)

src/components/calcite-pick-list/calcite-pick-list.tsx
driskull: Yep!

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: the filter is provided the `filteredData` array. The pickList is controlling that any items not in that filteredList get hidden. We could hide them by whatever mechanism (adding a class vs hidden attribute).

src/demos/calcite-pick-list.html
pr3tori4n: I've updated the code to pass the metadata via a property.

These sorts of implementation details will really be in the hands of the mapmaker team. The web component will support accepting rich data via any valid mechanism, and it will get parsed.

So the rich data is supported as a prop of the component. It can be passed in as either an attribute (which will probably make the most sense for those using JSX) or a property set via dot notation.

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: Yea there may be an opportunity here to tuck away more inside the filter internals, but we'll want to make sure we don't limit it's uses in doing so.

src/demos/calcite-pick-list.html
driskull: We don't want the mapmaker team thinking they can do this without JSX which is what it was doing above.

src/components/calcite-pick-list/resources.ts
pr3tori4n: I prefer camelCase over PascalCase personally, and that's what we've been using. You feel strongly?

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: No. I can drop the any here.

src/components/calcite-pick-list/resources.ts
driskull: I think he's referring to the "filter results".

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: > Also, is there a more expressive name than obj? 
it's a generic container for key:value pairs of unknown name and value type. Open to suggestions.

pr3tori4n:
@driskull @jcfranco responded to comments. Please take another look.l

src/components/calcite-pick-list/resources.ts
driskull: `filterPlaceholder: "Filter results"`

src/components/calcite-pick-list/calcite-pick-list.tsx
driskull: I just think it would be nice to componentize this logic so that it is consistent throughout any other components that want to filter items. Can be handled later.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -193,6 +196,31 @@ export class CalcitePickList {
...
+  };
+
+  getItemData(): Record<string, string>[] {
+    const result: Record<string, string>[] = [];
+    this.items.forEach((item) => {
+      const obj: Record<string, string> = {};
+      Array.from(item.attributes).forEach((attr) => {
+        obj[attr.name] = attr.value;
+      });
+      obj.metadata = item.metadata;
```
driskull: TypeScript: src/components/calcite-pick-list/calcite-pick-list.tsx:218:7
           Type 'object' is not assignable to type 'string'.

    L217:  });
    L218:  obj.metadata = item.metadata;
    L219:  result.push(obj);

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,103 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render() {
+    return (
+      <Host>
+        <label>
```
driskull: Boulder:

From aXe:

```
Form elements must have labels

Issue description
Ensures every form element has a label

Impact: critical
Learn more
Element location
#one calcite-filter input
Element source
<input placeholder="Filter results">
```

src/demos/calcite-pick-list.html
```diff
@@ -223,13 +235,13 @@ <h2>RTL</h2>
...
+        const pickList4 = document.querySelector("#one_rtl");
         pickList.addEventListener("calcitePickListSelectionChange", (event) => {
           console.log(event.detail);
         });
       </script>
 
-      <calcite-pick-list id="two">
+      <calcite-pick-list id="two_rtl">
         <calcite-pick-list-item text-heading="ID" text-description="This is an id for this field and people use it for things." value="ID">
           <calcite-action slot="secondaryAction" id="action-test" label="click-me" onClick="console.log('clicked');">
```
driskull: sand: aXe is complaining about these ids not being unique.

src/components/calcite-filter/calcite-filter.tsx
pr3tori4n: that doesn't make sense - there is a label.

driskull: @pr3tori4n probably because `textLabel` is empty by default and not being set on the component. Maybe it should have a default value if not set??

src/demos/calcite-pick-list.html
pr3tori4n: @asangma that was part of your PR for styling and updating the demos. Could you update these ids?

pr3tori4n:
@driskull I'm going to move forward with merging this. Can you open two follow up issues for the things Axe found?

driskull:
> Can you open two follow up issues for the things Axe found?

Feel free to create them as followup issues

PR #156, opened 2019-08-12
Merge _support folder into utils.

jcfranco:
**Related Issue:** N/A

## Summary

Merges folders since both include utilities.

## Notes

* Builds fine and tests are ✅.
* This does affect the generated output files.


driskull:
Should this util move too? `https://github.com/Esri/calcite-app-components/blob/master/src/utils/localStorage.ts`

jcfranco:
Maybe. I tackled the ones under components since those are definitely mergeable. @pr3tori4n WDYT?

driskull:
I think we could move the other utility in this folder as well.

jcfranco:
All component utils are now under the same folder. 🔧

PR #160, opened 2019-08-13
update @stencil/core to v1.2.5

pr3tori4n:
## Summary
- Update @stencil/core to v1.2.5
- Also add `.npmrc` file to make save-exact true by default
- All tests are passing.
- Demos look unaffected.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+save-exact=true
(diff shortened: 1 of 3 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+save-exact=true
```
jcfranco: 🎉🎉🎉🎉🎉🎉 *Forgot to submit earlier. 😄

Issue #164, opened 2019-08-15
Enhancement: allow the content slot of calcite-shell-panel to be hidden, but action bar remain visible


kevindoshier:
### Description
The use case I have is I need to toggle the visibility of the "content" slot when actions from the action bar are triggered, right now Im not able to do that
### Helpful Details

**Which Component:**
`<calcite-shell-panel>`
**Example Use Case:**

![image](https://user-images.githubusercontent.com/11093161/63113832-481e3300-bf48-11e9-8521-420b5a83ef24.png)

![image](https://user-images.githubusercontent.com/11093161/63113868-58361280-bf48-11e9-8926-5a622ba063a7.png)

cc: @driskull @jcfranco

kevindoshier:
This looks good now, thanks

PR #165, opened 2019-08-15
#164 add collapsed property to hide content

driskull:
**Related Issue:** #164 

## Summary

add collapsed property to hide shell panel content

### DIFF
--- a/src/components/calcite-shell-panel/calcite-shell-panel.tsx
+++ b/src/components/calcite-shell-panel/calcite-shell-panel.tsx
@@ -35,11 +35,11 @@ export class CalciteShellPanel {
   render() {
     const { collapsed, layout } = this;
 
-    const contentNode = !collapsed ? (
-      <div class={CSS.content}>
+    const contentNode = (
+      <div class={CSS.content} hidden={collapsed}>
         <slot />
       </div>
-    ) : null;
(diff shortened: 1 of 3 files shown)


src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -28,13 +33,13 @@ export class CalciteShellPanel {
   // --------------------------------------------------------------------------
 
   render() {
-    const { layout } = this;
+    const { collapsed, layout } = this;
 
-    const contentNode = (
+    const contentNode = !collapsed ? (
```
pr3tori4n: Weren't we running into an issue in IE11 or Edge (I forget) where this pattern wasn't working properly? To fix we had to instead conditionally add an attribute and use `display:none` if I recall correctly.

driskull: I don't recall that. I just recall not using `toggleAttribute` for edge.

driskull: ah, I think you're right. The Edge polyfill still lets slot content in. :(

driskull:
Added edge fixes

PR #166, opened 2019-08-15
Picker component: End to End Tests

pr3tori4n:
**Related Issue:** #9

## Summary

Adding tests for `<calcite-picker>`, `<calcite-picker-group>` and `<calcite-picker-row>`.
Some slight refactors to improve architecture for testing.


src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
```diff
@@ -89,7 +94,7 @@ export class CalcitePickerRow {
     if (!this.icon) {
       return null;
     }
-    if (this.icon === "grip") {
+    if (this.icon === ICON_TYPES["grip"]) {
```
driskull: Should we be using a type here instead of a constant?

```ts
type PickerIcon = "square" | "circle" | "grip";

@Prop({ reflect: true }) icon: PickerIcon:
```

That way if something other than one of those is set to the icon it will display an error.

I think the current way it is it won't be able to determine the value since its not using dot noation `ICON_TYPES.square`

You can still have the constant for use within the tests but for defining the type, I think it should be using a string???

My only concern is to make sure the typings are getting picked up correctly.

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -38,7 +38,7 @@ export class CalcitePicker {
 
   @Prop({ reflect: true }) editEnabled = false; /* ignored unless mode is configuration */
 
-  @State() selectedValues = new Set();
+  @State() selectedValues = {};
```
driskull: no more set :(

Would a map work?

src/components/calcite-picker/calcite-picker.tsx
```diff
@@ -199,11 +201,11 @@ export class CalcitePicker {
   getIconType() {
```
driskull: needs a return value for typings

src/components/calcite-picker/calcite-picker.tsx
jcfranco: Can this be typed?

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.tsx
pr3tori4n: A string enum seems to just compile down to a pojo. Would that help with typing here in some way I'm not aware of?
http://www.typescriptlang.org/play/#code/KYOwrgtgBAIglgJ2AYwC5wPYigbwFBSFQCqADlALxQBExACtQDQFEwYDu2V1MA8gOoA5Ji0IAZYADNUlGmICiAMQAqIolABKcAOYALGdw0BJAOIAJVcwC+QA

@driskull I can use dot notation if that helps the typescript compiler. Can you guys help me understand what typing error you guys are trying to avoid? Users aren't able to set the icon, it's all controlled internally from the picker.

pr3tori4n: Going with enum - works fine.

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: > Would a map work?

I think so, will explore it.

> Can this be typed?

It will be a map, with the key will be a string and the value will be a dom object reference. There typing info here you think will add information beyond saying its a map? I don't expect the shape of the dom object to be something we can define.

src/components/calcite-picker/calcite-picker-row/calcite-picker-row.e2e.ts
```diff
@@ -0,0 +1,24 @@
...
+    const isVisible = await row.isVisible();
+    expect(isVisible).toBe(true);
+  });
+
+  it("should toggle selected attribute when clicked", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(`<calcite-picker-row></calcite-picker-row>`);
+    const row = await page.find("calcite-picker-row");
+    expect(row.selected).toBe(false);
```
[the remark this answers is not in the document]
pr3tori4n: I think you're right actually, this doesn't seem to work :D

src/components/calcite-picker/calcite-picker.tsx
pr3tori4n: To expedite progress - let's look into using a map on the main PR. The tests should be unaffected. That way we can get the tests merged, which is a blocker for the other PRs right now. Sound good to you guys? @driskull @jcfranco

package-lock.json
```diff
@@ -1274,20 +1274,18 @@
...
         "@esri/calcite-colors": "^1.2.0",
         "@esri/calcite-ui-icons": "^2.0.1",
-        "calcite-fonts": "github:ArcGIS/calcite-fonts#278f9aac7b4fde778d30ae2a7fa466c3d11a9977",
         "npm-run-all": "4.1.5"
       },
       "dependencies": {
         "@esri/calcite-base": {
-          "version": "github:ArcGIS/calcite-base#490c128d4c9f5cd2d197512882b10898367fd3a5",
-          "from": "github:ArcGIS/calcite-base"
+          "version": "github:ArcGIS/calcite-base#3d18d243f5d2d9fb80a450b7a4950381a6048f1d",
```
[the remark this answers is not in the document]
pr3tori4n: no, but this may be more correct - it can be easy to forget to bump package-lock sometimes.

src/components/calcite-picker/calcite-picker.tsx
jcfranco: Sounds good.

pr3tori4n: After brief chat with Matt, going to leave as POJO unless we need the ability to track insertion order.

Issue #139, opened 2019-08-06
Bug: Shell footer not honoring theme


jcfranco:
### Summary

See cyan box in screenshot below:

<img width="1563" alt="Screen Shot 2019-08-06 at 10 12 12 AM" src="https://user-images.githubusercontent.com/197440/62560978-064c0900-b833-11e9-954d-4296fdf3cf2d.png">

### Actual Behavior

Dark theme font color not being honored.

### Expected Behavior

Footer text color to contrast and fit theme.

### Reproduction Steps

See Dark theme example from https://esri.github.io/calcite-app-components/demos/calcite-shell.html 

### Relevant Info 

Seen on Chrome.

driskull:
@asangma I think this is the last issue for V1.

asangma:
So, if you put a component in there, the theme is honored. Do we want generic styles inside our components? Seems overly prescriptive...?

driskull:
@asangma if the background color variable is being set, it seems like the color one should be too.

"child" components probably shouldn't set the `background-color` or `color`. Only parent ones should.

asangma:
Installed.

jcfranco:
Verified locally.

PR #174, opened 2019-08-19
Added foreground styles

asangma:
**Related Issue:** #139. 

## Summary
Added some styles for color

- Block
- Shell > footer


driskull:
@asangma can you just search for where those two variables are used and make sure they are in the right places?

asangma:
Thanks @driskull. I def was thinking along the same lines. Didn't know if we wanted to make all those changes tied to this issue, but I'm happy to do it.

driskull:
@asangma I'm ok with merging this and then having a separate issue to refactor some of it. 👍

asangma:
Coo. 
Refactor issue created: #175.

asangma:
Thanks y'all.!

Issue #182, opened 2019-08-20
Bug: Resolve error when using both `calcite-components` and `calcite-app-components`


driskull:
### Summary

Bug: Resolve error when using both `calcite-components` and `calcite-app-components`

https://teams.microsoft.com/l/message/19:fd15b51dacd24e70895ec1218a54ae06@thread.skype/1566323822537?tenantId=aee6e3c9-711e-4c7c-bd27-04f2307db20d&groupId=56fae21a-9407-4943-859f-a9bfcf0bbad3&parentMessageId=1566323822537&teamName=Calcite&channelName=Calcite%20Components&createdTime=1566323822537

### Actual Behavior

calcite and calcite-app components are not working together

> Im trying to bring calcite-components into a project but am running into this error

![MicrosoftTeams-image](https://user-images.githubusercontent.com/1231455/63372269-982b3a00-c33a-11e9-910c-22f6ec65cf9a.png)

### Expected Behavior

They should work together

### Reproduction Steps

Ask @kevindoshier 

### Relevant Info 

We may need to change the stencil config namespace to be `"calcite-app"`

If that doesn't solve it, we may need to rename our components to `calcite-app-action` ...etc

jcfranco:
Installed.

kevindoshier:
@jcfranco - thanks, can you update the tag, then I can test in the app

kevindoshier:
Looks good

PR #193, opened 2019-08-21
Issue #182: Use different component namespace for build

jcfranco:
**Related Issue:** #182 

## Summary

The shared `calcite` namespace was causing issues when both component libs were being used side-by-side.

**Note**: this does not change the component or event namespaces.

### DIFF
--- a/stencil.config.ts
+++ b/stencil.config.ts
@@ -4,7 +4,7 @@ import { sass } from "@stencil/sass";
 import autoprefixer from "autoprefixer";
 
 export const config: Config = {
-  namespace: "calcite",
+  namespace: "calcite-app",
   bundles: [
     {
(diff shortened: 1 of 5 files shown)


stencil.config.ts
```diff
@@ -4,7 +4,7 @@ import { sass } from "@stencil/sass";
 import autoprefixer from "autoprefixer";
 
 export const config: Config = {
-  namespace: "calcite",
+  namespace: "calcite-app",
```
driskull: I think you need to update the stencil doc config too

jcfranco: Good catch, thanks!

jcfranco: I also missed the demo script. Will fix.

Issue #185, opened 2019-08-20
Bug: calcite-floating-panel appears on wrong side


AdelheidF:
### Summary
```
     <calcite-shell-panel slot="contextual-panel" layout="leading">
          <calcite-floating-panel slot="floating-panel" position-type="side" hidden></calcite-floating-panel>
          <calcite-flow id="flow">
      </calcite-shell-panel>
```

As positionElement I use the content of a flow item.

### Actual Behavior
floating panel appears on right side of contextual panel

### Expected Behavior
floating panel appears on left side of contextual panel

AdelheidF:
I made sure that my built code has the latest changes. 
E.g. I checked that
...\arcgis-components\node_modules\@esri\calcite-app-components\dist\collection\components\calcite-action-pad\calcite-action-pad.css contains this
```
:host([hidden]) {
  display: none;
}
```

and when I switch my panel to be contextual I still see the floating panel open on the right side.

![image](https://user-images.githubusercontent.com/44481292/63469892-1d3c4f00-c420-11e9-8d07-49200f9c7910.png)

driskull:
@AdelheidF can you change the `calcite-shell-panel` layout to `trailing`?

driskull:
It should be trailing since the panel is on the right side.

AdelheidF:
![image](https://user-images.githubusercontent.com/44481292/63471281-8bcedc00-c423-11e9-91d4-faea724fb10d.png)

Same effect. the floating panel opens on the right side.

driskull:
removing all tips should handle gracefully and not show anylonger.

driskull:
Talk about empty state on tip manager.

driskull:
If not more than 1 tip, do not show pagination

driskull:
TipManager should have a min-height

driskull:
hmm that should work. I'll stop by or create a demo to reproduce.

driskull:
@jcfranco do I need to update the release number to get it to Adelheid?

jcfranco:
You need to use the release scripts, otherwise there's no distributable to ship.

AdelheidF:
Seems to work OK.

AdelheidF:
But I have to use position-type, not placement as the doc says. Same for action-pad.

PR #195, opened 2019-08-21
Component positioning & RTL fixes

driskull:
**Related Issue:** #185

## Summary

- Adds positioning fixes for the floating panel and action pad
  - positioning depends on shell panel leading/trailing value
  - adds RTL support
- Fixes rtl support for `calcite-flow-item` menu
- Adds shell demo file for RTL 

Bug: calcite-floating-panel appears on wrong side #185

@jcfranco @asangma  can we make this higher priority? @AdelheidF needs this fix to continue work.

### DIFF
--- a/src/components/calcite-action-pad/calcite-action-pad.tsx
+++ b/src/components/calcite-action-pad/calcite-action-pad.tsx
@@ -4,6 +4,14 @@ import { CalcitePlacement, CalciteTheme } from "../interfaces";
 
 import { getOffsetTop } from "../utils/position";
 
+import { CSS } from "./resources";
+
+import classnames from "classnames";
+
+import { getElementDir } from "calcite-components/dist/collection/utils/dom";
+
+import { CSS_UTILITY } from "../utils/resources";
+
 @Component({
   tag: "calcite-action-pad",
   styleUrl: "calcite-action-pad.scss",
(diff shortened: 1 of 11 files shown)


src/components/calcite-action-pad/calcite-action-pad.tsx
```diff
@@ -51,15 +59,29 @@ export class CalciteActionPad {
...
+    const rtl = getElementDir(el) === "rtl";
+
+    const closest = el.closest("calcite-shell-panel");
+    const layout = (closest && closest.layout) || "leading";
+
     return (
-      <Host style={style}>
-        <slot />
+      <Host>
+        <div
```
driskull: Added a div here because I don't think we should be adding classes or styles to the host element, right? May need updating in other places

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -64,6 +68,8 @@ export class CalciteFlowItem {
 
   @Element() el: HTMLElement;
 
+  @State() rtl: boolean = getElementDir(this.el) === "rtl";
```
driskull: maybe every component that needs rtl support should have a state like this? Should this be updated on every render or just during initialization?

src/components/calcite-action-pad/calcite-action-pad.tsx
driskull: checked and other places look good

driskull:
I don't think we need to support `dir` changing on the fly. So it should be good upon initialization. However, we should have a common way to handle RTL and only call `getElementDir` once. I'm not sure if it should be once per render or once per componentConnected to handle if a component was moved?

src/components/calcite-action-pad/calcite-action-pad.tsx
jcfranco: 👍

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -179,14 +187,20 @@ export class CalciteFlowItem {
   }
 
   render() {
+    const { el, showBackButton, heading } = this;
+
+    const rtl = getElementDir(el) === "rtl";
+    this.rtl = rtl;
```
jcfranco: I think computing RTL during render calls is fine, no need to store the value. If you want to avoid recomputing it a second time, I'd suggest adding a parameter to `renderBackButton` instead.

Issue #208, opened 2019-08-23
Bug: `position-type` should be `placement`


driskull:
### Summary

 we should be using placement but the css uses position-type

### Actual Behavior

 The `calcite-action-pad` and `calcite-floating-panel` should use a `placement` property but the CSS is referencing `position-type`.

### Expected Behavior

The components should use `placement` in their CSS

AdelheidF:
This looks good, tested it locally.

PR #209, opened 2019-08-23
#208 fix placement bug

driskull:
**Related Issue:** #208 fix placement bug

## Summary

CSS now uses `placement` instead of `position-type`


driskull:
@jcfranco I can make it so that if there is no placement defined the component won't show? Would that be sufficient?

jcfranco:
Yes, along with some descriptive text to call it out.

driskull:
@jcfranco I'm not sure how to do a proper demo for this without creating a bad example. I think this is something we can address when recreating the positioning logic

Issue #197, opened 2019-08-21
Bug: RTL calcite-tip-manager pagination icons point wrong direction


driskull:
### Summary

RTL calcite-tip-manager pagination icons point wrong direction

### Actual Behavior

Buttons should not point towards each other.

### Expected Behavior

Buttons point outward

### Reproduction Steps

1. View sample app and change html dir to rtl

![Screen Shot 2019-08-21 at 1 24 03 PM](https://user-images.githubusercontent.com/1231455/63465964-2117a380-c417-11e9-9965-8ae2ebac6c6a.png)

pr3tori4n:
@driskull In a rtl environment, should the right-side button be `previous tip` and the left-side button be `next-tip`?

driskull:
yep. that sounds right.

pr3tori4n:
@driskull please verify.

PR #222, opened 2019-08-23
pagination icons respect dir=rtl

pr3tori4n:
**Related Issue:** #197 

## Summary

rtl flips the chevrons, so logic added to flip them back.

### DIFF
--- a/src/components/calcite-tip-manager/calcite-tip-manager.tsx
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.tsx
@@ -192,6 +192,7 @@ export class CalciteTipManager {
       // TODO: Empty state
       return <Host />;
     }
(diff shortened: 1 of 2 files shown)


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -192,6 +192,7 @@ export class CalciteTipManager {
       // TODO: Empty state
       return <Host />;
     }
+    const environmentDir = this.el.closest("[dir=rtl]") ? "rtl" : "ltr";
```
jcfranco: Can you use the `getElementDir` util instead? Also, for consistency, rename to `dir`.

driskull: in other components we just call the const `rtl`

Issue #223, opened 2019-08-23
Enhancement: Add events to notify app when panel size has changed


driskull:
### Description

Add events to notify app when panel size has changed

### Acceptance Criteria 

There are events to notify an app when these properties change

#### Desired New Behavior

There are events to notify an app when these properties change

### Relevant Info 

The application knows when the size of a shell panel has changed. Things that can effect its width are:

- collapsed
- action bar text showing

kat10140:
@driskull how do i verify this?

driskull:
@kat10140 you can test the `calciteActionBarToggle` event on the `calcite-action-bar` and the `calciteShellPanelToggle` on the `calcite-shell-panel`.

kat10140:
Verified

PR #224, opened 2019-08-23
Enhancement: Add events to notify app when panel size has changed #223

driskull:
**Related Issue:** #223

## Summary

Enhancement: Add events to notify app when panel size has changed #223

### DIFF
--- a/src/components/calcite-action-bar/calcite-action-bar.e2e.ts
+++ b/src/components/calcite-action-bar/calcite-action-bar.e2e.ts
@@ -61,4 +61,20 @@ describe("calcite-action-bar", () => {
 
     expect(bar).toHaveAttribute("expanded");
   });
+
+  it("expanded change should fire event", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent("<calcite-action-bar></calcite-action-bar>");
+
+    const element = await page.find("calcite-action-bar");
+
+    const eventSpy = await page.spyOnEvent("calciteActionBarExpandedChange", "window");
+
+    element.setProperty("expanded", true);
+
+    await page.waitForChanges();
+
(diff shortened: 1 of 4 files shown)


driskull:
> Would something similar work here?

@jcfranco what are you suggesting? having the event names change? 

These are tied to a property though. Since we can't watch a property, this is how a user would know they changed. right?

jcfranco:
> what are you suggesting? having the event names change?

Yes. I'm suggesting a name change to `calcite<Bla>Toggle` for an event that fires based on expanding/collapsing. With regards to knowing the actual value, can't users get that from the `target`?

driskull:
> Yes. I'm suggesting a name change to calcite<Bla>Toggle for an event that fires based on expanding/collapsing

Ok, ill change to `Toggle` instead of `Change`.

> With regards to knowing the actual value, can't users get that from the target?

Yep, ill remove passing that info

driskull:
@jcfranco updated!

jcfranco:
My bad, I should've been clearer in my comment. I meant to suggest: `calciteActionBarToggle` and `calciteShellPanelToggle`.

driskull:
> My bad, I should've been clearer in my comment. I meant to suggest: calciteActionBarToggle and calciteShellPanelToggle.

Hmm. that seems kind of vague though. What if there's another toggle that happens with a different button? Its related to the attribute change

driskull:
@jcfranco updated

jcfranco:
> Hmm. that seems kind of vague though. 

That's why I wanted to know what your thoughts were. 😄

> What if there's another toggle that happens with a different button? Its related to the attribute change

My suggested event name reflects the change between expanded/collapsed and doesn't really care about how it happened. This doesn't work as nicely if there's a strong need to differentiate.

Issue #214, opened 2019-08-23
Bug: removing all tips tip-manager


kat10140:
### Summary
Removing all tips should handle gracefully.

If tip-manager doesn't have a tip, it needs a developer empty state. So it should empty but be min-height. (it should have a min height)

If not more than 1 tip, do not show pagination.

### Actual Behavior
![Screen Shot 2019-08-23 at 1 36 42 PM](https://user-images.githubusercontent.com/48069902/63622271-10983200-c5ab-11e9-9635-844c3565618b.png)

### Expected Behavior
No line across- empty with min height and buttons removed from dom.

### Reproduction Steps

1. Go to https://esri.github.io/calcite-app-components/demos/calcite-tip-manager.html. Click 'remove last tip' until all are gone. 

### Relevant Info 
all browsers.

pr3tori4n:
@kat10140 @asangma If I put a min-height, and nothing else, then it's just an empty-white box when there are no tips in there. It looks broken still. 
![image](https://user-images.githubusercontent.com/12614215/63625006-27428700-c5b3-11e9-848f-b4bc493a3f1e.png)

I have 2 competing suggestions.
1. We could add a message like "No tips available".
2. We hide the tip manager completely.

Does one of these sounds good, or is there something else you'd like me to do?

pr3tori4n:
Branch: `hrobbins/bug-tip-manager-no-tips-214`

kat10140:
@driskull @asangma Matt- I think you said to keep it min height with just an empty white box if I'm remembering correctly. I like Harry's suggestion of adding a 'no tips' message

driskull:
I'm not sure we need a no tips message because then we have to add a prop to take in that text.

@asangma said that having no tips isn't a valid use case anyway so we can just leave it empty.

kat10140:
that's right

pr3tori4n:
> @asangma said that having no tips isn't a valid use case anyway so we can just leave it empty.

In that case would it be better to just hide it completely? I'm having trouble seeing the benefit of keeping it around with a min-height.

asangma:
No empty state. Retain the TipManager if it's set to be visible.

UX shouldn't have a TipManager with no tips, but the dev may be in the middle of building an app and could benefit from know where/how the TipManager will work without having to have tips in there.

pr3tori4n:
Ok. PR opened for review.
https://github.com/Esri/calcite-app-components/pull/241

[...]

PR #241, opened 2019-09-04
Hrobbins/bug tip manager no tips 214

pr3tori4n:
**Related Issue:** #214 

## Summary
Add logic to handle scenarios where there are only 1 or 0 tips, even after initial load.


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -141,10 +141,13 @@ export class CalciteTipManager {
 
   setUpTips(): void {
     const tips = Array.from(this.el.querySelectorAll("calcite-tip"));
+    this.total = tips.length;
```
jcfranco: Just an observation, but is there a difference between checking `this.total` and `this.tips.length`?

pr3tori4n: functionally no. Performance-wise, it prevents a second lookup to length. Not winning any awards with that optimization, but its a small win.

Issue #126, opened 2019-08-01
Enhancement: TipManager - Arrow key navigation support


driskull:
### Description

TipManager arrow key navigation support

Using the arrow keys while a tip is the active element should allow for pagination.

### Acceptance Criteria 

- Right/Up arrow moves to next tip
- Left/Down arrow moves to previous tip

### Helpful Details

**Which Component:** TipManager
**Example Use Case:** User clicks on a tip or a tip is within focus and can use the arrow keys to paginate.

pr3tori4n:
@driskull  please verify.

driskull:
👍

PR #242, opened 2019-09-05
Hrobbins/enhancement tipmanager arrowkeynavigation 126

pr3tori4n:
**Related Issue:** #126 

## Summary

1. Gives the tipManager a default tabindex. 
Note: This is so the manager itself can receive focus for the purpose of listening to key events. If this isn't what was desired, please let me know and I can remove it. Without this you can tab focus into the manager, but you'd be focusing on either the close button, the link or one of the pagination buttons.
1. Adds event handlers for keydown events triggering pagination navigation for the arrow keys.


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -58,6 +58,8 @@ export class CalciteTipManager {
    */
   @Prop({ reflect: true }) theme: CalciteTheme;
 
+  @Prop({ reflect: true }) tabindex = 0;
```
driskull: I don't think we want to override default attributes for nodes. The only time I think we're doing that is for the `id` in cases where we must have an `id`.

In this case, I think we need to have a container within the host with a `tabIndex` of `0`.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -182,20 +187,57 @@ export class CalciteTipManager {
     this.nextTip();
   };
 
+  tipManagerKeyDownHandler = (event: KeyboardEvent): void => {
+    switch (event.key) {
+      case "ArrowUp":
+        event.preventDefault();
```
driskull: What was up doing?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -182,20 +187,57 @@ export class CalciteTipManager {
...
+      case "ArrowRight":
+        this.nextTip();
+        break;
+      case "ArrowDown":
+        event.preventDefault();
+      case "ArrowLeft":
+        this.previousTip();
+        break;
+      default:
```
driskull: Is this default necessary?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -182,20 +187,57 @@ export class CalciteTipManager {
...
+        break;
+    }
```
driskull: Might be nice for `Home` and `End` to go to the first or last tip.

driskull: Does nextTip() go to the first tip if its the last one?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: I reviewed the best practices before going forward with this. The best practice suggests that it's alright to use native props on the host, so long as you don't override it if a user sets it. I validated that Stencil's @Props() decorator works this way, if a user puts a tabindex, it will use that instead of the default value here.
https://developers.google.com/web/fundamentals/web-components/best-practices#do-not-override-author-set-global-attributes
[See Connected Callback here](https://developers.google.com/web/fundamentals/web-components/examples/howto-checkbox)

So I think we have the best of both worlds here:
1) We don't create a redundant DOM node when the host will suffice
2) We respect any values set by the end-user, and only resort to our default if none is set.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: native behavior is it scrolls the page up/down when you press the arrow keys. This prevents the page from scrolling when focused on the manager.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: Short of it:
No. If you think it would be better if it wasn't there, I'll remove.

Long of it.
I remember learning at one point in the past that a switch statement should always have a default, so I added this one. I just checked and the MDN page says its optional, so I was probably recalling a best practice from back in the day (don't have need of switch statements that often - but it works well for this case). I think that probably had more to do with the switch statement handling core logic, where we'd be fine in this case if nothing at all happened.

Preference?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: > Does nextTip() go to the first tip if its the last one?
Yes, no matter how it gets triggered it will revolve.

> Might be nice for Home and End to go to the first or last tip.
I can add this. Is there hesitation? @asangma @kat10140 thoughts?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
driskull: Minor concern (1) is that we are adding a tabIndex to an element automatically and the end user will see this. If it were in the shadowDOM it wouldn't be as exposed to the user. I'm not sure we want them to be able to change this value here.

Minor concern (2) is that this property will be documented in the readme and it really shouldn't be as it isn't a custom attribute.

driskull: @jcfranco ill let you make the call here

src/components/calcite-tip-manager/calcite-tip-manager.tsx
driskull: Yeah, lets remove it since its empty.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: Seems like concern #2 is an issue we've run into a few times - I'll submit a feature request with Stencil to be able to opt out a prop from being auto-doc'd

pr3tori4n: I tried moving the tabindex to the tipContainer but it loses focus each time the tip swaps, which leads to a jarring experience.

driskull: ok

PR #251, opened 2019-09-07
CalcitePopover component

driskull:
**Related Issue:** #


driskull:
cc @AdelheidF @kevindoshier @jbartley `calcite-popover` component for positioning - PR created.

src/components/calcite-popover/calcite-popover.e2e.ts
```diff
@@ -0,0 +1,59 @@
...
+  it("renders", async () => renders("calcite-popover"));
+
+  it("honors hidden attribute", async () => hidden("calcite-popover"));
+
+  it("popover defaults", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent("<calcite-popover></calcite-popover>");
+
+    const element = await page.find("calcite-popover");
```
jcfranco: The `defaults` test helper could simplify this test. See https://github.com/Esri/calcite-app-components/blob/master/src/components/calcite-block/calcite-block.e2e.ts#L10-L20 as an example.

src/components/calcite-popover/calcite-popover.e2e.ts
```diff
@@ -0,0 +1,59 @@
...
+      const positionElement = document.createElement("div");
+      document.body.appendChild(positionElement);
+      elm.positionElement = positionElement;
+    });
+
+    await page.waitForChanges();
+
+    const computedStyle = await element.getComputedStyle();
+
+    expect(computedStyle.transform).toBe("matrix(1, 0, 0, 1, 8, 26)");
```
jcfranco: Is there a less brittle way to test this?

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -0,0 +1,160 @@
+import { Component, Element, Host, Method, Prop, State, Watch, h } from "@stencil/core";
+
+import { CalcitePlacement } from "../interfaces";
+
+// https://www.npmjs.com/package/popper.js
+// https://popper.js.org/
```
jcfranco: IMO, this isn't needed since it's in `package.json`.

src/demos/calcite-shell.html
```diff
@@ -14,13 +14,53 @@
   </head>
 
   <body>
+    <calcite-popover placement="vertical" id="calcite-position-floating-panel">
+      <calcite-shell-floating-panel
+        hidden
```
jcfranco: Does the popover content need to set `hidden`? I would expect the component to handle this.

src/demos/calcite-shell.html
```diff
@@ -566,31 +559,36 @@ <h2 class="heading">My App</h2>
...
       <script>
         var actionPad = document.getElementById("action-pad");
+        var calcitePositionActionPad = document.getElementById("calcite-position-action-pad");
         var actionPadButton = document.getElementById("action-pad-button");
         actionPadButton.addEventListener("click", function() {
           actionPad.hasAttribute("hidden") ? actionPad.removeAttribute("hidden") : actionPad.setAttribute("hidden", "");
-
-          actionPad.positionElement = actionPadButton;
+          calcitePositionActionPad.positionElement = actionPadButton;
+          calcitePositionActionPad.reposition();
```
jcfranco: Doesn't this kick in when `positionElement` is set?

src/demos/calcite-shell.html
```diff
@@ -14,13 +14,53 @@
   </head>
 
   <body>
+    <calcite-popover placement="vertical" id="calcite-position-floating-panel">
```
jcfranco: I think it'd be clearer to rename the ID to include `popover`.

src/demos/calcite-shell-rtl.html
```diff
@@ -574,31 +559,36 @@ <h2 class="heading">My App</h2>
...
+        var calcitePositionFloatingPanel = document.getElementById("calcite-position-floating-panel");
+        var shellFloatingPanel = document.getElementById("calcite-shell-floating-panel");
         var shellFloatingPanelButton = document.getElementById("shell-floating-panel-button");
         shellFloatingPanelButton.addEventListener("click", function() {
           shellFloatingPanel.hasAttribute("hidden")
             ? shellFloatingPanel.removeAttribute("hidden")
             : shellFloatingPanel.setAttribute("hidden", "");
 
-          shellFloatingPanel.positionElement = shellFloatingPanelButton;
+          calcitePositionFloatingPanel.positionElement = shellFloatingPanelButton;
```
jcfranco: What do you think about supporting an element ID as well as a node? 

Also, if `positionElement` always needs to be set before calling `reposition`, would it be better to update the method signature to include this?

```tsx
calcitePopoverForFloatingPanel.position(elementOrId);
```

And lastly, how does one close the popover ('popunder' 😅)?

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -0,0 +1,160 @@
...
+  //
+  // --------------------------------------------------------------------------
+
+  @Method() async reposition(): Promise<void> {
+    const { el, placement: componentPlacement, popper, positionElement, xOffset, yOffset } = this;
+
+    const placement = componentPlacement === "vertical" ? "bottom-start" : "auto-start";
+    const offsetEnabled = !!(yOffset || xOffset);
+
+    const modifiers = {
```
jcfranco: Can this be typed?

jcfranco:
> Could this move to calcite-components at some point?

I think this would be useful at that level. cc @patrickarlt 

> Does the component need to be in the body?

It would trip me up to see it nested deep in the page and have it potentially move around, so body makes sense to me.

> Do we need/want a utility to create it in the body always?

We wouldn't need this if we make it a best practice to always set it on the body, right?

src/components/calcite-popover/calcite-popover.e2e.ts
driskull: I can probably just verify that the transform is there. Would that be better?

src/demos/calcite-shell.html
driskull: Yep. will remove

src/demos/calcite-shell.html
driskull: Nope. will set it on the component.

src/components/calcite-shell-floating-panel/calcite-shell-floating-panel.tsx
```diff
@@ -1,30 +1,10 @@
...
-
-import classnames from "classnames";
-
-import { getElementDir } from "../utils/dom";
-
-import { CSS_UTILITY } from "../utils/resources";
+import { CSS, TEXT } from "./resources";
 
 @Component({
   tag: "calcite-shell-floating-panel",
```
pr3tori4n: This panel doesn't float anymore, the floating is achieved via the popover. Probably should rename.

src/demos/calcite-shell-rtl.html
driskull: @jcfranco 

> What do you think about supporting an element ID as well as a node?

I don't think we need it unless you feel strongly about it.

> Also, if positionElement always needs to be set before calling reposition, would it be better to update the method signature to include this?

If you already set the node but want to reposition because of a scrolling div or something than passing the node is unnecessary. We could make it an option argument?

> And lastly, how does one close the popover ('popunder' 😅)?

Just add the `hidden` attribute :) We don't need a method to do that. I can watch the hidden property to stop the events and repositioning. I'll add that.

src/components/calcite-popover/calcite-popover.tsx
driskull: I'll see.

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+    if (popper) {
+      popper.options.placement = placement;
+      popper.options.modifiers = { ...popper.options.modifiers, ...modifiers };
+      popper.scheduleUpdate();
+      return;
+    }
+
+    if (el && placement && positionElement) {
+      const newPopper = new Popper(positionElement, el, {
```
pr3tori4n: the docs are somewhat unclear, but you may want to enable `removeOnDestroy`. Destroying it doesn't remove it from the dom by default?

src/demos/calcite-popover.html
```diff
@@ -0,0 +1,35 @@
...
+      <section class="example-container">
+        <h1>calcite-popover</h1>
+        <p>
+          see the
+          <a href="https://github.com/Esri/calcite-app-components/tree/master/src/components/calcite-popover/readme.md"
+            >calcite-action readme.</a
+          >
+        </p>
+
+        <style>
```
pr3tori4n: Think this is incomplete. There supposed to be a popover element example here?

src/demos/calcite-shell-rtl.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
```
pr3tori4n: Can we get a simple demo of a shell by itself, or maybe with only a shell panel? 

I think the demo files should focus on showing off a single component, and only add as many supporting components as necessary (like a tip manager is pointless without a couple tips). This seems to be a full-blown example app. That has it's use, but it makes it hard to see what the shell by itself actually does. 

Perhaps this could go under an  "advanced demo" section. Might be better to split off a new section of demos that show this level of integration.

src/components/calcite-action-pad/calcite-action-pad.tsx
```diff
@@ -1,17 +1,9 @@
-import { Component, Element, Host, Prop, State, Watch, h } from "@stencil/core";
```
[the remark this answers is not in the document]
driskull: Outside of PR scope.

src/components/calcite-popover/calcite-popover.e2e.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    expect(computedStyle.zIndex).toBe("999");
+    expect(computedStyle.position).toBe("absolute");
+    expect(computedStyle.transform).toBe("none");
+  });
+
+  it("popover positiones when positionElement is set", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(`<calcite-popover placement="horizontal"></calcite-popover><div>positionElement</div>`);
```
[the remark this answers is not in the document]
driskull: Correct.

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -0,0 +1,160 @@
...
+    this.reposition();
+  }
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Element() el: HTMLCalciteShellFloatingPanelElement;
```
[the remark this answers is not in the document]
driskull: Yep! Nice catch. copy pasta

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -0,0 +1,160 @@
...
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * Determines where the element will be positioned.
+   * horizontal: Positioned to the left or right of the positionElement.
+   * vertical: Positioned above or below the positionElement.
+   */
+  @Prop({ reflect: true }) placement: CalcitePlacement = "horizontal";
```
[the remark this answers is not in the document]
driskull: Going with horizontal/vertical because it will position it left/right depending on space. 

It sounds better than `above or below` or `left or right`

src/components/calcite-popover/calcite-popover.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    if (el && placement && positionElement) {
+      const newPopper = new Popper(positionElement, el, {
+        eventsEnabled: false,
+        placement,
+        modifiers
+      });
+
+      window.addEventListener("resize", newPopper.scheduleUpdate, { passive: true });
```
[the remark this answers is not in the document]
driskull: Copied this direct from `Popper` so I don't think its an issue.

src/components/calcite-shell-floating-panel/calcite-shell-floating-panel.tsx
driskull: Outside of PR scope but valid. @asangma can think of a better name. Maybe more relative to what it is like a small or temporary panel.

src/components/calcite-popover/calcite-popover.tsx
driskull: The component will handling removing the DOM.

src/demos/calcite-popover.html
driskull: Yep. its not done, listed in todos

src/components/calcite-popover/calcite-popover.tsx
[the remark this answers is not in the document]
pr3tori4n: I'd hesitate to deem example code as production ready - it's usually succinct so as not to confuse the reader with unrelated details. I'm not fully up-to-speed on passive:true so if that is the new way to handle throttling/debouncing then cool - consider me educated. But if it's not a suitable replacement, we should use a debounce/throttle here.

src/components/calcite-popover/calcite-popover.tsx
pr3tori4n: I read about `removeOnDestroy` in the Popper.js options docs. It's false by default. You know what it does (or doesn't do when disabled)? Seems like something we might want but its hard to tell. You sure it removes it from the dom with this being false? If so then what's this for?

src/components/calcite-popover/calcite-popover.tsx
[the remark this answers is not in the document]
driskull: Its not example code. its from popper.js source code. https://github.com/FezVrasta/popper.js/blob/fec3de12c87af087e049d4a868ef7b97494e4f0b/packages/popper/src/utils/setupEventListeners.js#L34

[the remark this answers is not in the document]
driskull: https://iamakulov.com/notes/resize-scroll/

src/components/calcite-popover/calcite-popover.tsx
driskull: Yeah, we don't need it. Its for someone who wants to remove the node when they call destroy on the popper. We don't need that because the component handles it.

driskull: https://github.com/FezVrasta/popper.js/blob/f41c861582ee75ccab44b2e3467c4ec63399dbfa/packages/popper/src/methods/destroy.js#L29

pr3tori4n: That article shows using a throttle for resize. Likewise, it states that

> Currently, only touchstart, touchmove and mousewheel event handlers could be made passive

Also, passive isn't supported on IE11. https://caniuse.com/#feat=passive-event-listener

src/demos/calcite-shell-rtl.html
driskull: Actually, don't need to watch hidden attribute. Removing referenceElement will stop events.

driskull:
> We wouldn't need this if we make it a best practice to always set it on the body, right?

correct

src/components/calcite-popover/calcite-popover.tsx
driskull: @pr3tori4n We don't support IE11. The function that gets called is already throttled.

PR #255, opened 2019-09-10
Hrobbins/npm publish preparation

pr3tori4n:
**Related Issue:** none

## Summary

The primary purpose of this PR is to prepare for publishing Calcite-App-Components to NPM. To support this, the following was done:
* I updated the config of the Stencil build to remove unwanted items. The dev build will build the demo-nav component, but the prod build will not. Also adjusted the config to only build the components directory. I validated both local dev and prod build are working as expected.
* new npm scripts have been added to automate 1) Versioning and 2) publishing.
  * When bumping the version, npm will install the latest dependencies, and run the tests before proceding. Then after the version bump we run the build for dist, and for docs. Then stage those changes as well as the version change. Then a version commit and a tag of that commit gets created. and after that we push that commit and the tag to remote.
  * There's a validation script that runs before publishing to verify you are on the master branch, and if not will give you a cool [ascii error ^_^](https://github.com/Esri/calcite-app-components/blob/hrobbins/npm-publish-preparation/support/prepublish.js#L8).
* 3 convenience scripts have been added `release:patch`, `release:minor` and `release:major` which will run all of the above for single-command deployment.
* .npmrc we added access=public as the default.

## Next Steps (I think Franco is working on these):
* After this PR, we can add a `postPublish` script that will generate a github release using [`gh-release`](https://www.npmjs.com/package/gh-release), similar to what the shells script was doing before - although I think we can skip the zip folder.
* [Automate our changelog](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-cli). For now we'll still need to manually curate our release notes and changelog.

## Verifying Locally
I've published to version `@esri/calcite-app-components@0.0.1-alpha.1` if you'd like to npm install it and verify the package is usable.
**IMPORTANT:** If you want to try running the script, modify the version to an alpha version. NPM versions can be uninstalled but cannot be re-used. If we accidentally publish to v2.x.x that version is used up.


pr3tori4n:
image

.gitignore
```diff
@@ -23,4 +23,4 @@ $RECYCLE.BIN/
 Thumbs.db
 UserInterfaceState.xcuserstate
 .env
-src/components.d.ts
+src/**/components.d.ts
```
jcfranco: Nitpick, `components.d.ts` is always generated directly under `src`, no?

stencil.config.ts
```diff
@@ -3,6 +3,13 @@ import { postcss } from "@stencil/postcss";
 import { sass } from "@stencil/sass";
 import autoprefixer from "autoprefixer";
 
+// Exclude demo components for prod build
```
jcfranco: Nitpick, can production be spelled out?

support/prepublish.js
```diff
@@ -0,0 +1,23 @@
+const chalk = require('chalk');
+const child_process = require("child_process");
+const branch = child_process.execSync("git rev-parse --abbrev-ref HEAD");
+
+if( branch.toString() === "master" ) {
```
jcfranco: Cool! 😎

kat10140:
CALCITE ERROR 😄

README.md
```diff
@@ -99,15 +99,20 @@ This will provide autocomplete of component names/properties, as well as additio
 - Notepad or your favorite HTML editor
 - Web browser with access to the Internet
 
-### Testing
+### Deployment
 
-#### Gotchas
+1. Checkout the master branch. Your git state must be clean (no pending changes).
```
jcfranco: Could you clarify what `Your git state must be clean (no pending changes).` means? Should the staging area, working tree or both be clean?

README.md
```diff
@@ -99,15 +99,20 @@ This will provide autocomplete of component names/properties, as well as additio
...
+1. Run `npm run release:<patch | minor | major>`.
+   Follow semantic versioning. Patch for bug fixes only. Major for breaking changes. Minor for the rest.
 
-Disabling headless mode and enabling devtools causes calls to `page.setContent(<html>)` to fail in end to end tests.
-A workaround is manually modifying the waitUntil option in `<projectDir>\node_modules\@stencil\core\dist\testing\index.js` Line 39000 to 'load'.
+This will create and push a new version commit and tag, then publish that commit to the npm public registry.
+
+**\*Aside:** It will also update the docs for github pages and the readmes for each component.
```
jcfranco: Nitpick: `github` -> `GitHub`, `readmes` -> `READMEs` or `README files`.

support/prepublish.js
```diff
@@ -0,0 +1,23 @@
+const chalk = require('chalk');
+const child_process = require("child_process");
```
jcfranco: Should this be `childProcess`?

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "calcite-app-components",
+  "name": "@esri/calcite-app-components",
   "version": "2.0.2",
-  "description": "A collection of calcite components for building single page applications.",
+  "description": "A collection of web components for building single-page mapping applications.",
```
driskull: I don't think we need to say `mapping`. it will work with anything in the center and not require a map.

.gitignore
pr3tori4n: it's added to wherever the config sirDir property is set to, which was changed to src/components to prevent some of our other source code from being added to the dist/

jcfranco: Ah, good to know. Thanks!

package.json
pr3tori4n: k removed.

Issue #259, opened 2019-09-11
Bug: A shell with only a contextual panel displays on the incorrect side


driskull:
### Summary

Bug: A shell with only a contextual panel displays on the incorrect side

### Actual Behavior

The contextual panel appears on the left side of the shell

### Expected Behavior

A user must have a primary panel. In the primary panel, they can specify a `layout` of `trailing`.

### Reproduction Steps

1. Open the shell demo page
2. Remove the primary panel node

driskull:
cc @AdelheidF

driskull:
@asangma is it a valid use case to just have a `contextual` panel? or should we just close this as `known issue` or `invalid`?

driskull:
cc @kellyhutchins

kellyhutchins:
Awesome! I'll test it out in my app

driskull:
You'll have to wait for our next release @kellyhutchins :(

kat10140:
I'm still seeing it on the left side @driskull :/

driskull:
@kat10140 you must have a primary panel. In the primary panel, you can specify a layout of trailing.

kat10140:
ok in the comment above we have:
``` Expected Behavior
The contextual panel should appear on the right side if just a contextual panel is used without a primary panel. ```

driskull:
Updated issue! Thanks

kat10140:
Mines still not flip flopping 😫

kat10140:
![Screen Shot 2020-01-06 at 2 52 45 PM](https://user-images.githubusercontent.com/48069902/71854825-37f02700-3094-11ea-9f9b-72fd1ee0219e.png)
i see it's technically working but the styling looks wack

driskull:
Can you replecate `/demos/shell/demo-app-full-window-reversed.html`?

kat10140:
![Screen Shot 2020-01-06 at 2 57 35 PM](https://user-images.githubusercontent.com/48069902/71855071-e7c59480-3094-11ea-80e2-21a321fa13c4.png)

driskull:
@kat10140 you may need to do an npm install and restart the build

kat10140:
still doing the same weirdness 🤷🏼‍♀️

driskull:
@kat10140 can you share the HTML you're using? Does the demo `/demos/shell/demo-app-full-window-reversed.html` look ok?

kat10140:
[...]

PR #261, opened 2019-09-11
Bug: A shell with only a contextual panel displays on the incorrect side

driskull:
**Related Issue:** #259

Bug: A shell with only a contextual panel displays on the incorrect side #259


driskull:
Actually this PR isn't good enough :(

We need to consider RTL as well and the `margin-left` won't work in that case. @asangma can you try something with flexbox? I tried and failed.

Issue #258, opened 2019-09-11
Bug: CalciteTipManager tabindex property should not be used


driskull:
### Summary

Bug: CalciteTipManager tabindex property should not be used.

The @Prop() name "tabindex" is a reserved public name. Please rename the "tabindex" prop so it does not
           conflict with an existing standardized prototype member. Reusing prop names that are already defined on the
           element's prototype may cause unexpected runtime errors or user-interface issues on various browsers, so
           it's best to avoid them entirely.

### Relevant Info 

```
[ WARN  ]  Build Warn: src/components/calcite-tip-manager/calcite-tip-manager.tsx:61:28
           The @Prop() name "tabindex" is a reserved public name. Please rename the "tabindex" prop so it does not
           conflict with an existing standardized prototype member. Reusing prop names that are already defined on the
           element's prototype may cause unexpected runtime errors or user-interface issues on various browsers, so
           it's best to avoid them entirely.

     L61:    @Prop({ reflect: true }) tabindex = 0;
```

pr3tori4n:
@driskull installed please verify.

PR #267, opened 2019-09-16
refactor: remove tabindex @Prop

pr3tori4n:
**Related Issue:** #258

## Summary
* This removes the @Prop tabindex.
* Preserve arrow navigation functionality
* Respects user-defined tabindex.

### DIFF
--- a/src/components/calcite-tip-manager/calcite-tip-manager.tsx
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.tsx
@@ -84,8 +84,6 @@ export class CalciteTipManager {
 
   observer = new MutationObserver(() => this.setUpTips());
 
-  tabindex = "0";
-
   // --------------------------------------------------------------------------
   //
   //  Lifecycle
@@ -94,9 +92,6 @@ export class CalciteTipManager {
 
   connectedCallback() {
     this.setUpTips();
-    if (this.el.hasAttribute("tabindex")) {
-      this.tabindex = this.el.getAttribute("tabindex");
-    }
   }
 
   componentDidLoad() {
@@ -245,17 +240,19 @@ export class CalciteTipManager {
       return <Host />;
     }
     return (
-      <Host onKeydown={this.tipManagerKeyDownHandler} tabindex={this.tabindex}>
-        <header class={CSS.header}>
-          <h2 class={CSS.heading}>{this.groupTitle}</h2>
-          <calcite-action text={this.textClose} onClick={this.hideTipManager} class={CSS.close}>
-            <CalciteIcon size="16" path={x16} />
-          </calcite-action>
-        </header>
-        <div class={classnames(CSS.tipContainer, this.direction)} key={this.selectedIndex}>
-          <slot />
+      <Host onKeydown={this.tipManagerKeyDownHandler}>
+        <div tabindex="0">
+          <header class={CSS.header}>
+            <h2 class={CSS.heading}>{this.groupTitle}</h2>
+            <calcite-action text={this.textClose} onClick={this.hideTipManager} class={CSS.close}>
+              <CalciteIcon size="16" path={x16} />
+            </calcite-action>
+          </header>
+          <div class={classnames(CSS.tipContainer, this.direction)} key={this.selectedIndex}>
+            <slot />
+          </div>
+          {this.renderPagination()}
         </div>
-        {this.renderPagination()}
       </Host>
     );
   }
--- a/src/components/calcite-tip-manager/calcite-tip-manager.tsx
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.tsx
@@ -58,8 +58,6 @@ export class CalciteTipManager {
    */
   @Prop({ reflect: true }) theme: CalciteTheme;
 
-  @Prop({ reflect: true }) tabindex = 0;
-
   // --------------------------------------------------------------------------
   //
   //  Private Properties
@@ -86,6 +84,8 @@ export class CalciteTipManager {
 
   observer = new MutationObserver(() => this.setUpTips());
 
+  tabindex = "0";
+
   // --------------------------------------------------------------------------
   //
   //  Lifecycle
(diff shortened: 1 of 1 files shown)


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -242,7 +245,7 @@ export class CalciteTipManager {
       return <Host />;
     }
     return (
-      <Host onKeydown={this.tipManagerKeyDownHandler}>
+      <Host onKeydown={this.tipManagerKeyDownHandler} tabindex={this.tabindex}>
```
driskull: I do still prefer having a container inside the custom element to handle the tabindex that way we are not modifying the tabindex on the host node. We should probably be keeping that logic internal to the component.

It's also more consistent with how a native element works. For example, a button doesn't automatically get a tabIndex but it is focusable by nature due to its internals.

pr3tori4n: You can override the default tabindex of button. By keeping the tabindex on the host, we allow that same override here. If we put the tabindex on a child container and ignore the tabindex on the host, then the user loses the ability to override the tabindex.
```<button>normal button</button>
<button tabindex="-1">can't focus</button>```

File: src/components/calcite-tip-manager/calcite-tip-manager.tsx [hunk 2]
```diff
@@ -86,6 +84,8 @@ export class CalciteTipManager {
 
   observer = new MutationObserver(() => this.setUpTips());
 
+  tabindex = "0";
```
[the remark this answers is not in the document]
pr3tori4n: It stores both the initial value and then the overridden value if one is present. so initialTabIndex is a misnomer.
There another name you prefer that's more descriptive?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
driskull: > If we put the tabindex on a child container and ignore the tabindex on the host, then the user loses the ability to override the tabindex.

That's not true, if the user sets `tabindex="-1"` on the host it won't receive focus through tabbing, nor will its children.

This is exactly how the `calcite-action` works.

pr3tori4n: https://codepen.io/pr3tori4n/pen/VwZEEGv

Tried it in this pen. tabindex="-1" in the outer div doesn't prevent focus on the inner div

driskull: These work the same:

```html
<style>
          .testClass {
            font-family: var(--calcite-app-font-family);
            display: flex;
            color: var(--calcite-app-foreground);
            fill: var(--calcite-app-foreground);
            background-color: var(--calcite-app-background);
            border-radius: var(--calcite-app-border-radius);
            padding: var(--calcite-app-cap-spacing-three-quarters) var(--calcite-app-side-spacing-three-quarters);
            margin: 0;
            justify-content: flex-start;
            align-items: center;
            border: none;
            width: auto;
            cursor: pointer;
            font-size: var(--calcite-app-font-size--1);
            line-height: normal;
            transition: color 125ms ease-in-out, fill 125ms ease-in-out, background-color 125ms ease-in-out;
            text-align: unset;
            position: relative;
          }
        </style>

        <button class="testClass" tabindex="-1">test</button>

        <calcite-action tabindex="-1" text="hello world" text-enabled></calcite-action>
```

driskull: Apparently, shadowDom works differently. There's a whole issue here:
https://github.com/w3c/webcomponents/issues/774

But basically this:

> Setting tabindex=-1 on a shadow host as the effect of disabling the focus of the component. We have builtin elements like input and video elements, and if tabindex=-1 is set on either element, then its content including buttons and media controls aren't focusable.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -94,6 +94,9 @@ export class CalciteTipManager {
 
   connectedCallback() {
     this.setUpTips();
+    if (this.el.hasAttribute("tabindex")) {
+      this.tabindex = this.el.getAttribute("tabindex");
```
driskull: I think we should be consistent with the `calcite-action` and not set a tabIndex on the host.

driskull: This may also be related: https://github.com/ionic-team/stencil/issues/1008

driskull: and this. https://github.com/ionic-team/stencil/issues/1623

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: Seems there are many people suggesting this is unexpected behavior. This may end up changing.

pr3tori4n:
Ok adding a wrapping element with tabindex set to 0. @asangma @kat10140 just an FYI that this means the tip manager will always be a focusable element in the tab order. There will be no way to override this. If that's desired, we'll need to make modifications.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -245,17 +240,19 @@ export class CalciteTipManager {
...
-      <Host onKeydown={this.tipManagerKeyDownHandler} tabindex={this.tabindex}>
-        <header class={CSS.header}>
-          <h2 class={CSS.heading}>{this.groupTitle}</h2>
-          <calcite-action text={this.textClose} onClick={this.hideTipManager} class={CSS.close}>
-            <CalciteIcon size="16" path={x16} />
-          </calcite-action>
-        </header>
-        <div class={classnames(CSS.tipContainer, this.direction)} key={this.selectedIndex}>
-          <slot />
+      <Host onKeydown={this.tipManagerKeyDownHandler}>
```
driskull: Minor: you could move the `onKeyDown` to the element that receives focus but it will bubble up in this case.

driskull:
> the tip manager will always be a focusable element in the tab order.

A user can add a `tabIndex="-1"` to the `calcite-tip-manager` to avoid keyboard tabbing into it.

Issue #269, opened 2019-09-17
Tooling: Remove classnames dependency


driskull:
## Summary

Tooling: Remove classnames dependency

## Measure of Success

Tooling: Remove classnames dependency

**Pain point:**

Unnecessary dependency

**Desired outcome:**

Dependency removed

### Resources 

https://github.com/Esri/calcite-components/pull/145#pullrequestreview-289404468

```tsx
class={{
  "container": true,
  "container--open": this.open
}}
```

pr3tori4n:
Why do you want to remove the classnames library? It's super useful, and incredibly small.
https://bundlephobia.com/result?p=classnames@2.2.6

driskull:
1. Consistency with `calcite-components`. They won't be using it.
2. We don't need it. We can just use an object and remove a dependency.

driskull:
@pr3tori4n @jcfranco VOTE TIME.

Do we do this or not?

My reasoning for removing this is stated above

- Consistency with `calcite-components` so we don't have to remove its use if we port a component.
- We can do what it does, its just different formatting
  - slightly less clean
  - but doesn't need a dependency or import in the file

jcfranco:
I vote to keep.

+ There's at least one other dependency that `calcite-components` doesn't have that would affect porting (`sortablejs`).
+ Maybe this is just me, but I'd assume any sort of port would require changes.
+ Pretty straightforward to `unclassnamesify`™.
+ Not a lot of porting going on at the moment.
+ 👍 to @pr3tori4n's [comment](https://github.com/Esri/calcite-app-components/issues/269#issuecomment-532787767).

driskull:
Ok! it has been decided

PR #270, opened 2019-09-17
Remove classnames dependency

driskull:
**Related Issue:** #269

## Summary

Remove classnames dependency


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -249,7 +251,14 @@ export class CalciteTipManager {
             <CalciteIcon size="16" path={x16} />
           </calcite-action>
         </header>
-        <div class={classnames(CSS.tipContainer, this.direction)} key={this.selectedIndex}>
```
driskull: This was missed during review but the prop shouldn't double as a classname

Issue #272, opened 2019-09-17
Doc: Pick-list and sub components


kat10140:
Document the pick-list along with pick-list-group, pick-list-item in readme's

PR #273, opened 2019-09-17
Kat10140/pick list doc

kat10140:
**Related Issue:** #272

### DIFF
--- a/src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
+++ b/src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
@@ -13,7 +13,7 @@ export class CalcitePickListGroup {
   // --------------------------------------------------------------------------
 
   /**
-   * The title used for all nested pick-list rows
+   * The title used for all nested calcite-pick-list rows.
    */
   @Prop({ reflect: true }) textGroupTitle: string;
 
--- a/src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
+++ b/src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
@@ -13,7 +13,7 @@ export class CalcitePickListGroup {
   // --------------------------------------------------------------------------
 
   /**
-   * The title used for all nested calcite-pick-list rows.
+   * The title used for all nested calcite-pick-list rows. Not to be confused with the heading of the entire component.
    */
   @Prop({ reflect: true }) textGroupTitle: string;
 
--- a/src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
+++ b/src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
@@ -47,13 +47,24 @@ export class CalcitePickListItem {
       this.emitChangeEvent();
     }
   }
-
+  /**
+   * Indicates what icon the pick-list-item has (or not).
+   */
   @Prop({ reflect: true }) icon: ICON_TYPES | null = null;
 
+  /**
+   * The heading for the pick-list-item.
+   */
   @Prop({ reflect: true }) textHeading: string;
 
+  /**
+   * The description for the pick-list-item.
+   */
   @Prop({ reflect: true }) textDescription: string;
 
+  /**
+   * The value of the pick-list-item.
+   */
   @Prop({ reflect: true }) value: string;
 
   // --------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


pr3tori4n:
@kat10140 This is ready to be re-based. Re-base this onto the hrobbins/valueList-split and and take another pass on the prop docs.

src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
```diff
@@ -13,7 +13,7 @@ export class CalcitePickListGroup {
   // --------------------------------------------------------------------------
 
   /**
-   * The title used for all nested pick-list rows
+   * The title used for all nested calcite-pick-list rows. Not to be confused with the heading of the entire component.
```
pr3tori4n: I think the heading prop for the pick list itself was removed so this second sentence doesn't make sense anymore.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -47,13 +47,24 @@ export class CalcitePickListItem {
...
   @Prop({ reflect: true }) textHeading: string;
 
+  /**
+   * The description for the pick-list-item.
+   */
   @Prop({ reflect: true }) textDescription: string;
 
+  /**
+   * The value of the pick-list-item.
+   */
```
pr3tori4n: Perhaps we can put a note on placement or note why/how it's used. Most of these comments don't provide more info than the prop name at the moment.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -28,30 +28,29 @@ export class CalcitePickList {
 
   /**
    * When true, the items will be sortable via drag and drop.
-   * Only applies when mode is configuration
+   * This only applies when mode is configuration.
```
pr3tori4n: This should be marked deprecated. Can you double check if you merged in latest from master?

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -28,30 +28,29 @@ export class CalcitePickList {
...
 
   /**
-   * Mode controls the presentation of the items in their selected and deselected states.
-   * Selection mode shows either radio buttons or checkboxes depending on the value of multiple
-   * Configuration mode relies on a color highlight on the edge of the item for selected
-   * Mode must be set to configuration for drag and drop behavior to work.
+   * This mode controls the presentation of the items in their selected and deselected states.
+   * Selection mode shows either radio buttons or checkboxes depending on the value of 'multiple'.
+   * Configuration mode relies on a color highlight on the edge of the item for the selected.
+   * For drag and drop behavior to work, configuration mode must be set.
```
pr3tori4n: This should be marked deprecated. Can you double check if you merged in latest from master?

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -28,30 +28,29 @@ export class CalcitePickList {
...
-   * When false, selecting a new item will deselect any other selected items.
+   * When false, only a single item can be selected at a time and selecting a new item will deselect any other selected items.
    */
   @Prop({ reflect: true }) multiple = false;
 
   /**
-   * The heading label for the entire Pick List.
-   * Not to be confused with the heading for an individual item or for a sub-group of items.
+   * The heading label for the entire Pick List
+   * (not to be confused with the heading for an individual item or for a sub-group of items).
```
pr3tori4n: This should be marked deprecated. Can you double check if you merged in latest from master?

src/components/calcite-pick-list/readme.md
```diff
@@ -1,5 +1,7 @@
 # calcite-pick-list
 
+The `calcite-pick-list` component contains `calcite-pick-list-items` stacked in a panel that can be selected or deselected via shift click to be shown on the view.
```
pr3tori4n: The wording here sounds like you have to use shift click. Perhaps add a period after `deselected` and note that shift click works for selecting multiple items at a time (deselecting does not work with shift-click).

pr3tori4n: It's probably also worth calling out the change event that gets fired whenever a selection changed that provides the full list of selections in a JavaScript Map.

src/components/calcite-pick-list-item/readme.md
```diff
@@ -1,5 +1,7 @@
 # calcite-pick-list-item
 
+The `calcite-pick-list-item` is a child element of `calcite-pick-list`. Items are displayed and can be selected or deselected to be shown on the view.
```
pr3tori4n: Let's describe the component a bit more. It will show either a radio button or checkbox, and accepts a secondaryAction slot for custom behavior per item.

Issue #274, opened 2019-09-18
Bug: PickList preselected items aren't sent in the event emitter


pr3tori4n:
### Summary
If a list is configured with pre-selected items, then they are not part of the internal state.
### Actual Behavior
Preselected items don't get added to the list of selected items.
### Expected Behavior
Preselected items get added to the list of selected items on initial load.
### Reproduction Steps

1. Setup a PickList with some of the items having the selected attribute.

pr3tori4n:
@kat10140 please verify.

kat10140:
Verified!

PR #275, opened 2019-09-18
fix: pre-populate state with preselected items

pr3tori4n:
**Related Issue:** #274 

## Summary

pre-selected items in a pickList are now added to the selectedValues map on initial load. This means the list of selectedValues will now be correct when setting pre-selected items on the pickList.


pr3tori4n:
Test Added

Issue #280, opened 2019-09-18
Enhancement: Add a disabled state to PickList Items


pr3tori4n:
#### Desired New Behavior
* Disabled items in the pick-list should not be clickable.
#### Desired Change to Look and Feel
* disabled items in the pick list should gray out
### Relevant Info 
This is a request from Adelheide.

### Helpful Details

**Which Component:** `<calcite-pick-list-item>`
**Example Use Case:** A user wants to allow multi-select, but only one of a type can be selected so they want to disable others.

pr3tori4n:
@kat10140 Please verify.

kat10140:
@pr3tori4n how are the `calcite-pick-list-items` disabled? I don't see a prop or anything

kat10140:
Looks like the state is working, though the 'graying out' isn't.  See gif below @pr3tori4n @asangma 

![pick list item disabled state](https://user-images.githubusercontent.com/48069902/66768331-30d7c500-ee67-11e9-9b67-cf502c514efa.gif)

kat10140:
Nevermind. This is working on the demo page, but not on my local @asangma @pr3tori4n. 
Verified.

kat10140:
Alan will work on the style here https://github.com/Esri/calcite-app-components/issues/361

PR #281, opened 2019-09-19
add disabled prop to pick-list-item

pr3tori4n:
**Related Issue:** #280 

## Summary

A user can now set a list item to be disabled, altering its visual state and preventing selection toggling on that item.


src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -12,6 +12,10 @@
   display: none;
 }
 
+:host([disabled]) {
+  opacity: 0.4;
```
driskull: what about the cursor? does that need to change?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -12,6 +12,10 @@
   display: none;
```
jcfranco: An aside, but his rule gets inherited, no? https://github.com/Esri/calcite-app-components/blob/dfc9ba3aa7e28301e93f1d4825f8b1ad845b7cfd/src/assets/styles/_base.scss#L7-L9

pr3tori4n: oh do I not need this in each component? @jcfranco

jcfranco: I think we get that for free since we have the global styles set up. @asangma can you confirm?

driskull: JC is correct

pr3tori4n: rad

pr3tori4n:
@asangma can you review the disabled styles?

pr3tori4n:
image

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -13,8 +13,13 @@
   background-color: var(--calcite-app-background-hover);
 }
 
-:host([hidden]) {
-  display: none;
+:host([disabled]) {
+  opacity: 0.4;
```
asangma: This should could use 
```
var(--calcite-app-disabled-opacity)
```

pr3tori4n: > One small boulder.

lol that's an oxymoron mang.

pr3tori4n:
@asangma updated please review.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
asangma: That's my style.

Issue #248, opened 2019-09-06
Enhancement: Add loading state and styles - multiple components 


kat10140:
### Description
For following components:
- [ ] action 
- [ ] block 
- [ ] flow
- [ ] flow item
- [ ] shell panel 
- [ ] pick list 
- [ ] value list 

#### Desired New Behavior
Have loading property thats a boolean. False by default. When loading is true, there should be a loading visual. 

#### Desired Change to Look and Feel
Loading visual.

pr3tori4n:
API Design
```js
interface AnyComponent {
  ...
  @Prop({ reflect: true }) loading: boolean = false;
  ...
}
```
```css
:host([loading]) {
  /* loading styles */
}
```

The proposition is to add a `loading` property which should be reflected. And add CSS based on the existence of the loading attribute.

driskull:
API looks good. Will there be any common styles or a funtional component all of these will use?

driskull:
This is related to https://github.com/Esri/calcite-app-components/issues/237 so we need to figure out the right approach.

driskull:
Shouldn't each component that needs it have a loading property that a user can toggle loading on?

Internally, it should use the calcite-loader or calcite-progress in some way.

asangma:
I think yes a prop.
And yes some (possibly all) will use loaders from calcite-components.

The prop will allow for additional steeze as needz.

driskull:
Can we maybe split these up into separate issues and close this one?

Each one of these components will have the same `loading` property but what they do internally will be different between each one. 

Some will use `<calcite-loader>`, some will use `<calcite-progress>` and some may use both. So the designs will be a little different between them.

Maybe...

For following components:
- [ ] action (Matt) 
- [ ] block (JC)
- [ ] flow (Matt) 
- [ ] flow item (Matt) 
- [ ] shell panel  (Matt) 
- [ ] pick list (Harry)
- [ ] value list  (Harry)

Does that sound right?

pr3tori4n:
When @asangma and I last discussed this, I was just going to add the prop and Alan was going to add styling based on that prop as phase 1 of this. What Matt's describing sounds like phase 2.

Is the new plan to skip phase 1 and go straight into phase 2?

driskull:
Seems like phase 1 and phase2 should go together or just adding the prop won't do anything.

I think it's more than styling too. It's altering the DOM with the loading components.

driskull:
[...]

PR #282, opened 2019-09-19
Add Loading State to Components

pr3tori4n:
**Related Issue:** #248 

## Summary
* loading property added.

As discussed after the mapmaker meeting. I'm going to merge these in and we can follow up with adding styling and behavior.


src/components/calcite-action/calcite-action.scss
```diff
@@ -139,3 +139,6 @@
   right: unset;
   left: var(--calcite-app-side-spacing-half);
 }
+
+:host([loading]) {
```
driskull: nitpick: some of these components won't need any host styling, they would just show a loading spinner inside. Maybe we should remove the host styling here.

Issue #268, opened 2019-09-16
Enhancement: PickList Styles


pr3tori4n:
### Description
Add styles for the PickList and it's child components.
### Acceptance Criteria 
- [ ] PickList is styled
- [ ] PickListItem is styled
- [ ] PickListGroup is styled
#### Desired Change to Look and Feel
* Whatever Alan wants.
#### Out of Scope
* ValueList

asangma:
MURGLED!
@kat10140 Assigning to you to for UI QA.

kat10140:
🤙🏼Closing

PR #283, opened 2019-09-19
PickerList styles

asangma:
**Related Issue:** #268 

## Summary
Styles and structural adjustments for 
* PickList
* PickListItem
* PickListGroup

Demo updates with more realistic content.

cc @kat10140


asangma:
Apologies for the big commits.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -160,6 +160,10 @@ export class CalcitePickListItem {
   render() {
     const { icon } = this;
 
+    const description = this.textDescription
+      ? [<span class={CSS.description}>{this.textDescription}</span>]
```
jcfranco: I don't think you need the array.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -20,23 +25,54 @@
...
+  flex-flow: column nowrap;
+  flex: 1 0;
+  overflow: hidden;
+  padding: 0 var(--calcite-app-side-spacing-half);
+  pointer-events: none;
+}
+
+.title {
+  // ------------------------
+  // ellipse after two lines
```
jcfranco: Could this be reused? If so, a (future) mixin could help with reuse and maintainability. It may be hard to distinguish which props/values are for the ellipsis and which ones aren't when updates are made to the declaration.

driskull:
Yeah, i renamed it

src/assets/styles/_mixins.scss
```diff
@@ -0,0 +1,13 @@
+// Mixins
+
+@mixin borderShadow() {
+  box-shadow: 0 1px 0 var(--calcite-app-border);
+}
+
+@mixin lineClamp($lines: 2) {
```
pr3tori4n: Suggest adding a comment with a link where this technique is described.

src/assets/styles/_mixins.scss
```diff
@@ -0,0 +1,13 @@
...
+  display: -webkit-box;
```
pr3tori4n: [This doesn't seem like a standard value for display.](https://developer.mozilla.org/en-US/docs/Web/CSS/display) Is this a kind of hack? Might be working adding a comment.

Should we wrap this in an @supports?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,11 +1,16 @@
...
   align-items: center;
-  padding: 12px;
-  background-color: white;
-  box-shadow: 0px 2px 3px 0px rgba(0, 0, 0, 0.75);
-  margin-bottom: 12px;
+  background-color: var(--calcite-app-background);
+  color: var(--calcite-app-foreground);
   cursor: pointer;
+  display: flex;
+  margin: 0 0 var(--calcite-app-cap-spacing-minimum) 0;
```
pr3tori4n: would margin-bottom work or do we need to zero out the others?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -20,23 +25,48 @@
...
   border-left-color: blue;
   &:host(.calcite--rtl) {
     border-left-color: none;
     border-right-color: blue;
   }
 }
 
+.icon {
+  color: var(--calcite-app-foreground-link);
+  flex: 0 0;
```
pr3tori4n: when using the shorthand `flex` property its a good idea to be explicit with all 3 values because the browsers are inconsistent with their defaults, namely Safari.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -20,23 +25,48 @@
...
+  margin: 0 var(--calcite-app-side-spacing-half);
+}
+
 .label {
-  flex-grow: 1;
-  margin: 0 12px;
-  min-width: 60px;
+  display: flex;
+  flex-flow: column nowrap;
+  flex: 1 0;
```
pr3tori4n: when using the shorthand `flex` property its a good idea to be explicit with all 3 values because the browsers are inconsistent with their defaults, namely Safari.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -20,23 +25,48 @@
...
+  font-family: var(--calcite-app-font-family-monospace);
+  font-size: var(--calcite-app-font-size--1);
+  margin-top: var(--calcite-app-cap-spacing-quarter);
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
+.action {
+  flex: 0;
```
pr3tori4n: when using the shorthand `flex` property its a good idea to be explicit with all 3 values because the browsers are inconsistent with their defaults, namely Safari.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -170,11 +174,11 @@ export class CalcitePickListItem {
         selected={this.isSelected}
       >
         {this.renderIcon()}
-        <div class={CSS.label}>
-          <h4 class={CSS.heading}>{this.textHeading}</h4>
-          <p class={CSS.description}>{this.textDescription}</p>
-        </div>
-        <div onClick={this.secondaryActionContainerClickHandler}>
+        <label class={CSS.label}>
```
pr3tori4n: maybe we should rename the prop from `textHeading` to `textLabel` to be more accurate. This would be a breaking change unless we deprecated but kept backwards compatibility for `textHeading` so I'd suggest we do that.

pr3tori4n:
Nothing blocking, but a few suggestions.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -170,11 +174,11 @@ export class CalcitePickListItem {
...
+          <span class={CSS.title}>{this.textHeading}</span>
```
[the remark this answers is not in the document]
asangma: We aren't really making them bold in this case.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -49,12 +49,6 @@ export class CalcitePickList {
    */
   @Prop({ reflect: true }) multiple = false;
 
-  /**
-   * The heading label for the entire Pick List.
-   * Not to be confused with the heading for an individual item or for a sub-group of items.
-   */
-  @Prop({ reflect: true }) textHeading: string;
```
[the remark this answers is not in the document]
asangma: Understood. I'm not sure anybody other than Adelheid has actually used this component yet tho. right?

src/assets/styles/_mixins.scss
asangma: comment added

src/assets/styles/_mixins.scss
asangma: I think we discussed a while ago that using @supports is mostly about ie11 no?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
asangma: 👍

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
asangma: done

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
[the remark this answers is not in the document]
pr3tori4n: the tags have semantic meaning that can be decoupled from the bold style. If one line has more "strength" or should be styled to be more noticeable in some way I think it fits.

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
asangma: done

src/components/calcite-pick-list/calcite-pick-list.tsx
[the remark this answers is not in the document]
pr3tori4n: yea I'll leave it to the group if we want to let this one slide and not bump major versions.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
asangma: We could do this in a separate issue. Yeah?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
asangma: done

src/components/calcite-pick-list/calcite-pick-list.tsx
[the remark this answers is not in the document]
asangma: I think also any header/heading stuff would be outside of the component going forward anyway.

src/assets/styles/_mixins.scss
pr3tori4n: IE11 is a poster-child use case, but no @supports is browser agnostic. It's a progressive enhancement tool that let's you say (if this is supported in this environment, use it) and you can apply graceful defaults for when it's not.

src/assets/styles/_mixins.scss
pr3tori4n: ty sir

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
pr3tori4n: yea non-blocking. although it could make sense doing it here because this is the PR that's changing the markup and semantic meaning.

asangma: I do agree tho.

asangma:
Thanks y'all! I made most of the suggested changes. Remaining suggestions can be in separate issues. MURGLING!

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
asangma: Added an issue #288

PR #293, opened 2019-09-20
remove undocumented `dismissed` attribute in favor of hidden. A devel…

driskull:
Removes undocumented `dismissed` attribute in favor of hidden.

**Related Issue:** None

## Summary

`dismissed` isn't a custom attribute that is defined on the component so we probably shouldn't use it as one.

A developer can remove hidden if necessary which is an added benefit. Currently, they can't unhide once hidden.

Eliminates some CSS as well.

### DIFF
--- a/src/components/calcite-tip/calcite-tip.scss
+++ b/src/components/calcite-tip/calcite-tip.scss
@@ -15,11 +15,6 @@ $tip-image-max-width: 100% !default;
   border-radius: var(--calcite-app-border-radius);
 }
 
-:host([hidden]),
-:host([dismissed]) {
-  display: none;
-}
-
 :host([selected]) {
   box-shadow: none;
   margin: 0;
--- a/src/components/calcite-tip/calcite-tip.scss
+++ b/src/components/calcite-tip/calcite-tip.scss
@@ -26,6 +26,10 @@ $tip-image-max-width: 100% !default;
   padding-right: 0;
 }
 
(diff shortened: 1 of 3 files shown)


driskull:
The other option is to not mess with the host at all and have a container in the widget to hide its contents. Which may be better 🤷‍♂

jcfranco:
Using `hidden` makes sense to me. According to https://developers.google.com/web/fundamentals/web-components/best-practices#dont-override, modifying the host is fine as long as it honors user-defined attributes.

jcfranco:
Looking into the component a bit, maybe we should hide the component's content instead of using the host's `hidden` attribute. A user could remove hidden and the component would be in a broken state where it considers itself hidden, but can still be displayed.

driskull:
Yep. we can do that but then I have to do it on `connectedCallback` or `componentWillLoad` and its a little messier.

pr3tori4n:
I'm generally not in favor of adding wrapping elements or extra DOM nodes where they are unnecessary. Like @jcfranco mentioned we should be fine to use the Host as a container so long as we're respecting user attributes.

driskull:
> A user could remove hidden and the component would be in a broken state where it considers itself hidden, but can still be displayed.

Yeah. this is the issue with having to keep monitoring the host's hidden attribute when a component is connected/disconnected or the attribute is changed. 

👍 for keeping it internal since it is internal state.

PR #295, opened 2019-09-23
PickList - keyboard toggles

pr3tori4n:
**Related Issue:** #

## Summary

PickListItems can be tabbed to and selected/deselected by pressing "space"

### DIFF
--- a/src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
+++ b/src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
@@ -110,6 +110,14 @@ export class CalcitePickListItem {
     this.emitChangeEvent(event.shiftKey);
   };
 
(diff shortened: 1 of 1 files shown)


src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -110,6 +110,14 @@ export class CalcitePickListItem {
     this.emitChangeEvent(event.shiftKey);
   };
 
+  pickListKeyDownHandler = (event: KeyboardEvent): void => {
+    if (event.key === " ") {
```
driskull: Should we support the enter key too?

pr3tori4n: We could. Usually that submits a form when working with generic components, so I erred on keeping things consistent. Still leaning that way.

Issue #265, opened 2019-09-13
Create calcite-panel component and use it within calcite-flow-item


driskull:
### Description

Create `calcite-panel` component and use it within `calcite-flow-item`

`calcite-panel` component that has the following:

- Named slot for leading header button
- Named slot for header content
- Named slot for trailing header button
- Generic slot for content
- Named slot for footer content

#### Desired New Behavior

- Create `calcite-panel` component that is dismissible
- Remove `calcite-shell-floating-panel
- Use `calcite-panel` within `calcite-flow-item`
- Add a `calcitePanelDismissedChange` event to `calcite-panel`
- Escape key on `calcite-panel` should fire dismissed event

driskull:
@asangma do we want a separate panel for this or would it be better if a user used the `calcite-flow-item` here and put a close button as the menu action?

Just thinking about consistency and how this is different

asangma:
That might be a good idea.
`flow-item`'s footer is optional, but the heading is currently centered. If we can restyle it so that the heading aligns flex-start when it's the first flow, this could work.

asangma:
But then devs would need to know to use a `flow-item` for this case yeah? Seems a bit unintuitive and also could lead to a lot of hand-rolled "floating panels".

driskull:
>  Seems a bit unintuitive and also could lead to a lot of hand-rolled "floating panels".

Yeah. 

Maybe I can just have the two of them share a functional component so they are at least consistent in their design and architecture.

driskull:
Thought about this a little more. I think I want to create a more generic `calcite-panel` component that has the following:

- Header
- Heading
- Named slot for leading header button
- Named slot for trailing header button
- Generic slot for content
- Named slot for footer content

Then the `calcite-flow-item` could use this internally.

For the "floating-panel" we could use the generic `calcite-panel` or we could have a `calcite-dismissible-panel` that uses it internally and adds a close button automatically.

We can discuss at the meeting Friday or here.

asangma:
Do we need the **leading header button**?

driskull:
> Do we need the leading header button?

Yes, for the back button on the flow item.

asangma:
Oooohhh...nice.

driskull:
@asangma installed. Can you double check CSS is good?

asangma:
Whoa...oldie. But a goodie.

PR #296, opened 2019-09-23
Component: CalcitePanel

driskull:
**Related Issue:** #265

## Summary

If this is component-related, please verify that:

- [x] code adheres to the conventions set in `calcite-example` - https://github.com/Esri/calcite-app-components/tree/master/src/components/calcite-example
- [x] changes have been tested with demo page in Edge

Create `calcite-panel` component and use it within `calcite-flow-item`

`calcite-panel` component that has the following:

- Named slot for leading header button
- Named slot for header content
- Named slot for trailing header button
- Generic slot for content
- Named slot for footer content

#### Desired New Behavior

- Create `calcite-panel` component that is dismissible
- Remove `calcite-shell-floating-panel
- Use `calcite-panel` within `calcite-flow-item`
- Add a `calcitePanelDismissedChange` event to `calcite-panel`
- Escape key on `calcite-panel` should fire dismissed event


src/demos/calcite-dismissible-panel.html
```diff
@@ -0,0 +1,41 @@
...
+          see the
+          <a href="https://github.com/Esri/calcite-app-components/tree/master/src/components/calcite-dismissible-panel/readme.md">
+            calcite-dismissible-panel readme.</a
+          >
+        </p>
+
+        <div class="panel-container">
+          <calcite-dismissible-panel heading="test">
+            <p>Content</p>
+          </calcite-panel>
```
pr3tori4n: typo on closing tag

driskull: hey! this isn't ready for review yet

pr3tori4n:
@driskull Seems to me we could make this a single component `<calcite-panel>` with an option to be dismissible.  Thoughts?

src/demos/calcite-dismissible-panel.html
pr3tori4n: oh glad you told me, was about to dive into it :D

driskull:
> Seems to me we could make this a single component <calcite-panel> with an option to be dismissible. Thoughts?

It seems odd to do this because it would mean one of the slots is conditional based on a property. We may not need the dismissible panel if adding the close button isn't too much for a dev to handle using just the `calcite-panel`.

pr3tori4n:
> one of the slots is conditional based on a property

This seems fine to me.

driskull:
> This seems fine to me.

Ok. we'll try it

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -0,0 +1,186 @@
...
+  @Event() calcitePanelDismiss: EventEmitter;
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  dismiss = (): void => {
+    this.calcitePanelDismiss.emit();
```
pr3tori4n: boulder: imo, this should toggle some internal State that hides the panel automatically. I wouldn't want to have to wire up an event handler to close it once I made it dismissible.

src/components/calcite-action/readme.md
```diff
@@ -25,8 +25,8 @@ See the [calcite-action demo](https://esri.github.io/calcite-app-components/demo
 
 - [calcite-action-bar](../calcite-action-bar)
 - [calcite-block-section](../calcite-block-section)
+- [calcite-dismissible-panel](../calcite-dismissible-panel)
```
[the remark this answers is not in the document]
driskull: these are auto generated after merge

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -153,9 +152,9 @@ export class CalciteFlowItem {
     const hasFooterActions = !!this.el.querySelector("[slot=footer-actions]");
 
     return hasFooterActions ? (
-      <footer class={CSS.footer}>
+      <div slot="footer">
```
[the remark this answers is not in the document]
driskull: Its in the panel component.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -0,0 +1,181 @@
...
+  renderHeaderTrailingContent(): VNode {
+    const { dismiss, dismissible, el, textClose } = this;
+
+    const hasHeaderTrailingContent = el.querySelector(`[slot=${SLOTS.headerTrailingContent}]`);
+
+    const contentNode = dismissible ? (
+      <calcite-action aria-label={textClose} text={textClose} onClick={dismiss}>
+        <CalciteIcon size="16" path={x16} />
+      </calcite-action>
+    ) : hasHeaderTrailingContent ? (
```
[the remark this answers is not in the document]
driskull: We could leave it there if both are set.

src/components/calcite-flow-item/calcite-flow-item.scss
```diff
@@ -4,34 +4,12 @@
   font-size: var(--calcite-app-base-font-size);
 }
 
-.container {
+calcite-panel {
```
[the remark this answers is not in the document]
driskull: @asangma can you take the CSS ones? I think these changes need review by you.

src/components/calcite-flow-item/calcite-flow-item.scss
```diff
@@ -44,10 +22,6 @@ h2.heading {
   flex: 1;
```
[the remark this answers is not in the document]
driskull: Alan can you take this?

src/components/calcite-flow-item/calcite-flow-item.scss
```diff
@@ -44,10 +22,6 @@ h2.heading {
...
 }
 
-.heading--first {
-  justify-content: flex-start;
-}
-
 .menu-button {
   flex: 0;
```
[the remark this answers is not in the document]
driskull: Alan can you take this?

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -0,0 +1,50 @@
+:host {
+  display: flex;
+  height: 100%;
```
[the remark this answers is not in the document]
driskull: Alan can you take this?

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -0,0 +1,50 @@
...
+  justify-content: flex-start;
+  position: relative;
+  z-index: 2;
+  background-color: var(--calcite-app-background);
+  border-bottom: 1px solid var(--calcite-app-border);
+  justify-content: flex-start;
+}
+
+.header-content {
+  flex: 1;
```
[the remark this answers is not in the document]
driskull: Alan can you take this?

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -0,0 +1,50 @@
...
+  text-align: center;
+}
+
+.content-container {
+  flex: 1;
```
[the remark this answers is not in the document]
driskull: Alan can you take this?

src/components/calcite-flow-item/calcite-flow-item.scss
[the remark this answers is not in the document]
driskull: Are these really pebbles?? This wasn't even changed in this PR.

driskull:
Addressed review. If the styles are holding it up for a while I will open a separate issue for them.

src/components/calcite-flow-item/calcite-flow-item.scss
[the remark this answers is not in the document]
pr3tori4n: yea forgive me for that. Reason for pebble status is they could result in bugs in some browsers.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -0,0 +1,202 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * Hides the panel.
+   */
+  @Prop({ reflect: true }) dismissed = false;
```
pr3tori4n: Pebble: This should probably be @State() since it's mutating. You can reflect a state value to a prop by setting it on the host `<Host dismissed={this.dismissed}>`

driskull: > This should probably be @State() since it's mutating.

@pr3tori4n Actually, I think I just need to set it as `mutable` since that is what we want. A user will need to be able to remove the dismissed prop to show it again.

Its like an input or a checkbox where a user can modify the value by interacting with the component or by modifying the property.

> You can reflect a state value to a prop by setting it on the host <Host dismissed={this.dismissed}>

Isn't that just a hacky way of making it mutable? Seems like it should just be declared mutable right?

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -0,0 +1,202 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * Hides the panel.
+   */
+  @Prop({ mutable: true, reflect: true }) dismissed = false;
```
driskull: This blog has as good comparison between the internal state vs mutable prop.

https://auth0.com/blog/creating-web-components-with-stencil/#Managing-Mutable-Properties-on-Stencil

The internal state way of doing it just requires more code.

What should we follow for our best practices @jcfranco ?

src/components/calcite-panel/calcite-panel.tsx
pr3tori4n: It's semantics. Everything is an object property in the raw JS language. State is a type of prop that's intended to track changes internal to the component, where as props are more of its options and configuration set from outside.

The React community has a lot of reference on this.
https://github.com/uberVU/react-guide/blob/master/props-vs-state.md

I ran into this when working on the the watcher for the pick-list-item selected prop. My learnings from that were that props should be treated as downward flow, set by the implementer or a higher-order component. Ideally, a component shouldn't mutate it's own props. State should hold dynamic data that changes internally. This split in responsibility helps avoid certain challenges. With the pick-list-item, watching for changes to the `selected` attribute (listening for changes from outside the component) conflicted when I was using the same property for modifying selected state from inside the component. The solution was to have a separate prop and state, and have changes to the prop update the state.

I'd say it's not necessary for internal state changes to reflect back to the DOM. But if we want that to work, it's possible.

In summary, it might work fine in some cases to have mutable props and this might be one of those cases, hence the pebble status. So you don't have to change it. But I do think it's a best practice that will help if the component grows in complexity. Your call.

driskull: > props are more of its options and configuration set from outside.

@pr3tori4n Props aren't just set from the outside in HTMLElements. `<input>`'s `value` is an example of a mutable prop. It can be set from inside as well as outside. This isn't the same as react.

>  props should be treated as downward flow, set by the implementer or a higher-order component. Ideally, a component shouldn't mutate it's own props. 

Where are you getting this from? I couldn't find anything about this with custom elements/web components. I found a lot of the opposite where props are being mutated. Native input elements work this way. 

> I'd say it's not necessary for internal state changes to reflect back to the DOM. But if we want that to work, it's possible.

It is necessary. otherwise how would someone know whether the panel was dismissed already or not? How does someone know whether one of the `pick-list-items` is `selected` already?

> But I do think it's a best practice that will help if the component grows in complexity. Your call.

Why is it a best practice to not use a mutable prop? I think there's some confusion on whether mutable props are acceptable or not. It seems like if the updated value should be exposed then mutable is the way to go.

If I have internal state for this, I wouldn't be able to expose the updated value to the user.

@jcfranco do you have an opinion here?

driskull: Its also worth noting `calcite-components` and `ionic` use a lot of mutable props for stateful props that need to be exposed to the user.

https://github.com/Esri/calcite-components/search?q=mutable%3A&unscoped_q=mutable%3A

https://github.com/ionic-team/ionic/search?q=mutable%3A&unscoped_q=mutable%3A

src/components/calcite-panel/calcite-panel.scss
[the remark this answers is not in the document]
driskull: Created issue for these. https://github.com/Esri/calcite-app-components/issues/303

src/components/calcite-panel/calcite-panel.scss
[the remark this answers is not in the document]
driskull: https://github.com/Esri/calcite-app-components/issues/302

src/components/calcite-panel/calcite-panel.tsx
pr3tori4n: > Where are you getting this from?

As I mentioned, my learning from the experience with the pick-list-item. If you're adding a watcher where you're reacting to a prop being modified from outside a component, then you'll want a separate state property to track internal changes. Outside of that scenario, it may not be necessary

> Why is it a best practice to not use a mutable prop?

This is my opinion, based on my understood definition of `component state` and patterns that exist in other systems. When you're mutating a prop internally, you're treating a prop as state. This violates the single responsibility principle. The prop is being used both as a means of configuration (including default value + configuration modifications) and to track internal state changes. 

Splitting these into two separate properties prevents conflicts from these two operations. One real example I ran into was with the pick-list-item. If there's a watcher on a prop reacting to external changes, it was also undesirably reacting to internal changes. This may very well be how the browser handles things internally.

But it's just a suggestion, and if you're not running into these problems because the complexity isn't there you can continue using a mutable prop if things are functional. I'm not dying on this hill - you're free to move forward with this PR either way you choose.

> <input>'s value is an example of a mutable prop

Input's value attribute and property aren't kept in sync. The value attribute is used to set the initial value of the form. The property then updates with edits from the user, but the attribute doesn't reflect those changes. This is a good example of an attribute being used for configuration but the property being treated as state and only updating internally.
https://javascript.info/dom-attributes-and-properties#property-attribute-synchronization

driskull: > Input's value attribute and property aren't kept in sync.

@pr3tori4n I'm not referring to the `attribute` on an input, that's a different topic that would be handled by the `reflect` option.

I'm referring to the `property`.  The `value`  `property` is updated when a user interacts with the element. It's not just internal state, its a mutable property.

driskull: > this is a good example of an attribute being used for configuration but the property being treated as state and only updating internally.

input value isn't just state. It doesn't just update internally. It can be updated internally or externally.

src/components/calcite-panel/calcite-panel.tsx
jcfranco: Nice find! 

Honestly, I wouldn't want to enforce one way or another. There are benefits of separating state and props, but it *depends on the component*. If it can be done simply and cleanly with prop metadata why not?

Issue #279, opened 2019-09-18
Enhancement: Pick-list and Value-list separation


kat10140:
## Description
We are separating removing the "configuration mode" from *pick-list* and creating *value-list* as a separate component. 

The main reasons for separating capabilities into two separate components are
* to have a cleaner story
  - the two modes can have mutually exclusive capabilities as well as shared capabilities
  - this is confusing for the developer user...and for us
* to allow for greater divergence between the two components
  - the UX differs between selecting items and working with a set of selected items
  - either UX requires different UI language
  - either UX requires different types of information about state to be known

PickList will focus on selecting one or many items from a fixed set of items. It will also allow for a single action that can trigger e.g. more information about an item. Options for this component will be limited to keep its functionality focused. This is a non-editable list. Any sorting or filtering would be done outside of the component and will affect visibility.

ValueList will focus on displaying a list of **selected** items. Each item can be configured by using an accompanying panel and by directly editing the title. Items can also be reordered via drag and drop. Items can be added or removed from this list as well.

### PickList
**PickList criteria**
- Options
  - single (default) or multi select 

**PickListItem criteria**
- Options
  - compact mode
- Each item will have:
  - radio button or checkbox based on single or multi respectively
    - placed flex-start
  - Label
  - Summary (optional)
    - only rendered when a value is passed
  - Slot for an action/icon (right side of card)

---

### ValueList
**ValueList criteria**
- Options
  - drag-enabled
  - label-editable
  - single (default) or multi (select mode will highlight card but will not have checkbox or radio)

**ValueListItem criteria**
- Options
  - label-editable (inheritable from ValueList)
- Each item will have
  - 6-dot drag and drop icon for dragon drop reordering (flex-start)
    - don't render when !drag-enabled
    - icon should be in the new calcite-ui-icons repo
  - Label
    - _label-editable_ can be overridden at item level
    - when in editing, renders an input + commit Action + cancel Action
  - Summary
  - Slot for an action/icon (right side of card)

### Helpful Details
[old pick-list issue](https://github.com/Esri/calcite-app-components/issues/9) just for reference

=================================

## API Design

```js
PickList class {
  // controls multi-select functionality
  @Prop({ reflect: true }) multiple = false;

  // emitted when selecting an item.
  @Event() calciteListSelectionChange: EventEmitter;

  // Used to get selected items outside an event listener
  @Method() async getSelectedItems()
}
```

```js
ValueList class {
  // for enabling drag and drop behavior
[...]

PR #297, opened 2019-09-23
Asangma/valuelist split

asangma:
**Related Issue:** #279

## Summary
Styles and some structural updates.


asangma:
Changes to demo are to ensure styling works for more realistic content.

asangma:
By design.

pr3tori4n:
I would say it might be worth it to try darkening it or something, so you can see the outline of the button.

asangma:
> I would say it might be worth it to try darkening it or something

I could see that. Let's consider it a future enhancement. I'm matching the styles that I ran by Mitch. I think if we see issues with this, we can update the :hover stat at the calcite-action level.

asangma:
MURGLING!

Issue #213, opened 2019-08-23
Bug: Nav bar doesn't span full width on iOS


kat10140:
### Summary
Nav bar doesn't span full width on iOS

### Actual Behavior
![Image from iOS](https://user-images.githubusercontent.com/48069902/63622019-602a2e00-c5aa-11e9-8670-cfadbe98f08d.png)

### Expected Behavior
spaaaaan

### Reproduction Steps

1. Go to https://esri.github.io/calcite-app-components/ on mobile device. Look at nav.

asangma:
@kat10140 I dunno if I'm the one to fix this one.

driskull:
Gonna assign to harry since its his nav design.

pr3tori4n:
@kat10140 Please verify.

kat10140:
@pr3tori4n this doesn't seem to be working in landscape mode, and in portrait, the nav doesn't show up at all

pr3tori4n:
@kat10140 can you let me know what the screen width is in landscape mode? I have the desktop nav beginning once the screen size is 600px. We may need to increase that.

Can you take a screenshot and show me what you're seeing in portrait mode?

kat10140:
@pr3tori4n- iPhone 6s res is 750 x 1334
Here are screenshots of both right now:
![Image from iOS (2)](https://user-images.githubusercontent.com/48069902/65186557-b316ca00-da1e-11e9-8223-6e3f3c42b4d8.png)
![Image from iOS (1)](https://user-images.githubusercontent.com/48069902/65186558-b5792400-da1e-11e9-9837-c542c5d7ece4.png)

kat10140:
@asangma can you comment on this for landscape?  you do have to scroll on landscape. should we just have a hamburger menu for landscape mode as well?

driskull:
Yeah, we need to redesign the nav. We're running out of room :)

pr3tori4n:
@kat10140 please verify.

kat10140:
Verified. Looks great!

kat10140:
it wont let me change labels??

PR #304, opened 2019-09-26
Hrobbins/nav landscape 213

pr3tori4n:
**Related Issue:** #213

## Summary

Updated Mobile nav
Now goes up to 1024px;
New and larger icon.


src/components/calcite-demo-nav/calcite-demo-nav.tsx
```diff
@@ -144,7 +144,9 @@ export class CalciteDemoNav {
   render() {
     return (
       <Host onBlur={this.blurHandler.bind(this)}>
-        <button class="hamburger" onClick={this.hamburgerClickHandler.bind(this)}></button>
+        <button class="hamburger" onClick={this.hamburgerClickHandler.bind(this)}>
```
jcfranco: For consistency, the class should be pushed to CSS object.

src/components/calcite-demo-nav/calcite-demo-nav.tsx
```diff
@@ -144,7 +144,9 @@ export class CalciteDemoNav {
...
+          <CalciteIcon size="32" path={hamburger32} title="Home" />
```
driskull: should we have a `textHome` prop for the title?

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -0,0 +1,103 @@
+import { Component, Element, Event, EventEmitter, Host, Prop, h } from "@stencil/core";
```
driskull: Can you apply the change  from the master branch? Seems like filter stuff is here.

src/components/calcite-demo-nav/calcite-demo-nav.tsx
pr3tori4n: I don't think so - this isn't intended to be used outside of our demo section so this doesn't need to be dynamic.

src/components/calcite-demo-nav/calcite-demo-nav.tsx
```diff
@@ -149,7 +150,9 @@ export class CalciteDemoNav {
   render() {
     return (
       <Host onBlur={this.blurHandler.bind(this)}>
```
jcfranco: Avoid `bind` here since it'll create a new function on each render.

src/components/calcite-demo-nav/calcite-demo-nav.tsx
```diff
@@ -12,7 +12,8 @@ const CSS = {
   isActive: "is-active",
   link: "link",
   open: "open",
-  close: "close"
+  close: "close",
+  hamburger: "hamburger"
```
jcfranco: [Nitpick] `icon` or `toggle` vs `hamburger`?

Issue #276, opened 2019-09-18
Bug: TipManager direction property is being used as a className


driskull:
### Summary

Bug: TipManager direction property is being used as a className

https://github.com/Esri/calcite-app-components/pull/270/files#diff-5920fe4ed9a333d0281e2bb47b7c0e79L252

### Expected Behavior

Class object is used for classnames.

pr3tori4n:
@driskull did you fix this?

driskull:
yes

driskull:
@pr3tori4n I assigned it to you for verification

driskull:
JC can you verify?

jcfranco:
Verified on https://esri.github.io/calcite-app-components/demos/calcite-tip-manager.html. Inspected shadow DOM and direction is no longer applied as a class.

PR #312, opened 2019-09-27
Update stencil to 1.5.3

driskull:
- Update stencil to 1.5.3
- Fix tip manager test (needed key for h2)
- fix issue #276 (added two classes)
- await test `E2EElement.click()` uses

**Related Issue:** #276

### DIFF
--- a/src/components/calcite-tip-manager/calcite-tip-manager.tsx
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.tsx
@@ -80,7 +80,7 @@ export class CalciteTipManager {
 
   @State() direction: "advancing" | "retreating";
 
-  groupTitle = this.textDefaultTitle;
+  @State() groupTitle = this.textDefaultTitle;
 
   observer = new MutationObserver(() => this.setUpTips());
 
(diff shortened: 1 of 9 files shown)


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -236,19 +236,29 @@ export class CalciteTipManager {
...
+    if (total === 0) {
       return <Host />;
     }
     return (
       <Host onKeydown={this.tipManagerKeyDownHandler}>
         <div tabindex="0">
           <header class={CSS.header}>
-            <h2 class={CSS.heading}>{this.groupTitle}</h2>
-            <calcite-action text={this.textClose} onClick={this.hideTipManager} class={CSS.close}>
+            <h2 key={selectedIndex} class={CSS.heading}>
```
jcfranco: Not sure why this is getting the key. `renderPagination` makes that subtree dynamic, so the nodes at that level need a unique key (I would need to test to confirm, but maybe the dynamic node is the only one that needs a key).

driskull: It needs the key because the textContent needs to be updated for the node and it didn't have a way to know that

jcfranco: Follow-up: `key` is being used here to force a re-render of that child. Thanks for the clarification, @driskull!

PR #320, opened 2019-09-30
Add dependency updater script.

jcfranco:
**Related Issue:** N/A

## Summary

Adds NPM script to make dependency updating easier.

Uses https://www.npmjs.com/package/updtr to update all dependencies (runs tests on each one and only updates the passing ones) and commits on success.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -150,6 +150,15 @@
         "@babel/helper-plugin-utils": "^7.0.0"
       }
     },
+    "@babel/runtime": {
+      "version": "7.6.2",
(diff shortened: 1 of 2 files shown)


jcfranco:
Via @pr3tori4n (Slack)

> hey Franco when running npm test, I'm noticing the Jest and Jest-CLI versions are getting auto-bumped back down to 24.8
> Stencil seems to want to control the Jest version

I found an issue related to this: https://github.com/ionic-team/stencil/issues/1479

I'll tweak the updater script to exclude these deps and let the testing that happens when updating dependencies also take care of keeping the lazy loaded dependencies in check.

Issue #307, opened 2019-09-27
Bug: User should not be able to use arrow keys on tip manager navigation buttons


annelfitz:
### Summary
For the navigation buttons in the tip manager, the user should only be able to interact with these buttons using the space bar or the enter key. These should not respond to the arrow keys. 

### Actual Behavior
When focused on the right arrow nav button, if you hit the left arrow key, it will go back. The function of this button should only be to view tips on the right, so we should avoid using arrow keys to interact with these navigation buttons.

### Expected Behavior
Only the space bar and enter key can interact with these buttons.

### Reproduction Steps

1. https://esri.github.io/calcite-app-components/demos/calcite-tip-manager.html
2. Focus on either the left or right arrow buttons of the tip manager, try using the arrow keys to move through the tips. You should only be able to use the space and enter

driskull:
I can take this one

annelfitz:
Looks good!

Issue #306, opened 2019-09-27
Bug: Keyboard Users should be able to scroll content inside of a tip manager


annelfitz:
### Summary
When tabbing through the tip manager, the user should be able to use the up and down arrow keys to scroll through the content of the tip manager. Currently when the overall div of the tip manager is in focus, the up and down and left and right buttons navigate the user through each of the tips, when it should support scrolling. The user should just be able to navigate through the tips from the arrow keys at the bottom of the tip manager.

### Actual Behavior
When using the arrow keys when the tip manager is focused, you are navigated through each of the tips.

### Expected Behavior
When using the arrow keys when the tip manager is focused, you should be able to scroll through the content of a tip.

### Reproduction Steps

1. https://esri.github.io/calcite-app-components/demos/calcite-tip-manager.html
2. Tab through the page to highlight the tipmanager, then use the arrow keys. Will navigate through the tips instead of scrolling through the content

driskull:
I can take this

annelfitz:
Just tested locally and the scroll is still not working where I would expect. When the entire tip manager is highlighted, that's when I would expect to be able to use the arrows to scroll through the content, but this is still not possible. Now it is possible to scroll once the link inside the tip is in focus, but I think the scrolling should be set when the focus is around the whole text, not just a few words
![calicite_tipmanager_scroll](https://user-images.githubusercontent.com/51344676/66140467-06f4e780-e5b7-11e9-82f3-d902a5b4bd18.gif)

annelfitz:
Ok I spoke too soon - this works in Firefox. A highlight appears around the content of the tip and scrolling is good! Not sure why it wasn't working in chrome ^^^
![Screen Shot 2019-10-03 at 8 30 29 AM](https://user-images.githubusercontent.com/51344676/66141195-26d8db00-e5b8-11e9-8477-87c77b41daf1.png)

driskull:
> I think the scrolling should be set when the focus is around the whole text

This isn't something we have control of. The scrollable node is just the one inside the tip manager.

annelfitz:
@driskull In Firefox, a focus outline appears around the scrollable content. Is there a way to do this for the other browsers? Currently (in all browsers but Firefox) the focus outline skips from the X of the tip manager to the link inside of the tip content. If the link is removed, there would be no way for keyboard users to scroll through tip content.

driskull:
Yeah, we can add a tabIndex to it.

driskull:
[...]

PR #321, opened 2019-10-01
#307 key navigation only works on specified target

driskull:
Bug: User should not be able to use arrow keys on tip manager navigation buttons 
Bug: Keyboard Users should be able to scroll content inside of a tip manager

**Related Issue:** 

#307 key navigation only works on specified target
#306 allow scrolling via arrows inside content
Fixes pagination text not updating. Needs key to know to update.

## Summary

Fixes accessibility issues.

### DIFF
--- a/src/components/calcite-tip-manager/calcite-tip-manager.tsx
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.tsx
@@ -187,7 +187,7 @@ export class CalciteTipManager {
     this.nextTip();
   };
 
-  tipManagerKeyDownHandler = (event: KeyboardEvent): void => {
+  tipManagerKeyUpHandler = (event: KeyboardEvent): void => {
     if (event.target !== this.container) {
       return;
     }
@@ -210,6 +210,10 @@ export class CalciteTipManager {
         this.selectedIndex = this.total - 1;
         break;
     }
+  };
+
+  storeContainerRef = (el: HTMLDivElement) => {
+    this.container = el;
   };
 
   // --------------------------------------------------------------------------
@@ -249,11 +253,7 @@ export class CalciteTipManager {
     }
     return (
       <Host>
-        <div
-          tabindex="0"
-          onKeyUp={this.tipManagerKeyDownHandler}
-          ref={(el) => (this.container = el as HTMLDivElement)}
-        >
+        <div tabindex="0" onKeyUp={this.tipManagerKeyUpHandler} ref={this.storeContainerRef}>
           <header class={CSS.header}>
             <h2 key={selectedIndex} class={CSS.heading}>
               {groupTitle}
--- a/src/components/calcite-tip-manager/calcite-tip-manager.tsx
+++ b/src/components/calcite-tip-manager/calcite-tip-manager.tsx
@@ -84,6 +84,8 @@ export class CalciteTipManager {
 
   observer = new MutationObserver(() => this.setUpTips());
 
+  container: HTMLDivElement;
+
   // --------------------------------------------------------------------------
   //
   //  Lifecycle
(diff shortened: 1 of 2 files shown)


driskull:
> The pagination is not working right. Always shows Tip 1/n

The Stencil.js update broke it. Will fix in this PR. Can we get this in before the release? @pr3tori4n

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -225,8 +231,8 @@ export class CalciteTipManager {
         >
           <CalciteIcon size="16" path={dir === "ltr" ? chevronLeft16 : chevronRight16} />
         </calcite-action>
-        <span class={CSS.pagePosition}>
-          {`${this.textPaginationLabel} ${this.selectedIndex + 1}/${this.total}`}
+        <span key={selectedIndex} class={CSS.pagePosition}>
```
jcfranco: I'm wondering if needing this is a Stencil bug. `selectedIndex` should trigger rendering.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -242,8 +248,12 @@ export class CalciteTipManager {
       return <Host />;
     }
     return (
-      <Host onKeydown={this.tipManagerKeyDownHandler}>
-        <div tabindex="0">
+      <Host>
+        <div
+          tabindex="0"
+          onKeyUp={this.tipManagerKeyDownHandler}
+          ref={(el) => (this.container = el as HTMLDivElement)}
```
jcfranco: Can you create a private method for this to avoid creating multiple functions on each render?

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -186,15 +188,17 @@ export class CalciteTipManager {
   };
 
   tipManagerKeyDownHandler = (event: KeyboardEvent): void => {
+    if (event.target !== this.container) {
```
jcfranco: [Suggestion] I think you could avoid `container` by checking if `currentTarget#onkeyup === this.tipManagerKeyDownHandler` or using a data attribute and checking for it. Feel free to ignore.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
driskull: Yeah, it does seem like it should update on render. Do you want to investigate?

jcfranco: That's a Texas-sized 10-4.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: sand: I'd suggest blacklisting the element over whitelisting - this method would prevent the event from correctly firing if it bubbled up from any child elements, and I'm not sure that's ultimately desired.

driskull: I think whitelisting is better than blacklisting. We only want the event to do something if it came from that node.

driskull:
> I'm fine to move forward if this fixes the issue. But I don't understand why the key attribute is needed. The fact that the selectedIndex is used here should trigger Stencil to re-render this. This seems like a Stencil bug we should report.

JC will look into it. We should fix so it works for release.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
driskull: Issue: https://github.com/Esri/calcite-app-components/issues/327

Issue #288, opened 2019-09-19
Enhancement: PickListItem rename textHeading to textLabel


asangma:
### Description
Rename the "textHeading" prop to be "textLabel" in order better match the DOM.

### Acceptance Criteria 
- change the prop name
- retain backward compat. for a bit

#### Desired New Behavior
Name change.

#### Desired Change to Look and Feel
No change.

### Relevant Info 
From PR : https://github.com/Esri/calcite-app-components/pull/283#discussion_r326396675

**Which Component:**
PickListItem

cc @kat10140

pr3tori4n:
@kat10140 please verify.

kat10140:
Verified & closin'

PR #325, opened 2019-10-01
feat: picklistitem rename textHeading prop to textLabel

pr3tori4n:
**Related Issue:** #288 

## Summary

* Renamed `textHeading` to `textLabel`
* Since the picklistitem uses a label now instead of a heading tag, it felt appropriate to rename the prop to more accurately reflect that.
* `textHeading` is backwards compatible but deprecated.


src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -212,7 +218,7 @@ export class CalcitePickListItem {
       >
         {this.renderIcon()}
         <label class={CSS.label}>
-          <span class={CSS.title}>{this.textHeading}</span>
+          <span class={CSS.title}>{this.textLabel ? this.textLabel : this.textHeading}</span>
```
driskull: nitpick: `this.textLabel || this.textHeading` should work & less code :)

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -59,6 +59,7 @@ export class CalcitePickListItem {
   }
 
   /**
+   * @deprecated This will be removed in a future build. Replaced by textLabel.
```
jcfranco: [Nitpick] I'd drop `This will be removed in a future build` since it doesn't add much value and users could track it down with the changelog. Sidebar, we could add a general note about deprecated properties in the README . cc @kat10140

Issue #309, opened 2019-09-27
Bug: Picklist - Secondary action button should not be used to check/uncheck box


annelfitz:
### Summary
The user should only be able to use the space bar to check/uncheck the box when focused on the entire element. When focused specifically on the secondary action button, you should not be able to check/uncheck the box.

### Actual Behavior
When focused specifically on the secondary action button, you can check/uncheck the box by using the space bar

### Expected Behavior
When focused specifically on the secondary action button, you should not be able to check/uncheck the box.

### Reproduction Steps

1. https://esri.github.io/calcite-app-components/demos/calcite-pick-list.html
2. Tab through to focus on the secondary action button in the picklist
3. Hit the space bar - it will check/uncheck the checkbox

annelfitz:
@pr3tori4n @driskull Looks good in chrome, but I'm noticing in firefox and safari that you cannot click on the secondary action. The focus does appear around it when you tab through, but physically clicking it removes the focus (if focus is already around it from tabbing). Otherwise, I am not seeing any indication that I actually clicked the button
![pick_list_no_click](https://user-images.githubusercontent.com/51344676/67504559-730abe80-f63e-11e9-90c2-ce9026bfe1fa.gif)

pr3tori4n:
@annelfitz can you check the console? All the button does in the demo is log out the string 'clicked'.
That should validate the click event is firing. But it sounds like even if that's working the focus styling is not good? @asangma would you be able to take a look at that?

driskull:
Focus styling issue. I think you need to set pointer-events to none on the svg element within.

pr3tori4n:
I agree the styling of the secondary action states is too subtle. I brought this up with @asangma a while back and he said it's by design that the background colors on hover match, but I think it's worth a second look if multiple people are confused by it. @asangma can you weigh in when you get a chance?

pr3tori4n:
@driskull the button in the `calcite-action` shadow dom is what receives focus when it's clicked. Doesn't seem there are any specific :focus styles on the button.
https://github.com/Esri/calcite-app-components/blob/master/src/components/calcite-action/calcite-action.scss#L6-L52

@asangma one solution you might consider is using rgba and adding a little transparency to the hover color, that way when they stack like the aciton on top of a list-item, it should get a little dark naturally.
https://jsfiddle.net/pr3tori4n/oesL13mk/1/
[...]

PR #326, opened 2019-10-01
fix: pickListItem. spacebar on secondaryAction no longer toggles selection

pr3tori4n:
**Related Issue:** #309 

## Summary
pressing spacebar while the secondary action is focused no longer checks/unchecks the list item.


pr3tori4n:
Moving the onclick to the label would severely reduce the hit area. I don't think that's a valid option. 
image
There something wrong with this approach?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -224,7 +221,7 @@ export class CalcitePickListItem {
       >
         {this.renderIcon()}
```
driskull: Seems like this should be inside the `label` element with a role of `checkbox` no? The icon is a replacement for the input checkbox and clicking the label toggles it.

driskull: For example: `<label><input /></label>` => `<label><icon /></label>`

pr3tori4n: I think that's a good call - will look into modifying. @asangma this will require styling changes, I'll ping you once I'm done so you can review.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -59,6 +59,7 @@ export class CalcitePickListItem {
   }
 
   /**
+   * @deprecated This will be removed in a future build. Replaced by textLabel.
```
jcfranco: [Nitpick] We can probably remove this: `This will be removed in a future build. `

pr3tori4n:
@driskull I made the updates you suggested, however I still had to keep the onkeydown event on the host because I was getting a typescript error when I tried putting it on the label. Therefore I also had to keep that condition in the event handler.

driskull:
Ok.

> onkeydown event on the host because I was getting a typescript error when I tried putting it on the label. 

I ran into that issue before and used a `onkeyup` instead. I'm not sure if typescript allows for putting an `onkeydown` on a div.

jcfranco:
That's weird. I don't see why it would be typed to allow one vs the other. Maybe it's a Stencil bug.

driskull:
yeah. I'm not sure either. Maybe some investigation or bug issue for them.

@pr3tori4n would `onKeyUp` work for the use case?

pr3tori4n:
neither onKeydown or onKeyup is passing typescript validation on the label element.

driskull:
> neither onKeydown or onKeyup is passing typescript validation on the label element.

Checked out local branch and this seems to work:

```
<label class={CSS.label} onClick={this.pickListClickHandler} onKeyUp={this.pickListKeyDownHandler}>
```

driskull:
@pr3tori4n Also seeing this: 
image

driskull:
We can merge this in and have a followup issue for the key handler. Lets fix the jsdoc though.

pr3tori4n:
It might be being fickle on the casing. I was using `onKeydown` and `onKeyup` not `onKeyUp`. Follow up issue sounds good tho.

driskull:
Created issue for the handler. https://github.com/Esri/calcite-app-components/issues/433

I think we just need to delete the jsdoc for anything with `@property` and `@type` as it is unnecessary.

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -58,17 +58,18 @@ export class CalciteFilter {
 
   filter = debounce((value: string): void => {
     const regex = new RegExp(value, "ig");
-    const find = (input, regex) => {
+
+    const find = (input: object, RE: RegExp) => {
       let found = false;
-      forIn(input, (value) => {
-        if (typeof value === "function") {
+      forIn(input, (val) => {
```
jcfranco: [Nitpick] Why the rename?

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -1,11 +1,11 @@
 :host {
+  display: flex;
   align-items: center;
+  // justify-content: space-between;
```
jcfranco: Remove commented out code.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -202,17 +202,18 @@ export class CalcitePickListItem {
 
     return (
       <Host
-        onClick={this.pickListClickHandler}
-        onKeydown={this.pickListKeyDownHandler}
         selected={this.isSelected}
+        onKeydown={this.pickListKeyDownHandler}
```
jcfranco: Shouldn't this be `onKeyDown`?

jcfranco:
> It might be being fickle on the casing. I was using onKeydown and onKeyup not onKeyUp. 

It's not being fickle, it's being correct. Look at `dist/types/stencil.core.d.ts` for reference.

src/components/calcite-filter/calcite-filter.tsx
driskull: Because of this tslint rule: https://palantir.github.io/tslint/rules/no-shadowed-variable/

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
driskull: Fixed and moved to label

src/components/calcite-pick-list-item/calcite-pick-list-item.scss
driskull: done

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -201,14 +198,15 @@ export class CalcitePickListItem {
...
-      >
-        <label class={CSS.label} onClick={this.pickListClickHandler}>
+      <Host selected={this.isSelected}>
+        <label
+          class={CSS.label}
+          onClick={this.pickListClickHandler}
+          onKeyDown={this.pickListKeyDownHandler}
+          role="checkbox"
+          aria-checked={this.isSelected}
+          tabindex="0"
```
jcfranco: [Nitpick] We should be using `tabIndex`, no? We can do inventory on this in a follow-up issue.

driskull: Yep. fixed! only found a couple

Issue #205, opened 2019-08-22
Enhancement: Switch mode for BlockSection


asangma:
### Description
Switch mode for BlockSection would result in the same expand/collapse behavior but would remove the triangle on the flex-start side and add a switch to the flex-end side.

**Which Component:**
BlockSection

### Mockups

![image-1](https://user-images.githubusercontent.com/197440/64465426-744d4f80-d0c1-11e9-8e4c-86f35f78fae8.png)

![image](https://user-images.githubusercontent.com/197440/64465425-744d4f80-d0c1-11e9-9f3b-40ed8b358774.png)

jcfranco:
@kat10140 Can you triage this one?

jcfranco:
@asangma Do you have an updated mockup for this?

kat10140:
@asangma

asangma:
Whoop! Let this one slip by.
I'll get you something more explicit than the above.

asangma:
Image on the left shows the section on/open. Image on the right shows the section off/closed.
![Screen Shot 2019-09-18 at 9 50 23 AM](https://user-images.githubusercontent.com/12503298/65168787-1b9f8000-d9fa-11e9-9952-e3b56320e58f.png)

asangma:
Basically it's the same as a normal section except that 
* the graphic is a switchhopefully from calcite-components

jcfranco:
Installed.

kat10140:
Verified! Closing (sorry I forgot to triage @jcfranco 🙃)

PR #334, opened 2019-10-03
Issue #205: Add switch mode to <calcite-block-section />

jcfranco:
**Related Issue:** #205

## Summary

* Sets up `@esri/calcite-components` as a peer dependency.
  - includes distributable from `node_modules` in `docs` output.
* Adds `toggle` property/attribute
* Adds testing util to setup a page w/ peer dependencies loaded
  - **note**: uses hardcoded unpkg link to get it working, but will update as soon as I get some info on how to use `addScriptTag` with `node_modules` (see https://stencil-worldwide.slack.com/archives/CG0TNT3EH/p1570060754020400).

#### API changes

```tsx
interface HTMLCalciteBlockSection {
  // enables switch when true, clickable header if false 
  toggle: boolean;  
}
```

#### Questions

* Do we want to fix the version of `@esri/calcite-components` until it's out of beta?
* Is `toggle` clear enough or would a a type property be clearer?


package.json
```diff
@@ -69,5 +70,8 @@
     "tslint-stencil": "1.0.1",
     "updtr": "3.1.0"
   },
+  "peerDependencies": {
+    "@esri/calcite-components": "1.0.0-beta.10"
```
driskull: Yes lets keep hard coded for now.

src/components/calcite-block-section/calcite-block-section.scss
```diff
@@ -7,3 +7,9 @@
   padding: 0 var(--calcite-app-side-spacing-half) var(--calcite-app-cap-spacing-half);
   animation: calcite-app-fade-in var(--calcite-app-animation-time) var(--calcite-app-easing-function);
 }
+
+.toggle--switch {
+  display: flex;
+  justify-content: space-between;
+  user-select: none;
```
driskull: nitpick: should `user-select:none` be built into the switch? Maybe an issue for them?

src/components/calcite-block-section/calcite-block-section.tsx
```diff
@@ -42,6 +43,16 @@ export class CalciteBlockSection {
...
   textCollapse = TEXT.collapse;
 
+  /**
+   * This property determines the look of the section toggle.
+   * When true, a toggle-switch will be displayed to indicate a capability.
+   * Otherwise, a clickable header is displayed.
+   *
+   * @todo revisit doc
+   */
+  @Prop()
```
driskull: nitpick: should this be reflected?

src/components/calcite-block-section/calcite-block-section.tsx
```diff
@@ -68,28 +79,44 @@ export class CalciteBlockSection {
...
   // --------------------------------------------------------------------------
 
-  onHeaderClick = () => {
+  toggleSection = () => {
     this.open = !this.open;
     this.calciteBlockSectionToggle.emit();
   };
 
+  handleSwitchClick = ({ currentTarget, target }: MouseEvent) => {
+    const isDupeClick = target === currentTarget;
```
driskull: Is this because the whole block is clickable? Should we maybe not have the area around the switch be a hit area for the block?

src/components/calcite-block-section/calcite-block-section.tsx
```diff
@@ -68,28 +79,44 @@ export class CalciteBlockSection {
...
+        <calcite-switch switched={open} onClick={this.handleSwitchClick} />
+      </label>
+    ) : (
       <calcite-action
         aria-label={toggleLabel}
-        onClick={this.onHeaderClick}
-        text={this.text}
+        class={CSS.toggle}
+        onClick={this.toggleSection}
+        text={text}
```
driskull: question: can you just do `text` here?

src/components/calcite-block-section/calcite-block-section.scss
jcfranco: This is for the label wrapping the switch.

src/components/calcite-block-section/calcite-block-section.tsx
jcfranco: 👍

src/components/calcite-block-section/calcite-block-section.tsx
jcfranco: It's caused by the wrapping label. It fires two clicks when you click on the switch. I'm going to open an issue on `calcite-components` about maybe stopping propagation or default behavior when there will be two clicks (it already has some code to check if its wrapped in a label).

src/components/calcite-block-section/calcite-block-section.tsx
jcfranco: I don't follow. You mean as the content of the action element? If so, when do you use one vs the other?

driskull: Nevermind. my bad :)

paulcpederson:
@jcfranco out of curiosity does this duplicate polyfills, etc between app-components and components? I've been wondering about this multiple stencil library use case for a while.

jcfranco:
@paulcpederson Good question. I would have to investigate to confirm, but I believe each component package would get its own loader and respective polyfills.

jcfranco:
@paulcpederson I've posted a question on Stencil's Slack for now: https://stencil-worldwide.slack.com/archives/C79EANFL7/p1570203000159300 🤞

src/demos/head.js
```diff
@@ -25,4 +25,6 @@
   loadCss("build/calcite-app.css");
   loadScript("build/calcite-app.esm.js", { type: "module" });
   loadScript("build/calcite-app.js", { noModule: true });
+  loadScript("node_modules/@esri/calcite-components/dist/calcite/calcite.esm.js", { type: "module" });
```
pr3tori4n: Thinking about the order of load here. Do you foresee any situations where there's something both calcite and calcite-apps are trying to set? If so do you know which one wins priority, the first or the second to do so? And in that case would we want ours or theirs to take priority?

tsconfig.json
```diff
@@ -13,6 +13,5 @@
     "jsx": "react",
     "jsxFactory": "h"
   },
-  "include": ["src", "types/jsx.d.ts"],
-  "exclude": ["node_modules"]
+  "include": ["src", "node_modules/@esri/calcite-components/dist/types/components.d.ts"]
```
pr3tori4n: It possible to exclude all of node modules with this as an exception? Worried what the side-effects of not having node_modules excluded would be.

stencil.config.ts
```diff
@@ -34,7 +34,10 @@ export const config: Config = {
     { type: "docs-readme" },
     {
       type: "www",
-      copy: [{ src: "../demos", dest: "demos" }],
+      copy: [
+        { src: "../demos", dest: "demos" },
+        { src: "../../node_modules/@esri/calcite-components/dist", dest: "node_modules/@esri/calcite-components/dist" }
```
pr3tori4n: nitpick: could simplify the dest directory structure to just "calcite" or something - no need to keep the 4 level directory structure.

pr3tori4n: Do we need to copy over the entire dist folder? We probably don't need the loader, types, esm-es5 and maybe less. Could you try just coyping over `../../node_modules/@esri/calcite-components/dist/calcite` and see if it's depending on any of the other directory content?

src/components/calcite-block-section/calcite-block-section.tsx
pr3tori4n: boulder: Swap from using an `onClick` event to an `onChange` event, and this problem goes away.

src/components/calcite-block-section/calcite-block-section.scss
pr3tori4n: Is this part of the requirements or is it a workaround for something? Not sure why disabling selection on this text is beneficial.

src/components/calcite-block-section/calcite-block-section.e2e.ts
```diff
@@ -15,9 +16,49 @@ describe("calcite-block-section", () => {
...
+    });
+
+    it("renders section text", async () => {
+      const page = await setUpPage(`<calcite-block-section text="test text" open="true"></calcite-block-section>`);
+      const element = await page.find(`calcite-block-section >>> .${CSS.toggle}`);
+      expect(await element.getProperty("text")).toBe("test text");
+    });
+  });
+
+  async function assertContentIsDisplayedAndHidden(page: E2EPage): Promise<void> {
```
pr3tori4n: nice abstraction 👍

src/components/calcite-block-section/calcite-block-section.tsx
```diff
@@ -42,6 +43,18 @@ export class CalciteBlockSection {
...
+   * This property determines the look of the section toggle.
+   * When true, a toggle-switch will be displayed to indicate a capability.
+   * Otherwise, a clickable header is displayed.
+   *
+   * @todo revisit doc
+   */
+  @Prop({
+    reflect: true
+  })
+  toggle = false;
```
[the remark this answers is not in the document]
jcfranco: Yep, this was a question I had in the PR description. I originally wanted a `mode` or `type` prop to describe this, but didn't have a good name for the default (`"header"`?).

src/components/calcite-block-section/calcite-block-section.scss
jcfranco: I disabled this mostly for consistency. The non-switch mode doesn't allow selection. cc @asangma

src/components/calcite-block-section/calcite-block-section.tsx
jcfranco: I did this originally, but had some weird behavior. I'll revisit.

pr3tori4n: I tried it out locally and it seemed to work fine - if you run into it again and get stuck, let me know and I'll help find a solution.

stencil.config.ts
jcfranco: I went with this (simple and gives context)

```js
{ src: "../../node_modules/@esri/calcite-components/dist/calcite", dest: "vendor/@esri/calcite-components" }
```

src/components/calcite-block-section/calcite-block-section.tsx
jcfranco: Fixed. My weirdness came from using `onCalciteSwitchChange`. 🙇

src/components/calcite-block-section/calcite-block-section.tsx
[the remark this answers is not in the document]
driskull: Yeah seems like a configuration on how the header is displayed. Something like `headerDisplay="switch | button"` or `headerToggleDisplay`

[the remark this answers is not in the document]
jcfranco: Building off of ☝️: `toggleDisplay: "button" | "switch"` ?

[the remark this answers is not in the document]
pr3tori4n: yea that works.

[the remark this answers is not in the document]
jcfranco: Renamed! 💥

jcfranco:
@paulcpederson Update on the [Stencil Slack question](https://github.com/Esri/calcite-app-components/pull/334#issuecomment-538447566):


> jcfranco  6 days ago
> Is there any info on bundling strategies when working with multiple component projects? For example, if my app uses foo-components and bar-components, will bundling lead to a single loader and set of polyfills or will I have duplicates?
> 
>   simonhaenisch  6 days ago
> if you
> ```
> import 'foo-components';
> import 'bar-components';
> ```
>   in your stencil project, then the project will consume those components and bundle them along with your project
> 
> simonhaenisch  6 days ago
> that'll work with recent Stencil versions at least
> 
> simonhaenisch  6 days ago
> if you want to use both your component libs just in a plain html/js context via scripts, then yeah you'll have two loaders, one per dist, afaik
> 
> simonhaenisch  6 days ago
> Stencil produces special meta information about your components and puts them in the dist folder, and if you import the dist in another Stencil project, then Stencil knows about all your components and will bundle them along with your project's components

paulcpederson:
nice, thanks @jcfranco So to make sure I understand that correctly, if somebody is using `esri-app-components` and that uses `calcite-components`, they will only have to pull in `esri-app-components` and they'll be able to use both?

jcfranco:
I don't know about `esri-app-components`, but as far as `calcite-app-components` are concerned, there's no bundling going on. `calcite-components` is a peer dependency and is up to whoever is pulling in the component packages to set it up.

The answer I got about smart bundling focuses on the parent project being a Stencil one. I'll follow-up and ask whether that is necessary or what other options are available.

paulcpederson:
sorry I meant calcite-app-components

Issue #338, opened 2019-10-07
Enhancement: PickListItem Event Bubbles enable


pr3tori4n:
### Description
The PickList is stopping propagation of the pickListItem change event. @dhrumil83 is requesting this event be available so he can listen to checked/unchecked of a single item.

#### Desired New Behavior
* stop propagation is removed from PickList and ValueList so child events bubble up.

**Which Component:** `<calcite-pick-list>` `<calcite-value-list>`
**Example Use Case:** User has a multi select pick list, and wants to listen to when is checked/unchecked. Doesn't need to know the full list of selections.

pr3tori4n:
@kat10140 @asangma can you vet this request. Is there any reason we shouldn't do this that I'm missing?

asangma:
I don't see an issue here.

pr3tori4n:
@kat10140 please verify.

kat10140:
Verified!

PR #341, opened 2019-10-07
feat: pickListItem events bubble (remove stopPropagation)

pr3tori4n:
**Related Issue:** #338

## Summary

Currently, the list-item-change events aren't propagated beyond the parent list. This change is to normalize them and allow them to propagate so they can be used by implementers.

* This enables bubbling the event for the list items.
* I also changed the event names so they were consistent with the API design - basically shared event names for both list types. `calciteListSelectionChange` and `calciteListItemChange` respectively.
* Added JSDocs for the events.
* There's a code change where I remove the re-targeting in value-list-item and just check the event.target in valueList and PickList. This normalized behavior between the two for listening to events on the value-list-item and pick-list-item.


pr3tori4n:
@kat10140 can you review the JSDocs in this branch.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -117,6 +129,8 @@ export class CalcitePickList {
       selectedValues.delete(value);
     }
     this.lastSelectedItem = item;
+    this.calciteListSelectionChange.emit(selectedValues);
+    // TODO: remove in the next major version bump.
```
jcfranco: Should we create an issue for stuff going away in a major release? It'd help keep everything in one place and easier to track.

src/components/calcite-pick-list/calcite-pick-list.e2e.ts
```diff
@@ -24,7 +24,7 @@ describe("calcite-pick-list", () => {
         const pickList = await page.find("calcite-pick-list");
         const item1 = await pickList.find("[value=one]");
         const item2 = await pickList.find("[value=two]");
-        const toggleSpy = await pickList.spyOnEvent("calcitePickListSelectionChange");
+        const toggleSpy = await pickList.spyOnEvent("calciteListSelectionChange");
```
jcfranco: Could this be simplified to `calciteListChange` or is that too generic? 

Sidebar, there's no `<calcite-list>` component in the works, right? Just thinking whether there'd be a potential event name clash.

src/components/calcite-pick-list/calcite-pick-list.tsx
driskull: Can we just add a `@deprecated` comment? then we can just search for things that have `@deprecated`

src/components/calcite-pick-list/calcite-pick-list.e2e.ts
driskull: > Sidebar, there's no <calcite-list> component in the works, right?

Not that I know of. Could there be in `calcite-components` at some point?

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: sure. added @deprecated.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -117,6 +129,10 @@ export class CalcitePickList {
       selectedValues.delete(value);
     }
     this.lastSelectedItem = item;
+    this.calciteListChange.emit(selectedValues);
+    /**
+     * @deprecated TODO: remove in the next major version bump.
```
jcfranco: I think this comment is no longer needed since you added `@deprecated` to the emitter.

pr3tori4n: technically not needed either way. Just thought it would be nice to flag this line of code. Ok with keeping it?

Issue #82, opened 2019-07-19
Research: Automated CHANGELOG/BREAKING_CHANGES doc generation


jcfranco:
### Background

For documentation purposes, it would be useful to have an automated way of keeping track of changes/breaking changes.

### Desired Outcome

`CHANGELOG.md` and/or `BREAKING_CHANGES.md` auto generated from, ideally, commit messages.

jcfranco:
Installed. 

Changelog will be autogenerated by picking up any commits between releases that conform to the [`Conventional Commits`](https://www.conventionalcommits.org/en/v1.0.0/#summary) spec. See [PR](#353) for more info.

PR #353, opened 2019-10-09
Issue #82: Auto-generate CHANGELOG

jcfranco:
**Related Issue:** #82

## Summary

Uses `standard-version` to automagically™ generate a changelog during the release process.

The release workflow has been broken up into two steps:

1. `npm run prepare:<major | minor | patch>`
1. `npm run release` (after verifying CHANGELOG is :ok_hand:)

This was done in order to have a chance to review the changelog before publishing.

### Notes

* Commits with messages following [`Conventional Commits`](https://www.conventionalcommits.org/en/v1.0.0/) will be picked up for processing.
  **Example**

  ```
  fix(some-component): honor some property
  ```
* Prefer following the `Conventional Commits` message template for merge commits (be it regular, squashed or rebased).
* `standard-version` could be used in the future to infer the version based on commit messages.

### Questions?

* Do we want anything else in the changelog besides features and fixes? By default, features and fixes are added to the changelog, but we can configure others to be picked up as well (see [supported types](https://github.com/conventional-changelog/conventional-changelog-config-spec/blob/master/versions/2.1.0/README.md#types)).
* Where's the best place to doc this (README, PR template, wiki page)? I'm leaning towards PR template.
* Would `prepare-release` be a more suitable script name than `prepare`? I'm open to suggestions.

### Resources

A few resources to help adhere to the `Conventional Commits` format:

- https://plugins.jetbrains.com/plugin/9861-git-commit-template
- https://www.npmjs.com/package/vscode-commitizen
- https://github.com/commitizen/cz-cli (can be enforced on commit, but also gets on your nerves if you do a lot of WIP commits).

cc @asangma @kat10140


driskull:
> Do we want anything else in the changelog besides features and fixes? By default, features and fixes are added to the changelog, but we can configure others to be picked up as well (see supported types).

Can we allow all of those and just use them if we need to?

What about for breaking changes?

> Where's the best place to doc this (README, PR template, wiki page)? I'm leaning towards PR template.

I would suggest Readme and PR template. :)

> Would prepare-release be a more suitable script name than prepare? I'm open to suggestions.

I like `prepare-release` as it's more explicit. I'm open to `release-the-kraken` as well :)

package.json
```diff
@@ -16,20 +16,20 @@
...
-    "version": "npm run build && npm run docs && git add .",
-    "postversion": "git push && git push --tags",
     "prepublishOnly": "node ./support/prepublish.js",
-    "release:patch": "npm version patch && npm publish",
-    "release:minor": "npm version minor && npm publish",
-    "release:major": "npm version major && npm publish",
-    "deps:update": "updtr --ex @types/jest @types/puppeteer jest jest-cli puppeteer && git add package*.json && git commit -q -m 'chore(deps): Bump versions.'"
+    "prepare:patch": "npm version patch",
+    "prepare:minor": "npm version minor",
+    "prepare:major": "npm version major",
```
pr3tori4n: I don't think we need these scripts. We can just run the standard commands these alias.

pr3tori4n: Are you talking about `preversion` and `postversion` scripts?
running npm version <foo> should still run the `preversion` and `postversion` scripts.

jcfranco: Ah, I misread your comment. Were you suggesting dropping these and just running `npm version <major | minor | patch>`?

Issue #364, opened 2019-10-09
Bug: bg-color refactor broke some lists


asangma:
### Summary
Post #345 pick-list and value-list need bg-color set to transparent.

### Actual Behavior
bg-color is white

### Expected Behavior
bg-color should be transparent.

### Reproduction Steps

1. go to pick-list or value-list demo

### Relevant Info 

cc @kat10140 @pr3tori4n

asangma:
💥

PR #365, opened 2019-10-09
fix(lists): added bg-color rules. 

asangma:
**Related Issue:** #364

## Summary
Added bg-color rules for lists.


asangma:
before
image


after
image

asangma:
For both pick-list and value-list, this allows the parent node to decide whether to show the item shape or not by matching the bg-color or using the light gray bg-color.

Issue #217, opened 2019-08-23
Bug: Block content styling


kat10140:
### Summary
- Padding on 'sectionless' block content.

### Actual Behavior
![63472250-60012580-c426-11e9-8ac8-658752beb42c](https://user-images.githubusercontent.com/48069902/63624260-b8fcc500-c5b0-11e9-890d-f758f680fb73.jpeg)
![63472208-44961a80-c426-11e9-9a72-b341b6d1754b](https://user-images.githubusercontent.com/48069902/63624266-bac68880-c5b0-11e9-9f0e-b41760976416.jpeg)

### Expected Behavior
to be sexy

### Reproduction Steps

1. Go to https://esri.github.io/calcite-app-components/demos/calcite-shell.html. Click on floating panel.  Click on any block heading.

asangma:
@kat10140 INSTARLED!

kat10140:
Lewkin gewd!

PR #376, opened 2019-10-11
General spacing fixes

asangma:
**Related Issue:** #217

## Summary
Started with the padding for Block and quickly realized that this was related to broader spacing issues.

@kat10140 Would you mind checking how generally combined spacing is working together, e.g. when a section is inside a block which is inside a panel, etc. and be another pair of eyes.

I think I got the spacing generally working better.

Also updated some demos to help better test the UI stuff.

### DIFF
--- a/src/components/calcite-block-section/calcite-block-section.scss
+++ b/src/components/calcite-block-section/calcite-block-section.scss
@@ -1,7 +1,7 @@
 :host {
   border-bottom: 1px solid var(--calcite-app-border);
   display: block;
-  margin-bottom: var(--calcite-app-cap-spacing-quarter);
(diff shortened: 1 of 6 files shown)


kat10140:
I think it looks good between heading and summary, padding in collapsible sections in blocks etc. The only thing I see that looks slightly off is the separator between section 2 toggle switch and the last photo in the block. looks a little smush compared to the rest. Not sure if that's really notable enough to bother. 

Screen Shot 2019-10-11 at 3 30 19 PM

asangma:
Thanks! I think the placeholder images may not be good tests for stuffs.
I am slowly updating our demos to use components in the blocks and sections and such so as to get a better idea of the spacing.

PR #378, opened 2019-10-14
Action : Loading & disabled

asangma:
**Related Issue:** #357

## Summary
* Added loading and disabled styles.
* Updated demo app to make it easier to see states.


src/demos/calcite-action.html
```diff
@@ -25,85 +25,17 @@ <h1>calcite-action</h1>
...
-              d="M5.674 13.406a1.498 1.498 0 0 1-.394-.71C3.787 12.335 3 11.757 3 11.5V8.88A9.97 9.97 0 0 0 8 10a12.159 12.159 0 0 0 3.654-.522l1.813-1.813a1.496 1.496 0 0 1 .533-.34V4.499C13.997 3.118 11.312 2 8 2S2.003 3.118 2 4.498V11.5c0 1.11 1.74 2.05 4.144 2.377zM8 3c3.281 0 5 1.113 5 1.5C13 4.885 11.281 6 8 6S3 4.885 3 4.5C3 4.113 4.719 3 8 3zM3 5.88A9.97 9.97 0 0 0 8 7a9.97 9.97 0 0 0 5-1.12v2A9.97 9.97 0 0 1 8 9a9.97 9.97 0 0 1-5-1.12zm3.735 6.466l.738-.738 2.084 2.088 4.97-4.97.738.737-5.708 5.705z"
-            />
-          </svg>
-        </calcite-action>
-
-        <h2>RTL</h2>
-
-        <div dir="rtl">
+        <div class="example-container">
+          <calcite-button onclick="toggleProperty('loading')">Toggle loading</calcite-button>
```
asangma: @jcfranco Think we should make this and the associated function into a demo component.

src/demos/calcite-action.html
```diff
@@ -131,8 +75,51 @@ <h3>Auto text</h3>
...
+                  d="M5.674 13.406a1.498 1.498 0 0 1-.394-.71C3.787 12.335 3 11.757 3 11.5V8.88A9.97 9.97 0 0 0 8 10a12.159 12.159 0 0 0 3.654-.522l1.813-1.813a1.496 1.496 0 0 1 .533-.34V4.499C13.997 3.118 11.312 2 8 2S2.003 3.118 2 4.498V11.5c0 1.11 1.74 2.05 4.144 2.377zM8 3c3.281 0 5 1.113 5 1.5C13 4.885 11.281 6 8 6S3 4.885 3 4.5C3 4.113 4.719 3 8 3zM3 5.88A9.97 9.97 0 0 0 8 7a9.97 9.97 0 0 0 5-1.12v2A9.97 9.97 0 0 1 8 9a9.97 9.97 0 0 1-5-1.12zm3.735 6.466l.738-.738 2.084 2.088 4.97-4.97.738.737-5.708 5.705z"
+                />
+              </svg>
+            </calcite-action>
+          </div>
         </div>
       </section>
     </main>
+    <script>
+      function toggleProperty(property) {
```
asangma: @jcfranco Think we should make this and the associated function into a demo component.

asangma:
cc @kat10140

paulcpederson:
We neglected to do this on `calcite-button` as well, but for these loading states it may not be a bad idea to investigate strategies for making them accessible. IIRC most loaders use `aria-busy="true"` or `role="status"` (most likely with `aria-live="polite"`) to help screen-reader users understand that the action is loading. Sighted users will see the loading indicator, but non-sighted users won't get any feedback that the app is performing an action.

paulcpederson:
We could also possibly bake-in this accessibility into the `calcite-loader` itself?

asangma:
@paulcpederson 

> We could also possibly bake-in this accessibility into the calcite-loader itself?

That'd be totally tubular.

asangma:
totes tubez

src/demos/calcite-action.html
pr3tori4n: what about a utility function? If the main thing is abstract this for re-use that will suffice without having to make a component.

asangma: @pr3tori4n either way. But I think @jcfranco is working on a "settings" panel for demos.

src/components/calcite-action/calcite-action.tsx
```diff
@@ -121,6 +121,8 @@ export class CalciteAction {
           title={labelFallback}
           aria-label={labelFallback}
           disabled={disabled}
+          aria-disabled={this.disabled}
```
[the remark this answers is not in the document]
asangma: updated. Thanks for catching that!

PR #379, opened 2019-10-14
Block : Loading & disabled

asangma:
**Related Issue:** #358

## Summary
* Loading and disabled styles
* Updated demo

cc @kat10140


kat10140:
Lewks gewd to meh!

asangma:
Thanks @kat10140

src/components/calcite-block/calcite-block.tsx
```diff
@@ -156,7 +165,11 @@ export class CalciteBlock {
     const hasContent = !!Array.from(el.children).some((child) => child.slot !== CONTROL_SLOT_NAME);
 
     return (
-      <Host aria-expanded={collapsible ? (open ? "true" : "false") : null}>
+      <Host
+        aria-expanded={collapsible ? (open ? "true" : "false") : null}
+        aria-busy={this.loading}
```
driskull: nitpicks: don't use this. add the props where `collapsible` is added.

asangma: done. 🌮

asangma: ohnvmd 
will fix

asangma:
Updated to use the newly installed Scrim. Please review.

src/components/calcite-block/calcite-block.tsx
```diff
@@ -156,12 +180,19 @@ export class CalciteBlock {
...
-      <article aria-expanded={collapsible ? (open ? "true" : "false") : null}>
-        {headerNode}
-        <div class={CSS.content} hidden={!hasContent || !open}>
-          <slot />
-        </div>
-      </article>
+      <Host
+        aria-expanded={collapsible ? (open ? "true" : "false") : null}
+        aria-busy={loading}
+        aria-disabled={disabled}
```
driskull: does this need the strings "true" "false" like aria-expanded?

driskull: Same with aria-busy

jcfranco: `aria-expanded` needs strings because there's a different meaning when not present:

👇from https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-expanded 

Value | Description
-- | --
true | The element, or another grouping element it controls, is expanded.
false | The element, or another grouping element it controls, is collapsed.
undefined (default) | The element, or another grouping element it controls, is neither expandable nor collapsible; all its child elements are shown or there are no child elements.

A boolean for `aria-busy` is OK because the [default is false](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-busy).

I think we can drop `aria-disabled` since the `disabled` attribute is reflected and has semantic meaning.

src/components/calcite-block/calcite-block.tsx
```diff
@@ -1,8 +1,10 @@
-import { Component, Element, Event, EventEmitter, Prop, h } from "@stencil/core";
+import { Component, Element, Event, EventEmitter, Host, Prop, h } from "@stencil/core";
 import { chevronDown16, chevronUp16 } from "@esri/calcite-ui-icons";
 import { CSS, TEXT } from "./resources";
 import CalciteIcon from "../utils/CalciteIcon";
 import { CalciteTheme } from "../interfaces";
+import { VNode } from "@esri/calcite-components/dist/types/stencil.core";
```
jcfranco: [fixme] Change to `import { VNode } from "@stencil/core/dist/declarations";`

src/components/calcite-block/calcite-block.tsx
```diff
@@ -26,6 +28,11 @@ export class CalciteBlock {
    */
   @Prop() collapsible = false;
 
+  /**
+   * Disabled is used to prevent the action from occurring.
```
jcfranco: This description seems off.

src/components/calcite-block/calcite-block.tsx
```diff
@@ -111,14 +118,30 @@ export class CalciteBlock {
     this.calciteBlockToggle.emit();
   };
 
+  renderScrim(): VNode {
```
jcfranco: [fixme] Drop method since it's not used.

src/components/calcite-block/calcite-block.tsx
jcfranco: [fixme] BTW, this is reverting https://github.com/Esri/calcite-app-components/commit/58a8f85b81db3362cf90f58bc944e97721200370#diff-5143375a6c30f688a32734e78d4fb3bc.

src/components/calcite-block/calcite-block.tsx
```diff
@@ -127,7 +150,8 @@ export class CalciteBlock {
           <h3 class={CSS.heading}>{heading}</h3>
           {summary ? <div class={CSS.summary}>{summary}</div> : null}
         </div>
-        <slot name={CONTROL_SLOT_NAME} />
+        {loading || disabled ? null : <slot name={CONTROL_SLOT_NAME} />}
```
jcfranco: I think this can be simplified to:

```tsx
const content = loading ?
      <calcite-loader inline is-active></calcite-loader> :
      !disabled ? <slot name={CONTROL_SLOT_NAME}/> :
        null;
```

asangma:
Thanks y'all! I'll get after it!

src/components/calcite-block/calcite-block.tsx
asangma: So did I need to do anything here other than remove `aria-disabled`?

jcfranco: That and moving the attributes back to the `<article>` element (see the diff above).

jcfranco:
This will need a test too. If pressed for time, create an issue and I'll help out.

src/components/calcite-block/calcite-block.tsx
asangma: COOL.

asangma: Also, to be clear, we don't need `aria-disabled` for accessibility because the `disabled` attribute does the job.

asangma: Please have another peek.

src/components/calcite-block/calcite-block.tsx
asangma: Please have another peek.

asangma:
Thanks y'all!

Issue #359, opened 2019-10-09
Enhancement: CalciteFlowItem - add loading & disabled properties


kat10140:
### Description

Enhancement: CalciteFlowItem add loading & disabled properties. 
CalciteFlowItem will then pass down props to CalcitePanel. 
All styles for CalcitePanel will live in the panels scss file.

### Acceptance Criteria 

Loading and disabled boolean properties on the component.

#### Desired New Behavior

- Loading should display a visual loading state on the component.
- Disabled should display a visually disabled state on the component and prevent user interaction as necessary.
- Loading should use any appropriate `calcite-components` for loading as necessary.

#### Desired Change to Look and Feel

Loading style and disabled style on the item, passed down to panel

### Relevant Info 

Related to #248

### Helpful Details

@asangma you can help me figure out what exactly needs to be blocked and what the ideal design should look like.

asangma:
Installed. @kat10140 
BYER BYER BYEEEEEER...

kat10140:
I see props on here, but no demo! I made a separate issue for that @407

PR #380, opened 2019-10-14
feat(flow-item, panel): added loading and disabled

asangma:
**Related Issue:** #359

## Summary
* props
* nodes
* styles


asangma:
Hold for #381

asangma:
Hold for #381

src/demos/calcite-flow.html
```diff
@@ -51,8 +51,14 @@ <h1>calcite-flow</h1>
             margin: 1rem 0.2rem;
             min-width: 5rem;
           }
+          .toggles {
+            margin: 1rem 0;
+          }
```
pr3tori4n: should be able to move this to `src/demos/demos.css` to share across all demo files.

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -49,3 +49,18 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background-color: var(--calcite-app-scrim);
+  z-index: 2;
```
pr3tori4n: curious if this might be lower than slotted content with z-index. Might need to raise this value.

asangma:
Please review again.
Using Scrim now.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -177,13 +188,19 @@ export class CalcitePanel {
...
+
   render() {
-    const { dismissed, dismissible, el, panelKeyUpHandler } = this;
+    const { disabled, dismissed, dismissible, el, loading, panelKeyUpHandler } = this;
 
     const rtl = getElementDir(el) === "rtl";
 
     return (
-      <Host>
+      <Host aria-busy={loading} aria-disabled={disabled}>
```
driskull: do booleans work here or does it need text string "true/false"?

src/demos/calcite-flow.html
asangma: I think we could but I also think this will get stripped out when we use the component @jcfranco is cooking up.

src/components/calcite-panel/calcite-panel.tsx
asangma: Seems to be working. I can 2xcheck.

asangma: Yeah...it works.

asangma: @kat10140 can you dubz checkz?

kat10140: it twerks @driskull

asangma: Dah...I need to move those attributes out of Host.
Also, I'll remove the `aria-disabled` attribute like we did in #379.

asangma: Thanks @kat10140

Issue #361, opened 2019-10-09
Enhancement: CalcitePickList - add loading & disabled properties


kat10140:
### Description

Enhancement: CalcitePickList add loading & disabled properties

### Acceptance Criteria 

Loading and disabled boolean properties on the component.

#### Desired New Behavior

- Loading should display a visual loading state on the component.
- Disabled should display a visually disabled state on the component and prevent user interaction as necessary.
- Loading should use any appropriate `calcite-components` for loading as necessary.

#### Desired Change to Look and Feel

Loading style and disabled style

### Relevant Info 

Related to #248

### Helpful Details

@asangma you can help me figure out what exactly needs to be blocked and what the ideal design should look like.

asangma:
@kat10140 Installed.

kat10140:
Verified!

PR #381, opened 2019-10-14
Loading disabled scrim with pick list

asangma:
**Related Issue:** #357 #358 #359 #361 #362

## Summary
Adding a general purpose "scrim" component for disabled and loading states.
I've written the same thing three times, so seems like it could be a thing.


src/components/calcite-scrim/calcite-scrim.tsx
```diff
@@ -0,0 +1,52 @@
+import { Component, Host, Prop, h } from "@stencil/core";
```
driskull: fixme: Does this need to be a component? I'm wondering if we want this to be a public component that people could use?

It might make more sense as a functional component.

src/components/calcite-scrim/calcite-scrim.tsx
```diff
@@ -0,0 +1,52 @@
...
+})
+export class CalciteScrim {
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * When true, scrim adds the aria-disabled attribute
```
driskull: These properties are a little odd because they mention aria.

src/components/calcite-scrim/calcite-scrim.tsx
asangma: @driskull I'll defer to your judgement. Dunno if it makes a difference, but this is likely to be transferred to calcite-components before long.

src/components/calcite-scrim/calcite-scrim.tsx
asangma: I could use some advice on this.
My thought was that if the user wants to handle `aria-busy` or `aria-disabled` on their own, they could turn the scrim's attributes off. But it'd be nice to have the scrim come with these by default.

asangma: @driskull Actually, I am gonna remove all the aria busy and disabled stuff and leave it to the user to place those attributes.

driskull: ok sounds good.

src/components/calcite-scrim/calcite-scrim.tsx
driskull: I'm just thinking this might be a candidate for a functional component because it wouldn't be used on its own. It would always be used within another component, right? Also, we don't want a user to use this component right?

asangma: If it's a FunctionalComponent, will I need to add its styles to **_base.scss**?

asangma: I think may also need to have the states `loading` and `disabled`.

driskull: Probably a mixin that components could share/import. Let's talk about this component before we decide to make it a functional component. Maybe a quick meeting or wait till friday?

asangma: I'll be in training on Friday.

pr3tori4n: I definitely support making this stuff more DRY. 💯 
I agree that we should explore other ways to abstract this code. A functional component may be best. Mixins might also work.

pr3tori4n: > If it's a FunctionalComponent, will I need to add its styles to _base.scss?

Yea you'll need to add either a mixin or a class to the global styles for it to be available across components. Double-checked and it doesn't seem a FunctionalComponent supports having it's own CSS, unless you inline it.

pr3tori4n:
@asangma I'd suggest splitting this into two branches and two PRs. One branch/PR to abstract the scrim, and the second PR to implement it in the pick-list. You can have the pick-list branch be based on the scrim abstraction branch until that one is merged in to master.

asangma:
> I'd suggest splitting this into two branches and two PRs

Thanks @pr3tori4n. I am keeping the pick-list stuff in there to make sure it all works. Not sure it makes a huge difference...?

asangma:
@driskull Have a look at the refactor. Based on @jcfranco's suggestion, I put the styles inline.

asangma:
Mind if I murgle?

driskull:
Murgle!

Issue #292, opened 2019-09-20
Enhancement: CalciteBlock host element `aria-expanded`


driskull:
### Description

Extremely minor: the `aria-expanded` attribute is being set on the host.

If we are going to set it on the host, we should make sure that it's not being set by the user.

The other option would be to move it to the `article` internal to the component.

jcfranco:
I'm not sure this is valid. `aria-expanded` is set on the host based on the state of the component. The user shouldn't be setting it and they are, the component will override so that the value is correct.

Ionic components do this too: https://github.com/ionic-team/ionic/blob/8a88dd25b66e72f30d2d3325ac3d3abe53396f74/core/src/components/checkbox/checkbox.tsx#L150-L154

driskull:
Ok. I'm fine with leaving this.

My suggestion was to just move this to the internals of the component so a user doesn't see it on the custom element and doesn't have the ability to change it.

jcfranco:
Do you know how we can test if both approaches give the same a11y hints to screen readers? The Accessibility tab from Chrome Dev Tools only reports `aria-expanded` when its on the host.

driskull:
I know the `aXe` toolbar supports shadowDom but I'm not sure if the built in `accessibility` tab does.

jcfranco:
Based on testing (axe) and a few resources I found (see below), there doesn't seem to be an issue w/ either approach, but I do kind of prefer hiding these details from the user if possible.

https://archive.substantial.com/blog/2014/02/05/accessibility-and-the-shadow-dom/
https://developer.paciellogroup.com/blog/2012/07/notes-on-web-components-aria/
http://marcysutton.github.io/accessibility-of-web-components/slides.html#/slide1

PR #389, opened 2019-10-15
Issue #292: Move block a11y attributes to shadow DOM.

jcfranco:
**Related Issue:** #292 

## Summary

Moved a11y attrs down to shadow DOM in favor of obfuscating implementation details.

### DIFF
--- a/src/components/calcite-block-section/calcite-block-section.tsx
+++ b/src/components/calcite-block-section/calcite-block-section.tsx
@@ -1,4 +1,4 @@
-import { Component, Element, Event, EventEmitter, Host, Prop, h } from "@stencil/core";
+import { Component, Element, Event, EventEmitter, Prop, h } from "@stencil/core";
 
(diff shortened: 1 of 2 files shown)


src/components/calcite-block-section/calcite-block-section.tsx
```diff
@@ -119,12 +119,12 @@ export class CalciteBlockSection {
       );
 
     return (
-      <Host aria-expanded={open ? "true" : "false"}>
+      <section aria-expanded={open ? "true" : "false"}>
```
pr3tori4n: nitpick: There is a non-zero cost to adding extra DOM elements. 
https://gtmetrix.com/reduce-the-number-of-dom-elements.html
Applications like the mapmaker will see a potential performance benefit from anywhere we can trim them.

Let's just be sure the gains outweigh the losses from something like this.

jcfranco: Noted, although I'm not too worried about this at this point.

PR #397, opened 2019-10-17
Kat10140/pick and value list doc

kat10140:
**Related Issue:** #272 
I keep forgetting to put the issue number in my commits 😒. Please remind me and I will try and be better about that.

## Summary
DAWKS


jcfranco:
> I keep forgetting to put the issue number in my commits

There is a way we could automate this by leveraging githooks + branch naming convention. 😈

PR #405, opened 2019-10-19
feat(value-list): loading and disabled states

asangma:
**Related Issue:** #362

## Summary
Using Scrim for loading and disabled states.

cc @jcfranco can you suggest a good way for me to not put attributes on Host in this case as well as in pick-list. I can add a wrapper `div` inside Host perhaps and put the attributes on that?


src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -265,21 +276,29 @@ export class CalciteValueList {
...
+    return this.loading || this.disabled ? (
+      <CalciteScrim loading={this.loading}></CalciteScrim>
+    ) : null;
+  }
+
   render() {
+    const { dataForFilter, handleFilter, disabled, filterEnabled, loading } = this;
     return (
-      <Host>
+      <Host aria-busy={loading}>
```
asangma: I want to move this attr out of Host.
Would a `div` be an appropriate wrapper inside Host?
@jcfranco @pr3tori4n @driskull

driskull: Yes, that would be sufficient.

asangma: Thanks!

Issue #407, opened 2019-10-21
Fix: panel styling and flow-item demo with loading and disabled states


kat10140:
### Description

Update calcite-flow-item demo with disabled and loading states. 
Minor calcite-panel .scss fix to accommodate those toggles.

asangma:
cool cool cool

kat10140:
@asangma  can you verify dis pls?

asangma:
VERFEED!

PR #408, opened 2019-10-21
DONT MERGE! flow-item demo

kat10140:
**Related Issue:** #407 

## Summary
Don't merge! PR for Alan to check out diff on flow-item demo.  Currently toggling the entire window instead of the component.


kat10140:
created new pr with branch name change

Issue #418, opened 2019-10-21
Enhancement: Remove padding from panel content-container


asangma:
### Description
Padding in panel's `content-container` has padding that probably should be removed.

### Acceptance Criteria 
No padding.

#### Desired New Behavior
No padding.

#### Desired Change to Look and Feel
No padding.

#### Out of Scope

**Which Component:** Panel

cc @kat10140

asangma:
<img width="476" alt="Screen Shot 2019-10-21 at 1 11 39 PM" src="https://user-images.githubusercontent.com/12503298/67240014-67808300-f405-11e9-8032-a58f8e274d79.png">

asangma:
<img width="360" alt="Screen Shot 2019-10-21 at 1 20 04 PM" src="https://user-images.githubusercontent.com/12503298/67240099-9c8cd580-f405-11e9-8e22-cdb0e7073dee.png">

asangma:
cc @AdelheidF

asangma:
verified

PR #434, opened 2019-10-22
style(panel, value-list): spacing tweaks/

asangma:
**Related Issue:** #418

## Summary
When value-list was in a panel or when pick-list was in a panel, the padding in panel was causing issues.


asangma:
Thanks @driskull

Issue #427, opened 2019-10-22
Bug: CalciteBlock header color is too dark for the dark theme


driskull:
### Summary

Bug: CalciteBlock header color is too dark for the dark theme

### Actual Behavior

The text color is black on dark gray

### Expected Behavior

The text color should be lighter.

### Reproduction Steps

1. See the shell demo page for dark theme.

### Relevant Info 

![image](https://user-images.githubusercontent.com/1231455/67314567-bab40d80-f4b9-11e9-84dc-8d0674b31692.png)

asangma:
@driskull Did I need to verify this?

asangma:
LOL

driskull:
done

PR #437, opened 2019-10-22
CalciteBlock: Set color vars for dark theme.

driskull:
**Related Issue:** #427

CalciteBlock: Set color vars for dark theme.

### DIFF
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -21,8 +21,6 @@ calcite-loader[inline] {
 }
 
 .header {
-  color: var(--calcite-app-foreground);
-  fill: var(--calcite-app-foreground);
   flex-grow: 1;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/calcite-block/calcite-block.scss
```diff
@@ -21,6 +21,8 @@ calcite-loader[inline] {
 }
 
 .header {
+  color: var(--calcite-app-foreground);
```
asangma: We can probably move
```
  color: var(--calcite-app-foreground);
  fill: currentColor;
```
into **_base.scss**

src/components/calcite-block/calcite-block.scss
```diff
@@ -45,6 +47,8 @@ calcite-loader[inline] {
   margin: 0;
   text-align: unset;
   width: 100%;
+  color: var(--calcite-app-foreground);
+  fill: var(--calcite-app-foreground);
```
asangma: Might only need `color` since the SVG is using `currentColor`.

PR #448, opened 2019-10-23
feat: Add summary property to calcite-flow-item

driskull:
**Related Issue:** #450

## Summary

Add summary property to calcite-flow-item


AdelheidF:
Issue: https://github.com/Esri/calcite-app-components/issues/450

driskull:
Thanks @AdelheidF!

@asangma this just needs CSS love.

asangma:
@driskull Do you think we should do this on Panel instead?

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -52,6 +53,11 @@ export class CalciteFlowItem {
    */
   @Prop() showBackButton = false;
 
+  /**
+   * Summary text.
```
pr3tori4n: pebble: Needs better description. Curious what the subtle difference between heading and summary are, as a summary can act as a heading like it does in our PRs sometimes.

src/components/calcite-flow-item/calcite-flow-item.e2e.ts
```diff
@@ -67,6 +67,16 @@ describe("calcite-flow-item", () => {
...
   });
 
+  it("should have default summary", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent('<calcite-flow-item summary="test"></calcite-flow-item>');
+
+    const element = await page.find(`calcite-flow-item >>> .${CSS.summary}`);
+
+    expect(element).toEqualText("test");
```
pr3tori4n: Suggestion to move the string to a TEXT content in resources, so updates to components keep the tests in sync.

driskull:
> Do you think we should do this on Panel instead?

@asangma we could, if we want to restrict what users put into a panel as having a heading/summary in the middle and nothing else. Do you want to go that route?

src/components/calcite-flow-item/calcite-flow-item.e2e.ts
driskull: Its local to this file so it wouldn't get out of sync.

asangma:
@driskull Based on our conversation and the fact that Adelheid doesn't need this anytime soon, do we want to move it to the next milestone?

driskull:
Yup.  moved

driskull:
Alan this just needs styling I believe.

driskull:
@asangma merged in latest master, resolved conflicts. Closing this PR. Can you just keep yours open and target master?

Issue #439, opened 2019-10-22
Bug: CalcitePickList - secondary action gets pushed outside container


driskull:
### Summary

Bug: CalcitePickList - secondary action gets pushed outside container

### Actual Behavior

Bug: CalcitePickList - secondary action gets pushed outside container

### Expected Behavior

Bug: CalcitePickList - secondary action stays within its container and text is either cut off or spans a second line.

### Reproduction Steps

1. See the calcite-pick-list demo

### Relevant Info 

![image](https://user-images.githubusercontent.com/1231455/67339789-bcdd9280-f4e0-11e9-8a59-1f547b8b8de4.png)

asangma:
Installed.

kat10140:
Yay good job Alern!

PR #458, opened 2019-10-24
fix(pick-list-item): summary text overflow fix

asangma:
**Related Issue:** #439 

## Summary
* fixed summary (`text-description`) overflow issue
* also cleaned up some SASS


asangma:
@pr3tori4n Naw...that letter orphans are a separate thing. But yeah.

Issue #456, opened 2019-10-24
Bug: value-list-item. Remove extra padding


asangma:
### Summary
Padding got added to `pick-list-item` >  `label`. 
`value-list-item` adds more padding.

### Actual Behavior
Too much padding.

### Expected Behavior
Less padding.

### Reproduction Steps

1. see demo

### Relevant Info 
#390

kat10140:
Cha ching

PR #459, opened 2019-10-24
fix(value-list-item): removed redundant padding

asangma:
**Related Issue:** #456

## Summary
Remove redundant padding.
This PR refrains from some needed styling that is coming in #458


asangma:
Thanks @kat10140

Issue #404, opened 2019-10-18
Enhancement: Block - control slot needed before the heading


kat10140:
### Description
We need a named slot that can be rendered _before_ the heading of a Block (we have one for the right corner right now). 

This will primarily be used for dragon dropz functionality by Dhrumil. (I added 'named' slot since we are talking about possibly having a dragon drop component that uses the dnd icon for blocks when the named slot meets the reqs) -  we can talk about this in the next meeting.

### Acceptance Criteria 
- Slot before the heading/title of Block component

#### Desired New Behavior
Slawt

#### Desired Change to Look and Feel
Slawt

jcfranco:
Installed.

PR #462, opened 2019-10-24
Issue #404: Add icon slot for calcite block header .

jcfranco:
**Related Issue:** #404

## Summary

Adds slot (name => `"icon"`) for trailing header icon.

@asangma Can you review the spacing between the icon and header? It may need some tweaking.

### DIFF
--- a/src/components/calcite-block/resources.ts
+++ b/src/components/calcite-block/resources.ts
@@ -15,3 +15,8 @@ export const TEXT = {
   collapse: "Collapse",
   expand: "Expand"
 };
+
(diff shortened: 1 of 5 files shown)


src/components/calcite-block/resources.ts
```diff
@@ -15,3 +15,8 @@ export const TEXT = {
   collapse: "Collapse",
   expand: "Expand"
 };
+
+export const enum SLOTS {
```
pr3tori4n: 👍

Issue #403, opened 2019-10-18
Enhancement: Block - remove collapsible icon when there's a control


kat10140:
### Description
As it is now, we have a collapsible property that gives us the icon in the top right corner of the block header. 

![Screen Shot 2019-10-18 at 3 57 44 PM](https://user-images.githubusercontent.com/48069902/67133079-16338200-f1c0-11e9-8793-7b922691aab6.png)

We also have a slot for a calcite action which would render a [..] icon. We don't want both appearing.
 We need a ternary probably in here:

```
const headerContent = (
      <header class={CSS.header}>
        <div class={CSS.title}>
          <h3 class={CSS.heading}>{heading}</h3>
          {summary ? <div class={CSS.summary}>{summary}</div> : null}
        </div>
        {content}
      </header>
    );
``` 

to check for the control and if so, hide the collapsible icon SO IT DUZN'T LOOK DUMB WITH 2 ICONS

#### Desired Change to Look and Feel
No collapsible icon when there is a [...] control 😬

kat10140:
verified

PR #468, opened 2019-10-25
Issue #403: Make block collapsible or display control

jcfranco:
**Related Issue:** #403 

## Summary

This makes collapsible blocks and blocks w/ a control mutually exclusive. 

**Note**: this PR should be reviewed/installed after #404 is merged.

### DIFF
--- a/src/components/calcite-block/calcite-block.e2e.ts
+++ b/src/components/calcite-block/calcite-block.e2e.ts
@@ -104,24 +104,26 @@ describe("calcite-block", () => {
       expect(summary.innerText).toBe("test-summary");
     });
 
-    it("supports a nested control", async () => {
-      const page = await newE2EPage();
-      await page.setContent(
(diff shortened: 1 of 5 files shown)


asangma:
Sorry if I misunderstand, but we still want blocks to have the collapsible capability when there's a control. We just wanna exclude the caret icon when there's a control.

jcfranco:
@asangma Gotcha. Won't that be a little confusing for users?

asangma:
For example, this block should still be expandable and collapsible.
image

asangma:
> Won't that be a little confusing for users?

When testing static mockups, we've seen users assuming that tapping/clicking on the header of the block would either expand it or "open" it in some way.

But I think we'll have full confidence once it's in the app...specifically in some of the configuration panels that Dhrumil is working on.

jcfranco:
Alrighty, I'll make some updates to this PR then and let you know when it's ready to review.

giphy

PR #476, opened 2019-10-28
Issue #403: Hide collapsible icon when control is provided.

jcfranco:
**Related Issue:** #403

## Summary

This hides the collapsible icon (downward/upward chevron) whenever there is a control.

### DIFF
--- a/src/components/calcite-block/calcite-block.e2e.ts
+++ b/src/components/calcite-block/calcite-block.e2e.ts
@@ -104,24 +104,47 @@ describe("calcite-block", () => {
       expect(summary.innerText).toBe("test-summary");
     });
 
-    it("supports a nested control", async () => {
-      const page = await newE2EPage();
-      await page.setContent(
-        `<calcite-block heading="test-heading" collapsible><input class="nested-control" slot=${SLOTS.control} /></calcite-block>`
-      );
-      const element = await page.find("calcite-block");
-      const elementToggleSpy = await element.spyOnEvent("calciteBlockToggle");
+    describe("adding a header control", () => {
+      it("allows users to add a control in a collapsible block", async () => {
+        const page = await newE2EPage();
+        await page.setContent(
+          `<calcite-block heading="test-heading" collapsible><input class="nested-control" slot=${SLOTS.control} /></calcite-block>`
+        );
+        const control = await page.find(".nested-control");
+        expect(await control.isVisible()).toBe(true);
 
-      const control = await element.find(".nested-control");
-      expect(await control.isVisible()).toBe(true);
-      expect(await control.getProperty("slot")).toEqual(SLOTS.control);
+        const controlSlot = await page.find(`calcite-block >>> slot[name=${SLOTS.control}]`);
+        expect(await controlSlot.isVisible()).toBe(true);
 
-      await control.click();
-      expect(elementToggleSpy).toHaveReceivedEventTimes(0);
+        const collapsibleIcon = await page.find(`calcite-block >>> .${CSS.toggleIcon}`);
+        expect(collapsibleIcon).toBeNull();
 
-      await element.click();
-      await element.click();
-      expect(elementToggleSpy).toHaveReceivedEventTimes(2);
+        const block = await page.find("calcite-block");
+        const blockToggleSpy = await block.spyOnEvent("calciteBlockToggle");
+
+        expect(await control.isVisible()).toBe(true);
+        expect(await control.getProperty("slot")).toEqual(SLOTS.control);
+
+        await control.click();
+        expect(blockToggleSpy).toHaveReceivedEventTimes(0);
+
+        await block.click();
+        await block.click();
+        expect(blockToggleSpy).toHaveReceivedEventTimes(2);
+      });
+
+      it("allows users to add a control in a non-collapsible block", async () => {
+        const page = await newE2EPage();
+        await page.setContent(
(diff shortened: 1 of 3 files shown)


src/components/calcite-block/calcite-block.e2e.ts
```diff
@@ -104,24 +104,47 @@ describe("calcite-block", () => {
...
+        const controlSlot = await page.find(`calcite-block >>> slot[name=${SLOTS.control}]`);
+        expect(await controlSlot.isVisible()).toBe(true);
+
+        const collapsibleIcon = await page.find(`calcite-block >>> .${CSS.toggleIcon}`);
+        expect(collapsibleIcon).toBeNull();
+
+        const block = await page.find("calcite-block");
+        const blockToggleSpy = await block.spyOnEvent("calciteBlockToggle");
+
+        expect(await control.isVisible()).toBe(true);
```
pr3tori4n: duplicate of line 114?

src/components/calcite-block/calcite-block.e2e.ts
```diff
@@ -104,24 +104,47 @@ describe("calcite-block", () => {
...
+        expect(await control.getProperty("slot")).toEqual(SLOTS.control);
```
pr3tori4n: this assertion valuable? See line 111.

src/components/calcite-block/calcite-block.e2e.ts
```diff
@@ -104,24 +104,47 @@ describe("calcite-block", () => {
...
-      expect(await control.getProperty("slot")).toEqual(SLOTS.control);
-
-      await control.click();
-      expect(elementToggleSpy).toHaveReceivedEventTimes(0);
-
-      await element.click();
-      await element.click();
-      expect(elementToggleSpy).toHaveReceivedEventTimes(2);
+    describe("adding a header control", () => {
+      it("allows users to add a control in a collapsible block", async () => {
```
pr3tori4n: Seems you're testing multiple things inside this one test block. Consider breaking this into separate isolated tests so it's clear what's working vs. what's breaking when this test passes/fails.

src/components/calcite-block/calcite-block.e2e.ts
```diff
@@ -104,24 +104,47 @@ describe("calcite-block", () => {
...
+
+        await control.click();
+        expect(blockToggleSpy).toHaveReceivedEventTimes(0);
+
+        await block.click();
+        await block.click();
+        expect(blockToggleSpy).toHaveReceivedEventTimes(2);
+      });
+
+      it("allows users to add a control in a non-collapsible block", async () => {
```
pr3tori4n: a bit confused with what this tests. Seems to be verifying that the control appears even if collapsible attr isn't present? I guess I'm missing the significance. Is there an implied relationship between those two properties? Could you add a comment that explains it?

src/components/calcite-block/calcite-block.tsx
```diff
@@ -152,6 +152,14 @@ export class CalciteBlock {
       </header>
     );
 
+    let hasControl = false;
+    const hasContent = Array.from(el.children).some((child) => {
+      const isControl = child.slot === SLOTS.control;
+      hasControl = hasControl || isControl;
```
pr3tori4n: pebble: array.prototype.some stops running on items in the array once it returns a true state.  It possible there'd be a bug where there is a control but it isn't found because `.some` stopped checking remaining items?

src/components/calcite-block/calcite-block.e2e.ts
jcfranco: copy🍝

src/components/calcite-block/calcite-block.e2e.ts
jcfranco: copy🍝

src/components/calcite-block/calcite-block.e2e.ts
jcfranco: There are multiple assertions, but they're all related to the same test.

src/components/calcite-block/calcite-block.tsx
jcfranco: Doh! Good catch. 🛠

src/components/calcite-block/calcite-block.e2e.ts
jcfranco: I forgot to drop that test. Fixing!

src/demos/calcite-block.html
```diff
@@ -141,6 +147,13 @@ <h2>Header + content (collapsible)</h2>
           </calcite-block-section>
         </calcite-block>
 
+        <h2>Header + collapsible + control = no collapsible icon</h2>
+
+        <calcite-block heading="no toggle icon when there is a control" collapsible>
+          <label slot="control">test</label>
```
asangma: Looks good.
Sand: maybe the demo can use a `calcite-action` with a `...` icon in the control slot? Mostly so as to be instructive.

Here's the SVG if you need it:
```
<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><path d="M14 9.05A1.05 1.05 0 1 1 15.05 8 1.05 1.05 0 0 1 14 9.05zm-6 0A1.05 1.05 0 1 1 9.05 8 1.05 1.05 0 0 1 8 9.05zm-6 0A1.05 1.05 0 1 1 3.05 8 1.05 1.05 0 0 1 2 9.05z"/></svg>
```

jcfranco: Shouldn't that be a part of the advanced demo?

Issue #461, opened 2019-10-24
Bug: PickList Filter match is too aggressive


pr3tori4n:
### Actual Behavior
When filtering, some items are matched due to some attributes containing one or two letters
### Expected Behavior
Only the following pickListItem attributes are used for filtering: `textLabel`, `textDescription`, `value`, and `metadata`

kat10140:
@AdelheidF want to check this in the current build?

AdelheidF:
Seems to work as I would expect now. Just did a bit of testing.

kat10140:
verified & closing

Issue #400, opened 2019-10-17
Bug: calcite-pick-list filter is hiding all items


AdelheidF:
### Summary

Using "@esri/calcite-app-components@4.2.1"

(1)
- [ ] When using the filter the list goes empty and doesn't come back when I remove the filter.

<img src=https://user-images.githubusercontent.com/44481292/67053556-a1e0dc00-f0f6-11e9-8616-fe04dda4508e.png width=200/>

<img src=https://user-images.githubusercontent.com/44481292/67053615-c1780480-f0f6-11e9-91e1-4e43d321f19d.png width=200/>

<img src=https://user-images.githubusercontent.com/44481292/67053642-d3f23e00-f0f6-11e9-8584-a2dc409bdead.png width=200/>

(2) 
- [x] When using slot menu-actions with a dropdown there's a CSS issue.
Separate issue created: https://github.com/Esri/calcite-app-components/issues/417

<img src=https://user-images.githubusercontent.com/44481292/67053591-b58c4280-f0f6-11e9-8938-42261abef845.png width=100/>

```
        <calcite-pick-list
          filter-enabled="true"
          multiple="true"
        >
          <calcite-dropdown slot="menu-actions" alignment="right">
            <calcite-action
              slot="dropdown-trigger"
              onclick=...
              label="Sort"
            >
              <svg>...</svg>
            </calcite-action>
            <calcite-dropdown-group grouptitle="Sort by">
              <calcite-dropdown-item active>Display name</calcite-dropdown-item>
              <calcite-dropdown-item>Type</calcite-dropdown-item>
            </calcite-dropdown-group>
          </calcite-dropdown>
          {items}
        </calcite-pick-list>

```

(3) 
- [ ] When in `multiple=false` mode I can deselect the single selected radio button by clicking on it again. I don't think it should let me do that. When in radio button mode there should be always one selected.

When using the "calciteListChange" event on a filtered list I assume I still get all checked items, correct? The filter doesn't affect the list of items. Can't test this currently.

### Reproduction Steps

local version of https://devext.arcgis.com/apps/mapviewer/index.html

jbartley:
What is the latest on this issue? @pr3tori4n

pr3tori4n:
@jbartley I'm trying to find the root cause. I'm having trouble reproducing this in isolation. I just found out that Adelheide has a test-app setup in the app-components repo so I'm going to take a look at that. This is a priority for me.

pr3tori4n:
[...]

PR #484, opened 2019-10-31
PickList and ValueList: share more code between lists 461

pr3tori4n:
**Related Issue:** #461, #400  

## Summary

TL;DR; I've abstracted some shared code for PickList and ValueList.
In working on a couple bugs for the lists, I've realized that they are already starting to diverge in undesirable ways. From the get-go the test suite was more extensive on pick-list than value-list. And there was quite a bit of duplicated code. I left them isolated since it seemed like these might diverge more based on information from @asangma and @kat10140 but so far they are still very similar.

## Changes

* Several methods have been abstracted to a new filed called `share-list-logic.ts` and both components are importing that as a dependency and use those functions for several internal methods.
* Several tests have been abstracted into `shared-list-tests.ts`, mainly the selection/deselection tests as well as the filtering logic.

## Future
I think it might be possible to also share render logic, but that would need to be abstracted into a functionalComponent at minimum. It might also be possible using higher-order component patterns to share more of the class configuration (props, states and other decorators). Going to pause on that rabbit hole for now so I can get these other bug fixes shipped first.

**_NOTE:_** This is being merged into one of my bug fix branches because I wrote the new filter tests for that bug, then decided to do this rather than copy all of those tests over.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,7 @@ Thumbs.db
 UserInterfaceState.xcuserstate
 .env
 src/**/components.d.ts
+
+# backup files
+*.bak
+*.orig
--- a/src/components/calcite-pick-list/calcite-pick-list.tsx
+++ b/src/components/calcite-pick-list/calcite-pick-list.tsx
@@ -198,16 +198,16 @@ export class CalcitePickList {
   }
 
   render() {
-    const { disabled, loading } = this;
+    const { dataForFilter, handleFilter, filterEnabled, loading, disabled } = this;
     return (
       <Host aria-disabled={disabled} aria-busy={loading}>
         <header>
-          {this.filterEnabled ? (
+          {filterEnabled ? (
             <calcite-filter
-              data={this.dataForFilter}
+              data={dataForFilter}
               textPlaceholder={TEXT.filterPlaceholder}
               aria-label={TEXT.filterPlaceholder}
-              onCalciteFilterChange={this.handleFilter}
+              onCalciteFilterChange={handleFilter}
             />
           ) : null}
           <slot name="menu-actions" />
--- a/src/components/calcite-pick-list/calcite-pick-list.tsx
+++ b/src/components/calcite-pick-list/calcite-pick-list.tsx
@@ -87,10 +87,7 @@ export class CalcitePickList {
 
   lastSelectedItem: HTMLCalcitePickListItemElement = null;
 
-  observer = new MutationObserver(() => {
-    this.setUpItems();
-    this.setUpFilter();
-  });
+  observer = new MutationObserver(sharedListMethods.mutationObserverCallback.bind(this));
 
   // --------------------------------------------------------------------------
   //
--- a/src/components/calcite-pick-list/calcite-pick-list.tsx
+++ b/src/components/calcite-pick-list/calcite-pick-list.tsx
@@ -11,7 +11,8 @@ import {
   h
 } from "@stencil/core";
 import { ICON_TYPES, TEXT } from "./resources";
-import { VNode } from "@esri/calcite-components/dist/types/stencil.core";
+import sharedListMethods from "./shared-list-logic";
+import { VNode } from "@stencil/core/dist/declarations";
 import CalciteScrim from "../utils/CalciteScrim";
 
 /**
@@ -53,7 +54,7 @@ export class CalcitePickList {
    */
   @Prop({ reflect: true }) mode: "selection" | "configuration" = "selection";
   /**
-   * Multiple works like conventional checkboxes and radio buttons.
+   * Multiple Works similar to standard radio buttons and checkboxes.
    * When true, a user can select multiple items at a time.
    * When false, only a single item can be selected at a time
    * and selecting a new item will deselect any other selected items.
@@ -104,14 +105,16 @@ export class CalcitePickList {
   //
   // --------------------------------------------------------------------------
 
+  connectedCallback() {
+    sharedListMethods.initialize.call(this);
+  }
+
   componentDidLoad() {
-    this.setUpItems();
-    this.setUpFilter();
-    this.observer.observe(this.el, { childList: true, subtree: true });
+    sharedListMethods.initializeObserver.call(this);
   }
 
   componentDidUnload() {
-    this.observer.disconnect();
+    sharedListMethods.cleanUpObserver.call(this);
   }
 
   // --------------------------------------------------------------------------
@@ -133,22 +136,7 @@ export class CalcitePickList {
   @Event() calcitePickListSelectionChange: EventEmitter;
 
   @Listen("calciteListItemChange") calciteListItemChangeHandler(event) {
-    const { selectedValues } = this;
-    const { item, value, selected, shiftPressed } = event.detail;
-    if (selected) {
-      if (!this.multiple) {
-        this.deselectSiblingItems(item);
-      }
-      if (this.multiple && shiftPressed) {
-        this.selectSiblings(item);
-      }
-      selectedValues.set(value, item);
-    } else {
-      selectedValues.delete(value);
-    }
-    this.lastSelectedItem = item;
-    this.calciteListChange.emit(selectedValues);
-    this.calcitePickListSelectionChange.emit(selectedValues);
+    sharedListMethods.calciteListItemChangeHandler.call(this, event);
   }
 
   @Listen("calciteListItemPropsUpdated") calciteListItemPropsUpdatedHandler() {
@@ -162,17 +150,7 @@ export class CalcitePickList {
   // --------------------------------------------------------------------------
 
   setUpItems(): void {
-    this.items = Array.from(this.el.querySelectorAll("calcite-pick-list-item"));
-    this.items.forEach((item) => {
-      const iconType = this.getIconType();
-      const compactString = this.compact ? "true" : "false";
-      item.setAttribute("icon", iconType);
-      item.setAttribute("compact", compactString);
-
-      if (item.hasAttribute("selected")) {
-        this.selectedValues.set(item.getAttribute("value"), item);
-      }
-    });
+    sharedListMethods.setUpItems.call(this, "calcite-pick-list-item");
   }
 
   setUpFilter(): void {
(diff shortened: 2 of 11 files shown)


src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -53,7 +54,7 @@ export class CalcitePickList {
    */
   @Prop({ reflect: true }) mode: "selection" | "configuration" = "selection";
   /**
-   * Multiple works like conventional checkboxes and radio buttons.
+   * Multiple Works similar to standard radio buttons and checkboxes.
```
pr3tori4n: The wording here was different from the valueList. Didn't see a good reason for these to use different language.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -104,14 +105,16 @@ export class CalcitePickList {
   //
   // --------------------------------------------------------------------------
 
+  connectedCallback() {
+    sharedListMethods.initialize.call(this);
```
pr3tori4n: Moving these back to connectedCallback was intentional. Once the logic is modified to use props instead of attributes, it shoouldn't need to be delayed until after first-render. #400

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (this.multiple && shiftPressed) {
+        this.selectSiblings(item);
+      }
+      selectedValues.set(value, item);
+    } else {
+      selectedValues.delete(value);
+    }
+    this.lastSelectedItem = item;
+    this.calciteListChange.emit(selectedValues);
+    // this.calcitePickListSelectionChange.emit(selectedValues); picklist only
```
pr3tori4n: remove

src/components/calcite-tip-manager/calcite-tip-manager.spec.ts
```diff
@@ -1,4 +1,4 @@
-import { CalciteTipManager } from "./calcite-tip-manager";
+// import { CalciteTipManager } from "./calcite-tip-manager";
```
pr3tori4n: I turned on squigglies and started fixing some things :)

src/components/calcite-value-list/calcite-value-list.e2e.ts
```diff
@@ -1,49 +1,14 @@
 import { newE2EPage } from "@stencil/core/testing";
-import { CSS, ICON_TYPES } from "./resources";
+import { ICON_TYPES } from "./resources";
```
pr3tori4n: I turned on squigglies and started fixing some things :)

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -11,7 +11,8 @@ import {
   h
 } from "@stencil/core";
 import { ICON_TYPES, TEXT } from "./resources";
-import { VNode } from "@esri/calcite-components/dist/types/stencil.core";
+import sharedListMethods from "./shared-list-logic";
```
driskull: nitpick: might be nicer to import the methods directly as needed instead of having to call `sharedListMethods.method` every time they are used.

`import {method} from "..."`

src/components/calcite-pick-list/shared-list-tests.ts
```diff
@@ -0,0 +1,229 @@
+import { E2EElement, E2EPage, newE2EPage } from "@stencil/core/testing";
+
+export const tests = {
+  selectionAndDeselection(listType) {
```
jcfranco: We should try and provide the type for anything we can.

src/components/calcite-pick-list/shared-list-tests.ts
```diff
@@ -0,0 +1,229 @@
...
+        filterInput.dispatchEvent(new Event("input"));
+      });
+      await item1.waitForNotVisible();
+
+      item1Visible = await item1.isVisible();
+      item2Visible = await item2.isVisible();
+      expect(item1Visible).toBe(false);
+      expect(item2Visible).toBe(true);
+    });
+    it("should match text in the text-description prop", async () => {
```
jcfranco: Nitpick: Adding newlines between tests/suites would improve readability.

src/components/calcite-tip-manager/calcite-tip-manager.spec.ts
jcfranco: This should be fixed or removed. #nodeadcomments

pr3tori4n:
@driskull I could move to a pure function approach instead of binding the scope, but that would require a good deal more effort refactoring. I was worried I'd have to do it that way, but was able to get it working with this approach, which allowed me to keep the code mostly the same. Is there a solid benefit you feel the functional approach would provide that would be worth that extra effort?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -132,23 +129,11 @@ export class CalciteValueList {
...
-      if (this.multiple && shiftPressed) {
-        this.selectSiblings(item);
-      }
-      selectedValues.set(value, item);
-    } else {
-      selectedValues.delete(value);
-    }
-    this.lastSelectedItem = item;
-    this.calciteListChange.emit(selectedValues);
+    sharedListMethods.calciteListItemChangeHandler.call(this, event);
```
jcfranco: Suggestion: Adding a `thisArg` to the shared methods may help clean up consuming code and also makes it clear that the method needs a different context.

.gitignore
```diff
@@ -24,3 +24,7 @@ Thumbs.db
 UserInterfaceState.xcuserstate
 .env
 src/**/components.d.ts
+
+# backup files
+*.bak
```
jcfranco: Where are these files coming from? If not from an IDE, I'd suggest using git stash or local branches.

driskull:
> Is there a solid benefit you feel the functional approach would provide that would be worth that extra effort?

The issue I see is that `this` is typed as `any` so it could be error prone to have it like this. So in the shared code file, anytime `this` is used the typing is lost.

image

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.setUpFilter();
+  }),
+  // LifeCycle functions
+  initialize() {
+    // connectedCallback
+    this.setUpItems();
+    this.setUpFilter();
+  },
+  initializeObserver() {
+    // componentDidLoad
```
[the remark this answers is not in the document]
pr3tori4n: The comments were mainly for my own benefit during the refactor. I can remove.
My hesitation to match them is that we may need to move them to a different lifecycle in the future, then we'd either have inconsistencies or have to rename these. I figured better to describe the function with its name. But if you feel strongly, I can modify.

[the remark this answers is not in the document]
jcfranco: I'm fine with dropping comments if it's not specifically tied to a lifecycle hook. The method names and logic are expressive enough.

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,110 @@
...
+      obj.label = item.textLabel || item.textHeading;
+      obj.description = item.textDescription;
+      obj.metadata = item.metadata;
+      obj.value = item.value;
+      result.push(obj);
+    });
+    return result;
+  }
+};
+export default sharedListMethods;
```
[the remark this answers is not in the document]
pr3tori4n: Read. I'm on board with treating named exports as the best practice and preferred going forward. I hesitate to say "never" use default exports. There may be a time and place for them.
👍

.gitignore
pr3tori4n: > I'd suggest using git stash or local branches.
👍 
 I don't feel strongly about keeping it it gitignore -  but its a small benefit to keep these file types ignored. Are you also asking to remove this?

pr3tori4n:
@driskull gotcha. Let me see if I can fix the typings in a simple way. If not I'll look at going with a dependency injection model.

.gitignore
jcfranco: Not right now. I think in the future this file should be revisited.

pr3tori4n:
@driskull I found a solution. Typescript doesn't infer `this` types from calls to `.bind, .call or .apply` automatically, but you can tell a function what it should expect as acceptable values of `this`
https://stackoverflow.com/questions/54710927/typescript-types-and-bind

src/components/calcite-pick-list/shared-list-logic.ts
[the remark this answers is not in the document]
driskull: Nitpick: For this method and the `cleanUpObserver` method, can you just pass the observer? It would probably be nicer to make these pure functions when they can be pure functions.

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,121 @@
...
+  initializeObserver(this: CalcitePickList | CalciteValueList) {
+    // componentDidLoad
+    this.observer.observe(this.el, { childList: true, subtree: true });
+  },
+  cleanUpObserver(this: CalcitePickList | CalciteValueList) {
+    // componentDidUnload
+    this.observer.disconnect();
+  },
+  // Listeners
+  calciteListItemChangeHandler(this: CalcitePickList | CalciteValueList, event): void {
```
driskull: fixme: Does event need to be typed?

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,121 @@
...
+      selectedValues.set(value, item);
+    } else {
+      selectedValues.delete(value);
+    }
+    this.lastSelectedItem = item;
+    this.calciteListChange.emit(selectedValues);
+    // this.calcitePickListSelectionChange.emit(selectedValues); picklist only
+  },
+  // Private Methods
+  setUpItems(this: CalcitePickList | CalciteValueList, tagname): void {
```
driskull: fixme: Can you type `tagName`?

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,110 @@
+export const sharedListMethods = {
+  getObserver: new MutationObserver(() => {
```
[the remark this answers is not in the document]
pr3tori4n: refactored a bit so it can be used.

src/components/calcite-tip-manager/calcite-tip-manager.spec.ts
pr3tori4n: The test is valuable but it can't be ran yet because of a Stencil bug. I can't skip the import so have to comment it or remove it. I'd prefer to keep it commented as the test will be easier to turn back on once the spec tests are working. But if you feel strongly, I'll remove.

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: Would you be ok if I just added a destructuring call after the import for now?

driskull: 👍 Its just a nitpick.

pr3tori4n: added destructuring statement. Avoiding splitting into multiple exports for now since it doesn't help with tree-shaking (they are all used).

src/components/calcite-pick-list/shared-list-logic.ts
pr3tori4n: done.

src/components/calcite-pick-list/shared-list-logic.ts
pr3tori4n: yep.

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,135 @@
+import { CalcitePickList } from "./calcite-pick-list";
+import { CalciteValueList } from "../calcite-value-list/calcite-value-list";
+
+function updateSelectedValuesMap(
+  selectedValuesMap: Map<string, HTMLCalcitePickListItemElement> | Map<string, HTMLCalciteValueListItemElement>,
```
driskull: Nitpick: Maybe you could add a couple types in this file for `Map<string, HTMLCalcitePickListItemElement>` and `Map<string, HTMLCalciteValueListItemElement>` as it's repeated a lot and lengthy.

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -0,0 +1,135 @@
...
+  item: HTMLCalcitePickListItemElement | HTMLCalciteValueListItemElement
+): void {
+  const selectedValues = (selectedValuesMap as any) as Map<
```
driskull: Question: Anyway to avoid the `any` and `as` used here?

jcfranco: https://github.com/Esri/calcite-app-components/issues/492

src/components/calcite-pick-list/shared-list-logic.ts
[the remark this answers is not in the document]
jcfranco: > The default export should be renamed to match the filename. 

This has not been addressed.

> There may be a time and place for them.

You can achieve the same without a default export. This has come up before: https://github.com/Esri/calcite-app-components/pull/33#discussion_r293452795

This can be done in a follow-up PR.

src/components/calcite-tip-manager/calcite-tip-manager.spec.ts
jcfranco: OK, let's keep this for now. Maybe add a comment with what you described? I know it'll stop me from commenting on the same thing in the future. 😄

pr3tori4n:
> This has not been addressed.

I removed it but forgot to push. It should show in the PR now.

> OK, let's keep this for now. Maybe add a comment with what you described? I know it'll stop me from commenting on the same thing in the future. 😄

There is a comment on line 5, I'll move it up to line 1.

PR #488, opened 2019-11-01
PickList and ValueList: share render methods between lists 461

pr3tori4n:
**Related Issue:** #461 

## Summary

Abstract render functions too.

### DIFF
--- a/src/components/calcite-pick-list/calcite-pick-list.tsx
+++ b/src/components/calcite-pick-list/calcite-pick-list.tsx
@@ -198,16 +198,16 @@ export class CalcitePickList {
   }
 
   render() {
-    const { disabled, loading } = this;
+    const { dataForFilter, handleFilter, filterEnabled, loading, disabled } = this;
     return (
       <Host aria-disabled={disabled} aria-busy={loading}>
         <header>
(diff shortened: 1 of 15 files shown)


src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -12,8 +11,7 @@ import {
 } from "@stencil/core";
 import { ICON_TYPES, TEXT } from "./resources";
 import sharedListMethods from "./shared-list-logic";
-import { VNode } from "@stencil/core/dist/declarations";
-import CalciteScrim from "../utils/CalciteScrim";
+import List from "./shared-list-render";
```
jcfranco: Nitpick: Is there a more discoverable location for the common list modules?

pr3tori4n: We could put them in utils maybe? Though I think it'd be about time to start adding sub-folders to utils, which might introduce some file reference refactors in other files.
Maybe `src/components/utils/functionalComponents`

Very open to other suggestions

pr3tori4n:
Thanks Franco 😄

Issue #487, opened 2019-11-01
Enhancement: Update spacing for Panel, Blocks, and Action to align with calcite components


mitc7862:
### Description
- Simplify the spacing of nested container elements

### Acceptance Criteria 
- text is consistently styled and aligned
- hierarchy of nested components is clear

#### Desired New Behavior
- No behavior change

#### Desired Change to Look and Feel
- Will post design

#### Out of Scope
- We aren't considering any updates that would require changes to the structure of the components themselves.

### Relevant Info 
- This enhancement was requested by the calcite component team in order to achieve greater design consistency across the two resources.  

### Helpful Details

**Which Component:** Block, Block section, Panel, ActionBar, ActionPad, Action
**Example Use Case:** n/a

asangma:
The update that removed padding causes more issues than it solves (#454).
Bringing it back as this issue suggests (and possibly enhancing it) will will negate the need to add a padding prop (#457).

asangma:
Also re #482

asangma:
re #186

asangma:
@mitch We can review on my local to verfee.

kat10140:
thumbified

PR #489, opened 2019-11-04
Spacing updates and calcite-components alignment

asangma:
**Related Issue:** #487

## Summary
Broad spacing and UI refinements.
Improves consistent hit areas and aligns better with calcite-components design.

cc @macandcheese @julio8a

### DIFF
--- a/src/components/calcite-pick-list/calcite-pick-list.tsx
+++ b/src/components/calcite-pick-list/calcite-pick-list.tsx
@@ -251,19 +251,22 @@ export class CalcitePickList {
 
   render() {
     const { disabled, loading } = this;
+    const hasMenu = this.el.querySelector("[slot=menu-actions]");
     return (
       <Host aria-disabled={disabled} aria-busy={loading}>
-        <header>
-          {this.filterEnabled ? (
-            <calcite-filter
-              data={this.dataForFilter}
-              textPlaceholder={TEXT.filterPlaceholder}
-              aria-label={TEXT.filterPlaceholder}
-              onCalciteFilterChange={this.handleFilter}
-            />
-          ) : null}
-          <slot name="menu-actions" />
-        </header>
+        {hasMenu ? (
+          <header>
+            {this.filterEnabled ? (
+              <calcite-filter
+                data={this.dataForFilter}
+                textPlaceholder={TEXT.filterPlaceholder}
+                aria-label={TEXT.filterPlaceholder}
+                onCalciteFilterChange={this.handleFilter}
+              />
+            ) : null}
+            <slot name="menu-actions" />
+          </header>
+        ) : null}
         <slot />
         {this.renderScrim()}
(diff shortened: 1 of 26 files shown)


src/components/calcite-shell-panel/calcite-shell-panel.scss
```diff
@@ -27,3 +24,10 @@ $shell-panel-max-width: 360px !default;
 .content[hidden] {
   display: none;
 }
+
+:host([layout="leading"]) slot[name="action-bar"]::slotted(calcite-action-bar) {
```
driskull: Can you verify this is working in edge. just curious.

asangma: @driskull I'll have to hand that testing over to someone who has an Edge device. 😬

asangma: TBH, it's a pretty minor styling update and could either be removed or allowed to not work on Edge.

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -251,19 +251,22 @@ export class CalcitePickList {
...
-            <calcite-filter
-              data={this.dataForFilter}
-              textPlaceholder={TEXT.filterPlaceholder}
-              aria-label={TEXT.filterPlaceholder}
-              onCalciteFilterChange={this.handleFilter}
-            />
-          ) : null}
-          <slot name="menu-actions" />
-        </header>
+        {hasMenu ? (
```
pr3tori4n: This will not render the filter in the event filter-enabled is true but there's no menu-actions slot. Not sure the intention but I'm pretty sure we don't want that side-effect.

src/components/calcite-shell-panel/calcite-shell-panel.scss
pr3tori4n: if this selector isn't understood by Edge, I'd expect the border to be absent. If that's ok then 👍 .

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: Looks like in the other file you're using `filterEnabled || hasMenu ?` which is probably what you want here.

asangma: Ah! Good catch.

asangma: Thanks @pr3tori4n. Fixed.

asangma:
@pr3tori4n @driskull Please feel free to merge once you confirm the requested changes as I may be in meetings the rest of the day and would like to get this into a release today.

asangma:
THANKS ROCKERS! \m/(>_<)\m/

PR #491, opened 2019-11-04
fix: calcite-action loading & calcite-flow-item footer in Edge browser

driskull:
**Related Issue:** None

## Summary

- Fixes loading spinner to show without the SVG in the CalciteAction in Edge.
  - The loading spinner was displaying above the SVG icon. This hides the svg when loading.
- Fixes footer display issue in CalciteFlowItem in Edge.
  - A slot with another slot attribute won't work correctly in the Edge Polyfill


driskull:
murgling. @jcfranco let me know if there's something I need to fix here.

PR #499, opened 2019-11-05
PickList/ValueList use props not attrs 400

pr3tori4n:
**Related Issue:** #400 

## Summary

Maquette sets properties on html elements it renders. It doesn't add attributes. Therefore we need to modify the logic in the lists so they are reading from props instead of attributes in order to detect starting state for lists or their child items.


pr3tori4n:
@driskull you mean a test that configures an item using props instead of attributes in `page.setContent`?

Two options. Do we want just one special test for this? Would be redundant to do each test twice, once with attrs and one with props - and would add a lot of testing time.

src/demos/calcite-pick-list-advanced.html
```diff
@@ -0,0 +1,451 @@
...
+    <script>
+      function toggleProperty(property) {
+        const blocks = document.getElementsByTagName("calcite-pick-list");
+        for (i = 0; i < blocks.length; i++) {
+          blocks[i].toggleAttribute(property);
+        }
+      }
+    </script>
+    <script>
+      function renderMaquette() {
```
jcfranco: I don't think we should include this in our demos. At least from reading this page, there's no context on why this exists, doesn't look exhaustive, and also opens a can of worms whether we need to update other demos for consistency. This is testing component creation via properties through maquette, which we could cover in E2E testing. Maybe we could use maquette as a util to quickly set up these tests since it does make it simple to set up, but I see this as something separate and test-related.

driskull:
Yeah, nevermind. It seems redundant.

pr3tori4n:
An integration test using Maquette would cover a majority of the functionality. Something to consider during the discussion.

src/demos/calcite-pick-list-advanced.html
driskull: Agreed. We shouldn't be testing integrations in this repo. I missed this

src/demos/calcite-pick-list-advanced.html
```diff
@@ -0,0 +1,429 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0" />
+    <title>Demos - PickList</title>
+    <script src="head.js"></script>
+    <script src="https://unpkg.com/maquette@3.3.0/dist/maquette.umd.js"></script>
```
jcfranco: Last maquette leftover.

Issue #367, opened 2019-10-09
Bug: calcite-panel - footer buttons width not correct


AdelheidF:
### Summary

Once I upgraded to 4.1.0 the footer buttons don't work as expected. 
Done and Cancel have a width of `half` and together should fill the whole panel's width.

![4](https://user-images.githubusercontent.com/44481292/66525100-3a4ce000-eaa9-11e9-8ac2-1af67ec993db.png)

### Reproduction Steps

https://energy.mapsdevext.arcgis.com/apps/mapviewer/index.html

driskull:
Very odd but on the mapviewer prototype I see this:

![image](https://user-images.githubusercontent.com/1231455/66525720-35892b80-eaab-11e9-8bb4-54d7691e549f.png)

But on the local build I see this:

![image](https://user-images.githubusercontent.com/1231455/66525741-45a10b00-eaab-11e9-8497-534ad56aa06f.png)


@jcfranco we may need to investigate what's going on with those missing CSS styles.

jcfranco:
@driskull Setting the footer slot to 100% fixes the issue, but I'm not sure if this is a regression or not since I see the same issue from the description w/ `calcite-components@beta.10` + `calcite-app-components@v4.0.0` ([pen](https://codepen.io/jcfranco-the-scripter/pen/JjjXrOE)) and `calcite-app-components@v4.1.0` ([pen](https://codepen.io/jcfranco-the-scripter/pen/jOOqGLR)).

@AdelheidF Which dependencies + versions were you using before moving to 4.1.0?

---

> we may need to investigate what's going on with those missing CSS styles.

Based on my testing (4.0.0 - 4.1.1), I think this has always been happening. I looked into the dist folder contents and the string used for to generate the internal style of the panel component looks correct and includes the footer declaration block in its entirety, it just isn't applied at runtime. The issue seems to manifest when using the output from `npm run build`. The `dist` folder produced by `npm start` works as expected. This may be a Stencil bug.

AdelheidF:
> Which dependencies + versions were you using before moving to 4.1.0?
@esri/calcite-components@1.0.0-beta.10

"@esri/calcite-app-components": "3.1.0",	    
"@esri/calcite-ui-icons": "^2.4.0",

jcfranco:
@AdelheidF I see the same w/ 3.1.0: https://codepen.io/jcfranco-the-scripter/pen/eYYzNdw?editors=1000 Is there something I missed?

[...]

PR #500, opened 2019-11-05
fix: flow-item and panel footer buttons

asangma:
**Related Issue:** #367 

## Summary
Fixes layout of calcite-button (or other generic button) in panel footer. flow-item inherits the fix with a bit of additional styling.

Also updated full-window demo to use calcite-button.


asangma:
Thanks @driskull!

Issue #501, opened 2019-11-05
Enhancement: Refactor all `getAttribute` and `hasAttribute` methods


pr3tori4n:
### Description
Maquette and other hyperscript libraries set properties on html elements and not attributes, even if the JSX syntax is similar to HTML attributes.

This means if we're using  `getAttribute` and `hasAttribute` in first-render lifecycle functions, it can cause bugs where the property and attribute are our of sync. So we need to rely on the properties over the attributes for discerning component logic.

### Acceptance Criteria 
This is mainly a refactor. The use of the above methods should be replaced by dot notation lookup of an elements property, and testing against undefined, respectively.
#### Desired New Behavior
The behavior should remain the same.
#### Desired Change to Look and Feel
The styles should remain the same.

### Helpful Details
**Which Component:** Any that use one of those methods.

kat10140:
@driskull should all this have been taken care of in the merge?

driskull:
Yes

pr3tori4n:
@kat10140 Please verify.

kat10140:
verified!

PR #502, opened 2019-11-05
fix: Set the property of the component instead of the attribute

driskull:
Set the property of the component instead of the attribute as a best practice. Also adds a selected property where it wasn't defined but being used.

**Related Issue:** #501

## Summary

Set the property of the component instead of the attribute as a best practice. Also adds a selected property where it wasn't defined but being used.


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -168,7 +168,7 @@ export class CalciteTipManager {
   showSelectedTip() {
     this.tips.forEach((tip, index) => {
       const isSelected = this.selectedIndex === index;
-      isSelected ? tip.setAttribute("selected", "") : tip.removeAttribute("selected");
+      tip.selected = isSelected;
       isSelected ? tip.removeAttribute("hidden") : tip.setAttribute("hidden", "");
```
pr3tori4n: you can also use the hidden property here I think.

Issue #402, opened 2019-10-18
Enhancement: Add test for Block loading and disabled


asangma:
### Description
Create test for Block loading/disabled.
re: https://github.com/Esri/calcite-app-components/pull/379#issuecomment-543803247

### Relevant Info 
From #379

**Which Component:**  Block 

Thanks @jcfranco!

jcfranco:
Installed.

PR #508, opened 2019-11-06
test(block): add tests for loading and disabled properties

jcfranco:
**Related Issue:** #402

## Summary

Adds missing Block tests.

**Note**: Left similarities between tests as is. Extracting into a helper didn't make things clearer. Open to suggestions though. 😄

### DIFF
--- a/src/components/calcite-block/calcite-block.e2e.ts
+++ b/src/components/calcite-block/calcite-block.e2e.ts
@@ -1,6 +1,8 @@
-import { newE2EPage } from "@stencil/core/testing";
+import { E2EElement, newE2EPage } from "@stencil/core/testing";
 import { CSS, SLOTS, TEXT } from "./resources";
 import { accessible, defaults, hidden, renders } from "../../tests/commonTests";
+import { setUpPage } from "../../tests/utils";
+import { CalciteBlock } from "./calcite-block";
 
 describe("calcite-block", () => {
   it("renders", async () => renders("calcite-block"));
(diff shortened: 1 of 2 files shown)


src/demos/calcite-block.html
```diff
@@ -154,6 +154,18 @@ <h2>Header + collapsible + control = no collapsible icon</h2>
...
+        <calcite-block disabled heading="this block is disabled" collapsible open>
+          <div>Some content</div>
+        </calcite-block>
+
+        <h2>Loading</h2>
+
+        <calcite-block loading heading="this block is loading" collapsible open>
+          <div>Some content</div>
+        </calcite-block>
+
```
pr3tori4n: I don't think you need these since @asangma added the toggle buttons at the top, which will toggle all blocks to be disabled/loading.

src/components/calcite-block/calcite-block.e2e.ts
```diff
@@ -28,6 +29,56 @@ describe("calcite-block", () => {
...
+
+    const block = await page.find("calcite-block");
+    block.setProperty("loading", true);
+    await page.waitForChanges();
+
+    await content.click();
+    expect(clickSpy).toHaveReceivedEventTimes(1);
+
+    expect(await page.find("calcite-block >>> calcite-loader")).toBeTruthy();
+  });
```
pr3tori4n: This is great imo. When it comes to tests sometimes its better to keep the code simple and not worry too much about it being DRY. Adding complexity can sometimes introduce bugs in the test code which is annoying.

jcfranco: Agreed.

src/demos/calcite-block.html
jcfranco: You can tell I haven't used those toggles. 😄Will remove!

asangma: Indeed. @jcfranco Do you think we need these since we have the buttons at the top of the demo?

Issue #503, opened 2019-11-05
Enhancement: Sticky Filter


kat10140:
### Description
As filter is now, if one scrolls with the pick or value list, the filter scrolls as well. We want the filter to stick at the top of the list at all times.

### Acceptance Criteria 
Sticky filter

#### Desired New Behavior
Filter to stick at top of panel when scrolling 

#### Desired Change to Look and Feel
Sticky

pr3tori4n:
In code review.

pr3tori4n:
@kat10140 please verify.

kat10140:
gonna verify this tomorrow when we release again in the new mapviewer

kat10140:
verified

PR #509, opened 2019-11-06
Hrobbins/picklist sticky filter 503

pr3tori4n:
**Related Issue:** #503 

## Summary

Add an option to Pick and Value Lists to enable a sticky filter. This causes the filter to stick to the top of the container if the user scrolls beyond the point where the filter is visible.

API Design additions:
```
class PickList { //and ValueList
  filterSticky: false
}
```
```
<calcite-pick-list filter-enabled filter-sticky>
</calcite-pick-list>
```


src/components/calcite-pick-list/shared-list-render.tsx
```diff
@@ -1,17 +1,18 @@
...
 };
 
 export const List = ({ props, text, ...rest }): VNode => {
-  const { disabled, loading, filterEnabled, dataForFilter, handleFilter } = props;
+  const { disabled, loading, filterEnabled, filterSticky, dataForFilter, handleFilter } = props;
   const { filterPlaceholder } = text;
   return (
     <Host aria-disabled={disabled} aria-busy={loading} {...rest}>
-      <header>
+      <header class={{ [CSS.sticky]: filterSticky }}>
```
pr3tori4n: This class has to go on the header, because of certain limitations of `position: sticky`. One limitation is that the immediate parent is the container that its bound to. The other limitation is it will only move over sibling elements. The filter is the only child of header, which means there's nothing for it to hover over. The `header` tag is the sibling to the items `<slot>` which allows this to function properly.

src/components/calcite-pick-list/calcite-pick-list.scss
```diff
@@ -13,6 +13,11 @@ header {
   justify-content: flex-end;
   align-items: center;
   margin-bottom: var(--calcite-app-cap-spacing-half);
+  &.sticky {
```
pr3tori4n: this doesn't have to be a compound selector - thanks to style encapsulation. I thought it a slight advantage to indicate in the CSS which element this modifier belongs to. LMK if you think it should be otherwise.

src/components/calcite-pick-list/shared-list-render.tsx
driskull: nitpick: Maybe the property should be `headerSticky` or `stickyHeader` then?

driskull: Could there be anything else in the header at some point? The menu actions are stuck as well so maybe header being sticky makes more sense.

pr3tori4n:
@driskull regarding change to `headerSticky` - I think `filterSticky` is more intuitive to the end-users as to what functionality they'll expect. The fact it's actually the header element shouldn't be too important to end users, it's more of an implementation detail imo.  

> Could there be anything else in the header at some point? 

Yea potentially. We'll have to revisit this logic if that ever happens. I don't think it's something we need to figure out now though.

driskull:
But the menu actions are stuck too, and they aren't tied to the filter are they?

pr3tori4n:
@driskull oh that's a good point. They are on the same visual "row" so maybe that's desired? I think we'd have to adjust the DOM at the very least to prevent the menu-actions from being sticky. @asangma you think its ok if the menu-actions along with the filter stick or do we need to split them?

pr3tori4n:
>  they aren't tied to the filter are they

No. so maybe stickyHeader is more accurate and better. I'm just worried anyone who isn't familiar with the internals isn't going to know what the "header" is. There another term that's maybe more descriptive?

src/components/calcite-pick-list/calcite-pick-list.tsx
```diff
@@ -62,6 +62,11 @@ export class CalcitePickList {
    */
   @Prop({ reflect: true }) filterEnabled = false;
 
+  /**
+   * When true, the filter will stick to the top of the pick list when scrolling would otherwise hide filter.
+   */
+  @Prop({ reflect: true }) filterSticky = false;
```
driskull: @pr3tori4n do we even need the option to disable this? It seems like we'd want it always to behave the same way.

src/components/calcite-pick-list/shared-list-render.tsx
driskull: If we need the prop, I think making it act on the header makes more sense.

asangma: I don't think this should be an option.

asangma: Basically, the filter should just be sticky.

driskull: Yup. makes sense. The issue says `always` and doesn't mention an option.

src/components/calcite-pick-list/calcite-pick-list.tsx
asangma: +1

src/components/calcite-pick-list/shared-list-render.tsx
kat10140: I'm happy with just the filter being sticky too

asangma:
@pr3tori4n This should also be tested in 

* a panel
* in a popover in panel

pr3tori4n:
> @pr3tori4n This should also be tested in 
a panel
in a popover in panel

@kat10140  could you assist me with that?

pr3tori4n:
@asangma @driskull ok I'll remove the option.
@asangma what about the menu-actions slot?

> But the menu actions are stuck too, and they aren't tied to the filter are they?
They are on the same visual "row" so maybe that's desired? I think we'd have to adjust the DOM at the very least to prevent the menu-actions from being sticky. @asangma you think its ok if the menu-actions along with the filter stick or do we need to split them?

asangma:
The menu-actions should stick with everything else in the header.

asangma:
Should still ensure this works in a Panel in a Popover.

src/components/calcite-pick-list/calcite-pick-list.tsx
pr3tori4n: removed the prop.

src/demos/calcite-pick-list.html
```diff
@@ -69,7 +69,7 @@ <h2>Pick List - Default</h2>
       </script>
 
       <h2>Pick List - multi-select filter-enabled</h2>
-      <calcite-pick-list id="two" multiple filter-enabled>
+      <calcite-pick-list id="two" multiple filter-enabled filter-sticky>
```
driskull: Can you remove the prop here?

pr3tori4n:
> Should still ensure this works in a Panel in a Popover.

@kat10140 can you assist with this advanced demo during verification?

Issue #506, opened 2019-11-06
Bug: ActionPad got missed during the styling update


asangma:
### Summary
Remove the extra space.

### Actual Behavior
<img width="391" alt="Screen Shot 2019-11-06 at 7 18 24 AM" src="https://user-images.githubusercontent.com/12503298/68311014-c4b34000-0065-11ea-8cbe-ff25286e33cf.png">

### Expected Behavior
![image](https://user-images.githubusercontent.com/12503298/68311133-fcba8300-0065-11ea-9a74-9b478989de24.png)

### Reproduction Steps

1. Demo

### Relevant Info 
Can this be in a patch?

asangma:
Verfee please!

kat10140:
🤙🏼

PR #511, opened 2019-11-06
style(action-pad): added missed styles

asangma:
**Related Issue:** #506

## Summary
Added missing styles to align with calcite-components.

cc @russroberts


asangma:
Thanks @driskull!

Issue #455, opened 2019-10-24
Bug: calcite-block-section -  styling updates for multi-line text


AdelheidF:
### Summary

It's using `calcite-block-section` here with `toggle-display="switch"` and it doesn't look quite right for multi-line text next to the toggle button.

<img src=https://user-images.githubusercontent.com/44481292/67453908-132ffb80-f5dd-11e9-874b-dad749923921.png width=200/>

### Reproduction Steps

http://devext.arcgis.com/apps/mapviewer/index.html

### Relevant Info 

cc @asangma

asangma:
@kat10140 verf plearse.

kat10140:
![Screen Shot 2019-11-07 at 2 24 33 PM](https://user-images.githubusercontent.com/48069902/68432978-57d59e00-016a-11ea-91e7-6be31e8d73f8.png)
@asangma

kat10140:
![Screen Shot 2019-11-07 at 2 27 43 PM](https://user-images.githubusercontent.com/48069902/68433197-cfa3c880-016a-11ea-94d6-76301a2bbf9b.png)
@asangma that might have happened bcuz of that super long, very creative string LOL

kat10140:
![Screen Shot 2019-11-07 at 2 29 40 PM](https://user-images.githubusercontent.com/48069902/68433316-11347380-016b-11ea-87df-ea6859943bfb.png)
its working as long as no one puts in a 50 letter word. but also- if the input is super long it hides the checkbox/input below it

PR #515, opened 2019-11-07
style(block-section): label spacing in switch mode

asangma:
**Related Issue:** #455

## Summary
Added some spacing and RTL spacing for when block-section is in switch mode and the label is long.

Includes some updates to full-window demo.


asangma:
cc @AdelheidF

asangma:
Thanks @driskull Updating to master then mergle.

Issue #385, opened 2019-10-15
Testing: Update e2e tests to include a11y helper


jcfranco:
With #37 installed, e2e tests for all components should be updated to cover a11y.

@driskull 

- [x] `calcite-action` 
- [x] `calcite-action-bar`
- [x] `calcite-action-group`
- [x] `calcite-action-pad`
- [x] `calcite-flow`
- [x] `calcite-flow-item`
- [x] `calcite-panel`
- [x] `calcite-shell`
- [x] `calcite-shell-panel`

@jcfranco 

- [x] `calcite-block`
- [ ] `calcite-block-section`
- [x] `calcite-example`

@pr3tori4n 

- [x] `calcite-filter`
- [x] `calcite-pick-list`
- [x] `calcite-pick-list-group`
- [x] `calcite-pick-list-item`
- [x] `calcite-tip`
- [x] `calcite-tip-group`
- [x] `calcite-tip-manager`
- [x] `calcite-value-list`
~~- [ ] `calcite-value-list-item`~~ doesn't have any tests since it's basically just a wrapping container around `pick-list-item`.

pr3tori4n:
@jcfranco @driskull 
Can we either put component assignments in the first comment for each component or split this into 3 issues, assigned to us individually with the components we each should handle?

jcfranco:
> Can we either put component assignments in the first comment for each component 

👍Please review and update assignments if needed.

pr3tori4n:
`value-list-item` doesn't have any tests since it's basically just a wrapping container around `pick-list-item`.

pr3tori4n:
@jcfranco please verify.

jcfranco:
I missed setting up a test for `calcite-block-section`. I'll create a separate issue for this. Aside from that, all components are using the `accessible` helper.

PR #532, opened 2019-11-08
Hrobbins/a11y e2e tests 385

pr3tori4n:
**Related Issue:** #385 

## Summary

Adding accessibility tests to `pickList` components, `tip` components and `valueList` components.


src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -223,14 +223,13 @@ export class CalcitePickListItem {
       ) : null;
 
     return (
-      <Host selected={this.isSelected}>
+      <Host selected={this.isSelected} role="checkbox" aria-checked={this.isSelected}>
```
pr3tori4n: axe didn't like these attributes on the label tag.

driskull: This is ok, but maybe it should go on the icon? The icon is essentially the replacement for the checkbox.

src/components/calcite-tip-manager/calcite-tip-manager.e2e.ts
```diff
@@ -157,10 +164,10 @@ describe("calcite-tip-manager", () => {
       );
       const tipManager = await page.find("calcite-tip-manager");
       const newTipId = "newTip";
-      await page.evaluate((newTipId) => {
+      await page.evaluate((newId) => {
```
driskull: noice

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
pr3tori4n: We could do that, but then what do we do for valueListItems?

driskull: I'm not sure, but I think the role should be tied to the icon node because the secondary action shouldn't be nested within the checkbox role.

pr3tori4n: As discussed, opening a new issue to investigate further. Moving forward with this in the meantime.

Issue #530, opened 2019-11-08
Enhancement: Add property to toggle opening/closing of TipManager


driskull:
### Description

Enhancement: Add property to toggle opening/closing of TipManager

### Acceptance Criteria 

Currently, the tip manager is closed by adding `hidden` attribute but because we need to prevent the transition animation from happening when reopening, we should have a property that handles opening/closing the tip manager.

### Relevant Info 

This is what happens when reopening after paginating:

![Nov-08-2019 11-08-06](https://user-images.githubusercontent.com/1231455/68503601-1d2a3f00-0218-11ea-85c7-b71dc187ff08.gif)

### Helpful Details

The `@state direction` needs to be removed when closing the TipManager. Since we need to set this, we need a property to open/close.

**Which Component:** TipManager
**Example Use Case:**

Propose adding a `closed` or `dismissed` property to the TipManager.

driskull:
cc @asangma @kevindoshier

kat10140:
Verified

PR #533, opened 2019-11-08
feat: Add property & event for toggling the opening or closing of the TipManager

driskull:
**Related Issue:** #530

## Summary

- Add `closed` property to TipManager
- Add container class for div
- Add new event and deprecate old event
 - Update tests
- Fix shadowed var


src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -28,6 +28,17 @@ export class CalciteTipManager {
...
   // --------------------------------------------------------------------------
+  /**
+   * Alternate text for closing the Tip Manager.
+   */
+  @Prop({ reflect: true }) closed = false;
+
+  @Watch("closed")
+  closedChangeHandler() {
+    this.direction = null;
+    this.calciteTipManagerClosedChange.emit();
```
driskull: Event added so that someone could know if opened/closed.

src/components/calcite-tip-manager/calcite-tip-manager.tsx
```diff
@@ -132,9 +143,15 @@ export class CalciteTipManager {
 
   /**
    * Emitted when the component has been closed.
+   * @deprecated
```
driskull: Deprecate old event.

src/demos/calcite-shell-full-window.html
```diff
@@ -551,7 +551,7 @@ <h3 class="heading" slot="header-content">WE CARE A LOT</h3>
       var tipManagerButton = document.getElementById("tip-manager-button");
       var tipManager = document.getElementById("tip-manager");
       tipManagerButton.addEventListener("click", function(){
-        tipManager.hasAttribute("hidden") ? tipManager.removeAttribute("hidden") : tipManager.setAttribute("hidden", "");
+        tipManager.closed = !tipManager.closed;
```
driskull: image

src/components/calcite-tip-manager/calcite-tip-manager.tsx
pr3tori4n: If the intent of this rename is to indicate a toggling of state - maybe drop the Closed part.

src/components/calcite-tip-manager/calcite-tip-manager.e2e.ts
```diff
@@ -41,18 +41,25 @@ describe("calcite-tip-manager", () => {
...
+      const isVisible = await container.isVisible();
       expect(isVisible).toBe(false);
 
       expect(eventSpy).toHaveReceivedEvent();
+
+      const tipManager = await page.find("calcite-tip-manager");
+
+      const isClosed = await tipManager.getProperty("closed");
+
+      expect(isClosed).toBe(true);
```
pr3tori4n: re-checking visibility here would be better than validating the property value.

driskull: Its already checked above.

Issue #466, opened 2019-10-25
Refactor: Fix Stencil build warnings


jcfranco:
The following warnings started showing up after moving over to Stencil v1.7.4 in `calcite-value-list`, `calcite-pick-list` and `calcite-pick-list-item` E2E tests:

```
%c[STENCIL-DEV-MODE] color:#4c47ff;font-weight: bold 
The state/prop "selected" changed during rendering. 
This can potentially lead to infinite-loops and other bugs. 
```

See https://travis-ci.com/Esri/calcite-app-components/builds/133595791 for more details.

cc @kat10140

driskull:
Installed 🎉

jcfranco:
Verified locally.

PR #537, opened 2019-11-09
fix: Stencil build warnings 

driskull:
**Related Issue:** #466

## Summary

Refactor: Fix Stencil build warnings 

- Makes `selected` a mutable reflected property
- Ties event to anytime `selected` is changed instead of the method being called.
- Adds a private variable `shiftPressed` for when multiple selections are allowed instead of having to pass it to the method
- Removes unnecessary onclick handler for secondary action
- Debounce call to `calciteListChange` event so that multiple succeeding events are not fired.


src/components/calcite-pick-list/shared-list-tests.ts
```diff
@@ -17,7 +17,7 @@ export const tests = {
 
         await item1.click();
         await item2.click();
-        expect(toggleSpy).toHaveReceivedEventTimes(2);
+        expect(toggleSpy).toHaveReceivedEventTimes(listType === "pick" ? 3 : 4);
```
pr3tori4n: Boulder: This is an error. The events are getting duplicated with the code like this.

There may be a simpler way - If we change the attribute name that the state reflects.
```selected={this.isSelected}```
to
```isSelected={this.isSelected}```

Or we can remove it entirely. Now that we're using props and not attributes, the functionality doesn't depend on the existence of this attribute.

Comment (driskull) [reply src/components/calcite-pick-list/shared-list-tests.ts]: I'll research and see why its being called more for value list.

Comment (driskull) [reply src/components/calcite-pick-list/shared-list-tests.ts]: Found out that the items were not properly being matched. Now the event call count is the same because the value is what is being matched.

File: src/components/calcite-pick-list/shared-list-logic.ts [hunk 2]
```diff
@@ -84,7 +84,7 @@ export const sharedListMethods = {
   },
   deselectSiblingItems(this: CalcitePickList | CalciteValueList, item: pickOrValueListItem) {
     this.items.forEach((currentItem) => {
-      if (currentItem !== item) {
+      if (currentItem.value !== item.value) {
```
driskull: This was trying to match a `CalcitePickListItem` to a `CalciteValueListItem`. Changed to use `value` since one wraps the other.

pr3tori4n: Good catch.

src/components/calcite-pick-list/shared-list-tests.ts
```diff
@@ -17,7 +17,7 @@ export const tests = {
 
         await item1.click();
         await item2.click();
-        expect(toggleSpy).toHaveReceivedEventTimes(2);
+        expect(toggleSpy).toHaveReceivedEventTimes(3);
```
pr3tori4n: This is still one time too many. The event should trigger once for every click. 2 Clicks should be 2 events emitted.

driskull: The event is called one more time since it is called when selected is set to false now due to `selected` being the source of truth for that item event.

Previously, setting selected to false on an item didn't trigger an event. Now it does, hence the 1 more event firing. Users should know when something is set to false as well.

pr3tori4n: Hmm. I'm concerned this may cause bugs. If you're reacting to the list changing - you're going to get the same list of selected items twice in this scenario.
I'd expect the `calciteListItemChange` event to fire 3 times (once for selecting item one, once for deselecting item one, and again for selecting item 2).
But the `calciteListChange` event I'd expect to fire twice (one for when item one is the selected item, and once for when item two is the selected item).

driskull: > I'm concerned this may cause bugs. If you're reacting to the list changing - you're going to get the same list of selected items twice in this scenario.

Logically, it shouldn't cause any bugs. Can we not fire the event if the same selected item list exists? That logic would need to be added.

> 2). But the calciteListChange event I'd expect to fire twice (one for when item one is the selected item, and once for when item two is the selected item).

Seems like some kind of throttle would have to happen for this?

pr3tori4n: We could short-circuit the event from firing if the list didn't change but we'd need to start storing the previous state - could be error prone. It's only needed when utilizing the mutable prop method - that's why I avoided that. Depending on what the callback does - it could cause errors in the app, like if it were incrementing some value it would be off. 

Have you given any consideration toward the simpler approach of keeping the prop and state split and changing the name of the prop reflecting the state? Would be a 1-line change.

driskull: > Have you given any consideration toward the simpler approach of keeping the prop and state split and changing the name of the prop reflecting the state? Would be a 1-line change.

If we did that, `isSelected={this.isSelected}`, then `isSelected` would have to be undocumented as a `@Prop` otherwise it would be the same issue. I don't think we should go that route because of that.

driskull: So, we can either...

1. Be OK with the additional event
2. Not emit the event if there is no change.

driskull: @jcfranco can you chime in?

src/components/calcite-pick-list/shared-list-tests.ts
driskull: All edits to tests reverted. Fixed issues.

src/components/calcite-pick-list/shared-list-tests.ts
driskull: Fixed: not emitting event when no selection occurs and its not using multiple.

driskull:
@pr3tori4n @jcfranco could you please re-review?

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -30,6 +31,7 @@ export const sharedListMethods = {
   initialize(this: CalcitePickList | CalciteValueList): void {
     this.setUpItems();
     this.setUpFilter();
+    this.emitCalciteListChangeEventDebounced = debounce(this.emitCalciteListChangeEvent, 10);
```
jcfranco: Debouncing seems like a workaround to an underlying problem. What's causing the multiple event firing?

driskull: The "List" event listens for anytime a "ListChild" selected property changes.

In order to act like a radio button (new selection removes old ones) or when multiple items are selected at the same time, there are multiple selection changes happening which results in multiple "List" event firings.

@pr3tori4n was saying that there should only be one event firing even if multiple items were changed.

jcfranco: Chatted w/ @driskull and this is here to batch all change events within a single tick caused by the internal item updating. Based on that, waiting to notify until the next tick would do the job.

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -30,6 +31,7 @@ export const sharedListMethods = {
   initialize(this: CalcitePickList | CalciteValueList): void {
     this.setUpItems();
     this.setUpFilter();
+    this.emitCalciteListChangeEventDebounced = debounce(this.emitCalciteListChangeEvent, 0);
```
jcfranco: Nitpick: Do we need to expose both the debounced and non-debounced method? Could we do something like this?

***disclaimer: untested code ⚠️***

```tsx
// shared-list-logic.ts
initialize(this: CalcitePickList | CalciteValueList): void {
  // ...
  this.emitCalciteListChange = debounce(this._internalCalciteListChangeEvent, 0);
}

_internalCalciteListChangeEvent(this: CalcitePickList | CalciteValueList): void {
  // previously `emitCalciteListChange` from pick-list and value-list
  this.calciteListChange.emit(this.selectedValues);
}
```

```tsx
// pick-list & value-list
emitCalciteListChangeEvent: () => void;

// no more 👇
// emitCalciteListChangeEvent(): void {
//   this.calciteListChange.emit(this.selectedValues);
// }
```

driskull: noice!

Issue #513, opened 2019-11-06
Enhancement: Panel. Add scale property to set dynamic height


asangma:
### Description
Add an optional property to **panel** which would result in defining the max-height.

Proposal
```
/** specify the scale of the panel, defaults to m */
  @Prop({ reflect: true, mutable: true }) heightScale: "s" | "m" | "l" = "m";
```

```
.container {
  ...
  overflow: auto;
}

:host[height-scale="s"] .container {
 max-height: 33vh;
}

// etc
```

### Acceptance Criteria 
User can define panel scale.
Works with pick-list with a sticky header. (#509)

#### Desired Change to Look and Feel
~Could also add a style to set the width dynamic in line with the shell-panel's responsive width.~
This would need to see an update in calcite-components popover which currently has a max-width set.

#### Out of Scope
Setting panel width.

### Helpful Details

**Which Component:**
**Example Use Case:**
**map viewer beta**
- Currently devs are having to set heights, etc. for each panel...specifically inside popovers.
- Symbol styler could use a mutable scale prop

driskull:
@asangma so this would be for panel, not popover?

asangma:
@driskull Ya.

asangma:
@driskull @kat10140 
I'm gonna start on this for the next release.
I am thinking the property would be specific to height. I also think that down the road a panel could have a property to allow for dynamic width similar to shell-panel for cases when it is in a popover.

Updating initial comment with proposed prop name.

driskull:
Sounds good. I can help if needed. 👍

driskull:
@asangma I don't think we want the default to be "m" or any value right? Because the flow-manager we want to take up 100% height by default. Seems like they should opt into the sizes right?

kat10140:
i tried editing the heightScale in the console in the demos and nothing is happening 😫

PR #541, opened 2019-11-11
feat: CalcitePanel - Add scale property to set dynamic height #513

driskull:
**Related Issue:** #513

## Summary

feat: CalcitePanel - Add scale property to set dynamic height #513

### DIFF
--- a/src/components/calcite-panel/calcite-panel.scss
+++ b/src/components/calcite-panel/calcite-panel.scss
@@ -3,9 +3,14 @@
   height: 100%;
   font-size: var(--calcite-app-base-font-size);
   position: relative;
+
+  --calcite-app-panel-max-height-small: 40vh;
+  --calcite-app-panel-max-height-medium: 60vh;
+  --calcite-app-panel-max-height-large: 80vh;
 }
 
 .container {
+  transition: max-height var(--calcite-app-animation-time) var(--calcite-app-easing-function);
   width: 100%;
   height: 100%;
   padding: 0;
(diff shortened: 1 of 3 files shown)


src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -54,6 +54,11 @@ export class CalcitePanel {
    */
   @Prop({ reflect: true }) dismissible = false;
 
+  /**
+   * Specifies the maxiumum height of the panel.
+   */
+  @Prop({ reflect: true }) heightScale: "s" | "m" | "l";
```
asangma: I think this may need to be mutable so that users can update the size if/when content changes.

pr3tori4n: Would we add separate `scale` props to control other parts of the panel if that became desired in the future? If the intent would be to have one `scale` per component, I'd drop the height part since it may expand beyond that one day.

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -14,6 +14,18 @@
...
+}
+
+:host[height-scale="m"] .container {
+  max-height: 33vh;
+}
+
+:host[height-scale="l"] .container {
+  max-height: 33vh;
+}
+
```
pr3tori4n: @asangma you're going to modify these values right?

src/components/calcite-panel/calcite-panel.tsx
driskull: > Would we add separate scale props to control other parts of the panel if that became desired in the future

Potentially. This only affects panel height so we can either say that in both the property name and the readme description or just the readme description.

asangma: The property having a name specific to the height is probably good since it clearly excludes width.

This is one of the places where we probably want to treat the width like we treat the width of the shell-panels, e.g. the developer shouldn't need to worry too much about it.

src/demos/calcite-shell-full-window.html
```diff
@@ -541,6 +541,14 @@ <h3 class="heading" slot="header-content">WE CARE A LOT</h3>
           <calcite-pick-list-item text-label="about the NASA shuttle falling in the sea" value="2"></calcite-pick-list-item>
           <calcite-pick-list-item text-label="about starvation and the food that Live Aid bought" value="3"></calcite-pick-list-item>
           <calcite-pick-list-item text-label="about disease, baby rock, Hudson, rock," value="4"></calcite-pick-list-item>
+          <calcite-pick-list-item text-label="about disease, baby rock, Hudson, rock," value="4"></calcite-pick-list-item>
```
jcfranco: Nitpick: Should these be different? If not, can the label be updated to reflect why they're repeated?

driskull: Updated with lorem ipsum

Issue #545, opened 2019-11-11
Enhancement: document default slots: pick and value list


kat10140:
### Description
Document the default slot for the pick and value list. While you're in there maybe go ahead and note what properties are optional and what are required.

#### Which Component
Pick-list and value-list

driskull:
We don't document any default slots. Should we start doing so?

pr3tori4n:
I definitely think so. It's a quality of life improvement for someone trying to implement our components, and aren't super clear how they work.

kat10140:
verified

PR #549, opened 2019-11-13
docs: add jsdocs for slot's for the list elements

pr3tori4n:
**Related Issue:** #545 

## Summary

Add missing JSDoc information for the default slots and a few others.


src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -1,6 +1,9 @@
 import { Component, Element, Host, Method, Prop, h } from "@stencil/core";
 import { ICON_TYPES } from "../calcite-pick-list/resources";
 
+/**
+ * @slot secondaryAction - A slot for adding a calcite-action or other button. Placed at the end of the item.
```
driskull: Maybe `A slot intended for adding a calcite-action or calcite-button`?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -29,6 +29,7 @@ const {
 } = sharedListMethods;
 
 /**
+ * @slot default - A slot for adding pick-list-item elements or pick-list-groups elements. Items are displayed as a vertical list.
```
jcfranco: Looks like Ionic doesn't provide a name for the default slot: https://github.com/ionic-team/ionic/blob/34dfc3ce98d1e30af7e2f82276a63bc077726773/core/src/components/button/button.tsx#L12 I think we should follow suit to avoid a "default" named slot in the doc.

src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
```diff
@@ -1,6 +1,9 @@
 import { Component, Host, Prop, h } from "@stencil/core";
 import { CSS } from "./resources";
 
+/**
+ * @slot default - A slot for adding pick-list-item elements.
```
driskull: `A slot for adding calcite-pick-list-item elements.` Or do we want to say `HTMLCalcitePickListItem`?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -15,6 +15,9 @@ import { CSS } from "./resources";
 import { ICON_TYPES } from "../calcite-pick-list/resources";
 import CalciteIcon from "../utils/CalciteIcon";
 
+/**
+ * @slot secondaryAction - A slot for adding a calcite-action or other button. Placed at the end of the item.
```
driskull: Maybe A slot intended for adding a calcite-action or calcite-button?

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: Hmmm. Maybe there's a better way to convey this info, or a better name? I feel it's valuable information to provide to anyone who's using one of these components for the first time, to both understand the required props and the expected children. It's like needing to know that if you're using a <select> tag, you need to fill it with <option> elements that have `value` attributes.

driskull: I think the "usage" folder will help out here once @kat10140 can get to those doc examples

src/components/calcite-pick-list-group/calcite-pick-list-group.tsx
pr3tori4n: I like the tagname since that's what our users will be typing out.

src/components/calcite-value-list/calcite-value-list.tsx
jcfranco: I'm not denying that it's valuable. It's just that the current tooling will generate a string field for the slot name when provided and could be misleading.

driskull: I agree, we don't want to put a name here. Otherwise users may do this: `<div slot="default"></div>`

pr3tori4n: Ahhhh so you're saying just do it like this: ` * @slot - A slot for...`?

jcfranco: Yes.

driskull:
What about the other component unnamed slots?

driskull:
https://github.com/Esri/calcite-app-components/issues/561

Issue #554, opened 2019-11-14
Bug: panel header + action height is weird when content is scrolling


asangma:
### Actual Behavior
![panel min-height](https://user-images.githubusercontent.com/12503298/68816040-147ba380-0632-11ea-8744-f8b8486f09d6.gif)

### Expected Behavior
No border weirdness.

### Reproduction Steps

1. Flow demo

### Relevant Info 
Could set flex-basis instead of min-height.
Could also set flex-basis based on the height of Action which is basically our size basis (16px) * 3.

asangma:
This was not actually fixed by the referencing PR.
![panel header min-height](https://user-images.githubusercontent.com/12503298/69077080-aea06a80-09e9-11ea-9d6e-83558cd2ec62.gif)

kat10140:
PR #591

asangma:
@kat10140 VERFEE

kat10140:
verified!

Issue #547, opened 2019-11-12
Bug: panel. Don't render header leading or trailing nodes when empty


asangma:
### Summary
re: header. Only render related notes when content is passed to leading and trailing slots in header.

### Actual Behavior
When leading slot is empty, flex justify-content still acts like there's stuff in there.
![image](https://user-images.githubusercontent.com/12503298/68703325-45c77700-053f-11ea-866c-e99bfe541bb6.png)

### Expected Behavior
![image](https://user-images.githubusercontent.com/12503298/68703334-4bbd5800-053f-11ea-845b-4f87a25a2e23.png)

### Reproduction Steps

1.

driskull:
@asangma if we remove the nodes when the slot is not populated the `close` button would be to the left. How would we address that?

asangma:
The header is `justify-content: space-between`. It should automatically layout like the above screen if there's no leading slot. I will have to check what it looks like when there's not trailing. Either way, it looks weird currently.

asangma:
@kat10140 Please verfee. :)

kat10140:
verified!

PR #556, opened 2019-11-14
fix(panel): header-content layout when leading slot is empty

asangma:
**Related Issue:** #547 

## Summary
* Added a ternary to only render `header-leading-content` node when slot has content
* Updated related styles
* Includes a small fix for #554 

## Before
image

### After
image

### DIFF
--- a/src/assets/styles/_base.scss
+++ b/src/assets/styles/_base.scss
@@ -19,7 +19,7 @@
   display: flex;
   align-items: center;
   justify-content: space-between;
-  min-height: var(--calcite-app-header-min-height);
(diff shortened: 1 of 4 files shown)


asangma:
Hmmm...good point. But would that be a breaking change?

driskull:
Yes, it would be a breaking change but seems like one that would benefit in the future.

asangma:
Pre Slack conversation, removing the `header-content` slot will be a separate issue to be discussed in the [regular sync](https://github.com/Esri/calcite-app-components/wiki/2019-Weekly-Meeting#2019-11-15).

driskull:
YUP. good to merge

Issue #356, opened 2019-10-09
Enhancement: ValueList - Add ability to drag and drop with keyboard


asangma:
### Summary
User should be able to tab into the drag handle, hit space, then use the up/down arrows to move items.

### Actual Behavior
Can't tab into drag handle.

### Expected Behavior
Be able to tab into drag handle.

### Reproduction Steps

1. demo page

### Relevant Info 
See how Bookmarks widget works in edit mode.

cc @driskull @kat10140

pr3tori4n:
@driskull did you use sortable for the Bookmarks widget? If so can I scavenge the code you used to get that working - can you share a code snippet?

driskull:
Yes, its in the bookmarks view.

pr3tori4n:
Modifying this to be an enhancement issue, since its a request for additional functionality that wasn't present on first pass.

pr3tori4n:
@kat10140 please verify. Please double-check Edge and Safari.

kat10140:
works in Chrome, Firefox, Safari, Edge. Closing

PR #559, opened 2019-11-14
feat: value list drag drop keyboard support 356

pr3tori4n:
**Related Issue:** #356 

## Summary

The drag handles can now be tabbed into, activated by pressing "Space" and then moved up and down by pressing "ArrowUp" and "ArrowDown" keys. Blurring away from a handle automatically de-activates it.

Note: @asangma the drag handle is now adjacent to the pick-list-item instead of being nested inside it. I was able to style it to pretty much match the previous styling. If we were to move it outside the value-list-item however, thing's get a bit more interesting.

Did the work for splitting the valueList demos into simple and advanced in this branch. Will do the others in `hrobbins/advanced-demos`


src/components/calcite-value-list-item/calcite-value-list-item.scss
```diff
@@ -18,3 +19,24 @@ calcite-pick-list-item[selected] {
...
+  padding: var(--calcite-app-cap-spacing-three-quarters) var(--calcite-app-side-spacing-half);
+  background-color: var(--calcite-app-background);
+  color: var(--calcite-app-foreground-subtle);
+  line-height: 0;
+  cursor: move;
+  @include borderShadow();
+  &--activated {
+    background-color: black;
+    color: white;
+  }
```
pr3tori4n: @asangma please update these styles. They are placeholder.

src/components/calcite-value-list/calcite-value-list.e2e.ts
```diff
@@ -43,4 +43,54 @@ describe("calcite-value-list", () => {
...
+    beforeEach(async () => {
+      page = await newE2EPage();
+      await page.setContent(`<calcite-value-list multiple drag-enabled>
+        <calcite-value-list-item value="one" text-label="One"></calcite-value-list-item>
+        <calcite-value-list-item value="two" text-label="Two"></calcite-value-list-item>
+        <calcite-value-list-item value="three" text-label="Three"></calcite-value-list-item>
+      </calcite-value-list>`);
+    });
+    it.skip("works using a mouse", async () => {
+      // TODO: remove skip once https://github.com/GoogleChrome/puppeteer/issues/1376 addressed
```
pr3tori4n: See link for why this test won't work - there's a bug. Leaving it in but skipping.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -73,6 +81,13 @@ export class CalciteValueListItem {
     this.pickListItem.toggleSelected(coerce, emit);
   }
 
+  /*
+   ** @internal
+   */
+  @Method() async activateHandle() {
+    this.handleActivated = true;
```
driskull: fixme: Do we need a method for this if its just a boolean? Can we handle with a property?

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -81,22 +96,51 @@ export class CalciteValueListItem {
...
   //
   //  Render Methods
   //
   // --------------------------------------------------------------------------
 
+  renderHandle() {
+    const { icon } = this;
+    if (icon === ICON_TYPES.grip) {
+      return (
+        <a
```
driskull: fixme: Can you make this a `div` with a button role?  I don't think an anchor makes sense semantically.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -81,22 +96,51 @@ export class CalciteValueListItem {
...
+          class={{ [CSS.handle]: true, [CSS.handleActivated]: this.handleActivated }}
+          tabindex="0"
+          data-js-handle="true"
```
driskull: question: What is this data attribute for?

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -81,22 +96,51 @@ export class CalciteValueListItem {
...
+          data-activated={this.handleActivated}
```
driskull: fixme: Instead of `data-activated` can you use `aria-pressed`?

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -81,22 +96,51 @@ export class CalciteValueListItem {
...
+      <Host data-id={this.guid}>
+        {this.renderHandle()}
         <calcite-pick-list-item
           compact={this.compact}
           ref={this.getPickListRef}
           disabled={this.disabled}
           selected={this.selected}
           metadata={this.metadata}
-          icon={this.icon}
+          icon={null}
```
driskull: question: Do you need to pass null here?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -202,6 +203,38 @@ export class CalciteValueList {
 
   getItemData = getItemData.bind(this);
 
+  keyDownHandler = (event) => {
+    const handleElement = event.composedPath().find((item) => {
```
driskull: question: composedPath doesn't work in edge. Is this polyfilled?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -202,6 +203,38 @@ export class CalciteValueList {
...
+    }
+    const value = event.target.value;
+    const startingIndex = this.items.findIndex((item) => {
+      return item.value === value;
+    });
+    let newIndex;
+    switch (event.key) {
+      case "ArrowUp":
+        newIndex = startingIndex - 1;
+        newIndex = newIndex < 0 ? this.items.length - 1 : newIndex;
```
driskull: nitpick: `this.items.length` could be stored as a const.

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -202,6 +203,38 @@ export class CalciteValueList {
 
   getItemData = getItemData.bind(this);
 
+  keyDownHandler = (event) => {
+    const handleElement = event.composedPath().find((item) => {
+      return item.dataset && item.dataset.jsHandle;
```
driskull: nitpick: I think we can use optional chaining now with latest TS.

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -202,6 +203,38 @@ export class CalciteValueList {
...
+        newIndex = newIndex < 0 ? this.items.length - 1 : newIndex;
+        break;
+      case "ArrowDown":
+        newIndex = startingIndex + 1;
+        newIndex = newIndex > this.items.length - 1 ? 0 : newIndex;
+        break;
+      default:
+        return;
+    }
+    const order = this.sortables[0].toArray();
```
driskull: question: can there be multiple sortables?

src/components/utils/dom.ts
```diff
@@ -11,6 +11,13 @@ export function nodeListToArray(domNodeList): Element[] {
   }
 }
 
+export function arraymove(arr, fromIndex, toIndex) {
```
driskull: fixme: rename to `arrayMove`?

Is there a lodash function we can use for this? Does this belong in a dom util if its just for arrays?

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: Would an actual button be more appropriate?

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: My concern with a prop is that it might show up as part of the API and the user really shouldn't mess with this. Will try a prop with @internal to try and omit docs about it. I remember having issues with that before though.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
driskull: You could, but you'll have to stop the click event from triggering.

pr3tori4n: outside of a form and with no click handler - would there be any effect?

driskull: Clicking shouldn't trigger handle activation

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: It's used as a hook to verify the keyDown event originated from this element.
https://github.com/Esri/calcite-app-components/blob/hrobbins/valueList-dragDrop-keyboard-support-356/src/components/calcite-value-list/calcite-value-list.tsx#L207-L213

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: doubled-checked. The default is null. Removing.

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: I believe that comes with version 3.7, and seems we're using 3.6.3. Is there an update in master that bumps this?

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: Originally, when the architecture needed to support subgroups, I had to create a sortable for each group, but group them using the group prop. Since that requirement was dropped, The heirarchy flattened and there's only one now.

src/components/utils/dom.ts
pr3tori4n: might be a better home for it. It's too simple to bother loading lodash for imo. I'm honestly surprised there isn't a native DOM equivalent.

renamed.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: @internal prop seems to be working. Going forward with it.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: It won't because I'm listening to onKeyDown on the button to activate it.

src/components/calcite-value-list/calcite-value-list.tsx
driskull: We're on 3.7.2 on master.

src/components/utils/dom.ts
driskull: Can you move it into an array.ts file?

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: yea it is...but testing in edge I'm seeing a different issue, the event.target is different because the shadow dom redirect doesn't happen...must fix.

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: sweet, merged it in. optional chaining ftw.

src/components/calcite-value-list-item/calcite-value-list-item.scss
asangma: Will do. I think we should use an explicit name e.g. `--dragging`.

pr3tori4n: Name can change. Just bear in mind this is the state for after a use has focused the handle, and pressed space to activate keyboard sorting. The actual dragging isn't animated when using the keyboard, the positions jump to their new spots.

asangma: @pr3tori4n Ah...nvmd. I see that it's related to the handle.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -183,27 +183,19 @@ export class CalcitePickListItem {
...
+    const path =
+      icon === ICON_TYPES.square
+        ? selected
+          ? checkSquare16
+          : square16
+        : selected
+        ? circleFilled16
+        : circle16;
+    return (
+      <span class="icon">
```
driskull: nitpick: Can we add this class to the CSS object?

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -19,32 +23,37 @@ export class CalciteValueListItem {
   /**
    * Compact reduces the size of the item.
    */
-  @Prop({ reflect: true }) compact = false;
+  @Prop({ reflect: true }) compact? = false;
```
driskull: nitpick: If you set the value to false, the optional flag won't do anything.

driskull: You would need to do `@Prop({ reflect: true }) compact?: boolean;`

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -199,6 +199,56 @@ export class CalciteValueList {
 
   getItemData = getItemData.bind(this);
 
+  keyDownHandler = (event) => {
+    const handleElement = event.composedPath().find((item) => {
+      return item.dataset?.jsHandle;
+    });
+    const valueListElement = event.composedPath().find((item) => {
```
driskull: nitpick: maybe this could be a method in the dom utility?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -199,6 +199,56 @@ export class CalciteValueList {
...
+    // Only trigger keyboard sorting when the internal drag handle is focused and activated
+    if (!handleElement || !valueListElement.handleActivated) {
+      return;
+    }
+    const lastIndex = this.items.length - 1;
+    const value = valueListElement.value;
+    const startingIndex = this.items.findIndex((item) => {
+      return item.value === value;
+    });
+    let appendInstead = false;
```
driskull: nitpick: Maybe the logic below could be in a private method that either returns an index or null if it should append?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -199,6 +199,56 @@ export class CalciteValueList {
...
+      return item.dataset?.jsHandle;
+    });
+    const valueListElement = event.composedPath().find((item) => {
+      return item.tagName?.toLowerCase() === "calcite-value-list-item";
+    });
+    // Only trigger keyboard sorting when the internal drag handle is focused and activated
+    if (!handleElement || !valueListElement.handleActivated) {
+      return;
+    }
+    const lastIndex = this.items.length - 1;
```
driskull: nitpick: `this.items` is referenced a few times. maybe have a const for items?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -224,6 +274,7 @@ export class CalciteValueList {
   }
 
   render() {
-    return <List props={this} text={TEXT} />;
+    console.log("render");
```
driskull: fixme: remove console.log :)

src/components/calcite-value-list-item/calcite-value-list-item.tsx
driskull: Applies to other places.

pr3tori4n: When you say it won't do anything, what do you mean. Do you mean it won't appear as optional in the docs?

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: possibly, the inner function may need to be unique. If we start seeing more of a pattern arise then we could abstract this.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
driskull: correct

src/components/calcite-value-list/calcite-value-list.tsx
pr3tori4n: I created a function called moveNode and added it to the dom utils but I"m not using it right now. One of the particular challenges with this scenario is there are elements that are valid dom children that we have to skip. I'd be down to go over it with you to see if there's something we can't make work. But in its current form it wouldn't skip the header - which is why `sortable.sort` was breaking in Edge.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
pr3tori4n: Ok. I think having the value is more valuable. So long as the required props that have no default value show up as required in the docs I think that will suffice.

driskull: Ok. You can remove the `?` then if you set a default value. There's no reason to say its optional when its set automatically.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -84,22 +95,51 @@ export class CalciteValueListItem {
...
+          <CalciteIcon size="16" path={drag16} />
+        </span>
+      );
+    }
+  }
+
   render() {
     return (
-      <Host>
+      <Host data-id={this.guid}>
```
driskull: fixme: I'm hesitant about polluting the dom with data attributes unless the data-attribute is within the shadow dom. If we need it to be a property can we just setup an `id` or `guid` prop for the unique identification?

pr3tori4n: This is what sortable uses by default to identify items for sorting.
https://github.com/SortableJS/Sortable#options
See `dataIdAttr`

driskull: Ok. I guess this can be an exception.

Issue #463, opened 2019-10-25
Docs: Clean up demo pages


jcfranco:
## Summary

Demo pages should be simple and focus on their own component.

## Measure of Success

**Pain point:**

Some demo pages have more than one component that is unrelated and adds complexity (e.g., [calcite-block.html](https://github.com/Esri/calcite-app-components/blob/master/src/demos/calcite-block.html). These should be cleaned up and possibly moved over to one or more advanced demo pages.

**Desired outcome:**

Simpler demo pages and, optionally, addition of advanced pages.

pr3tori4n:
@kat10140 please verify.

kat10140:
verified

PR #562, opened 2019-11-14
advanced demos 463

pr3tori4n:
**Related Issue:** #463

PSA: I'd recommend pulling down the branch and reviewing the demo pages in a local browser. You'll get a much clearer idea of what's been changed than by comparing `.html` files.

## Summary

- Simplified any demo pages that had complex data or integrations. Each demo page should only show a demo of the 1 component, so it's easier to differentiate the behavior of one another.
  - In some cases, I replaced child components with a placeholder image, to fill out space.
  - I also added a few demos to show how some props worked that were absent or combined with another demo.
- Any demo pages that have complex data or multiple component integrations were moved under a new advanced demo dropdown. Two exceptions - the action-bar and action-pad, which are pretty much just wrappers around actions.
- I removed all custom styling from the demo pages that constrained widths. The main goal here is to remove any "magic" that's happening in getting a component to look a certain way - such as appear to be only 400px wide, when it's default behavior is to fill up its container. 
- In place of custom, invisible, width constrained containers, I put a consistent container box that's the same size for all pages and has a visual chrome to set it apart from the component itself. This new container is resizable horizontally, so it's easy enough to resize it to see how the components look under more constrained widths (to test line wrapping, etc.)

### Extras
- I abstracted the disabled and loading toggles, so these don't need to be copy-pasted to every file going forward.
- For the flow-item advanced demo, a big part of it is custom JavaScript that's needed for implementation. I've introduced a CDN link to prism.js and added a simple `<details>` element to act as an accordion to demo the script.
- I updated the demoPageSync.js node script to use chokidar instead of fs.watch. fs.watch was firing off several duplicate events (see image). This has corrected a number of permissions errors I was getting. There's still a glitch where sometimes the `.html` file in the `www` folder won't get updated, but another `CTRL+S` usually remedies that.
fsWatch_vs_chokidar


src/components/calcite-action-bar/calcite-action-bar.scss
```diff
@@ -1,6 +1,6 @@
 :host {
   height: 100%;
-  display: flex;
+  display: inline-flex;
```
pr3tori4n: The action bar was fillings its container. It seemed that behavior was *undesired* so I changed this to inline-flex to remedy that.

src/components/calcite-action-pad/calcite-action-pad.scss
```diff
@@ -1,5 +1,5 @@
 :host {
-  display: flex;
+  display: inline-flex;
```
pr3tori4n: The action pad was fillings its container. It seemed that behavior was *undesired* so I changed this to inline-flex to remedy that.

src/components/calcite-flow/calcite-flow.tsx
```diff
@@ -21,6 +21,11 @@ export class CalciteFlow {
    */
   @Prop({ reflect: true }) theme: CalciteTheme;
 
+  /**
+   * When true, Flow responds to the back event of flow item
+   */
+  @Prop({ reflect: true }) handleBack = false;
```
pr3tori4n: The flow seemed to be missing an implementation that didn't require additional boilerplate to wire it up. This new config option provides a simple means of seeing the flow in action. `false` by default.

driskull: fixme: I'd prefer using a promise instead of this.

```tsx
@Prop() beforeBack: (el: HTMLElement) => Promise<void> = () =>
```

See: https://github.com/Esri/calcite-components/blob/master/src/components/calcite-modal/readme.md

src/components/calcite-flow/calcite-flow.tsx
```diff
@@ -53,6 +58,18 @@ export class CalciteFlow {
...
   }
 
+  // --------------------------------------------------------------------------
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  @Listen("calciteFlowItemBackClick") handleCalciteFlowItemBackClick() {
+    if (this.handleBack) {
```
driskull: fixme: Should be like this: https://github.com/Esri/calcite-components/blob/master/src/components/calcite-modal/calcite-modal.tsx#L185

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -6,7 +6,6 @@ import { CalciteLayout } from "../interfaces";
 
 /**
  * @slot action-bar - A slot for adding a `calcite-action-bar` to the panel.
- * @slot action-pad - A slot for adding a `calcite-action-pad` to the panel. The action pad will be positioned relative to the shell panel when displayed.
```
driskull: question: 😱 should we consider breaking change?

src/demos/toggles.js
```diff
@@ -0,0 +1,43 @@
+(function() {
```
driskull: nitpick: Should we have this in typescript? :)

support/demoPageSync.js
```diff
@@ -1,24 +1,38 @@
...
-  }
+const indexWatcher = chokidar.watch(sourceIndexFile, {
+  ignored: /(^|[\/\\])\../, // ignore dotfiles
+  persistent: true,
+  ignoreInitial: true
+});
+indexWatcher.on('change', (path) => triggerBuild( sourceDir, destinationDir, path ) );
 
-  copyFileSync(`${sourceDir}${changedFile}`, `${destinationDir}${changedFile}`);
+// watch(sourceDemoDir, (type, changedFile) => triggerBuild(sourceDemoDir, destinationDemoDir, changedFile));
```
driskull: fixme: can we remove comments?

src/demos/advanced/flow-adding-items.html
```diff
@@ -0,0 +1,140 @@
...
+          </div>
+          <div slot="footer-actions">
+            <button>Save</button>
+            <button>Cancel</button>
+          </div>
+        `;
+        flowNode.appendChild(newNode);
+      });
+    </script>
+    <script src="https://unpkg.com/prismjs@1.17.1/components/prism-core.min.js"></script>
```
driskull: fixme: I don't think we need prism, since these will be used as demo/test-apps instead of the public facing storybook and docs which can have the code samples in markdown.

src/demos/demos.css
```diff
@@ -6,7 +6,29 @@ body {
 .example-container {
   padding: 0 24px 24px;
 }
+@media (min-width: 1025px) {
```
asangma: A background color does help see where the component ends, but I'm not sure if a restricted container inside of the window helps.

Also feel free to access our color vars like var(--calcite-app-background-content) for the general gray bg-color.

src/demos/toggles.template
```diff
@@ -0,0 +1,12 @@
+<template>
+  <style>
+    .toggles {
+      background-color: rgba(51,51,51,0.2);
```
asangma: Consider
```
background-color: var(--calcite-app-background-content);
```

src/demos/calcite-shell.html
```diff
@@ -37,567 +44,56 @@ <h2>Demo Navigation</h2>
           <li>View the <a href="calcite-shell-full-window.html" target="_blank">Full window demo</a>.</li>
```
asangma: I believe this link needs to be updated.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
pr3tori4n: The slot is absent. Seems this doc line never got cleaned up. So I don't think it's a breaking change. just a Doc Typo.

src/demos/advanced/shell-demo-app-full-window.html
```diff
@@ -0,0 +1,595 @@
...
+          container: "viewDiv",
+          map: map,
+          zoom: 4,
+          center: [15, 65]
+        });
+      });
+    </script>
+  </head>
+  <body>
+    <nav>
```
asangma: full-window demo should show what an app would actually look like in the full window. Should nix the nav and toggles.

src/demos/demos.css
asangma: I do agree that setting a max-width on the `example-container` would be good, but we can probably set it's width to a reasonable vw value or something.

src/components/calcite-demo-nav/calcite-demo-nav.tsx
```diff
@@ -1,19 +1,22 @@
 import { Component, Host, Prop, State, h } from "@stencil/core";
```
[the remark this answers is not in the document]
pr3tori4n: We'll still want a simple way to navigate between them. Perhaps we can move it back to a native web component and just house it under the demos folder.

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -83,6 +83,9 @@ export class CalciteFilter {
       return found;
     };
 
+    if (this.data.length === 0) {
+      console.warn("No data was passed to calcite-filter");
```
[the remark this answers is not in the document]
pr3tori4n: This shouldn't have been in this PR. removing.

src/components/calcite-flow/calcite-flow.tsx
pr3tori4n: discussed with @driskull and am going to move this to a separate PR, because the modified approach would be a breaking change.

src/demos/toggles.js
pr3tori4n: This folder doesn't get built automatically right now I don't believe. We'd have to introduce a new step in the build to do that. personally, it doesn't seem worth it right at this moment, but I think it may be a good step in the future. You ok with holding off for now?

src/demos/advanced/flow-adding-items.html
pr3tori4n: k. removing.

src/demos/demos.css
pr3tori4n: Used some calcite colors. There a specific width you were thinking?
I'm happy to adjust this - the aim is to give a frame of reference for the components, and the gist I got from the inline widths used in so many demos was full width of the screen was too wide.

src/demos/toggles.template
pr3tori4n: fixed

src/demos/toggles.js
driskull: Yep. hold off!

src/demos/calcite-block.html
```diff
@@ -54,96 +41,19 @@ <h2>Icon + Header</h2>
 
         <h2>Header + content (always open)</h2>
 
-        <calcite-block heading="Header" open>
-          <div>Some content</div>
+        <calcite-block heading="Header is longer so it breaks onto two lines" open>
+          <div>Some content that's really interesting. So interesting you're not yawning right now.</div>
```
jcfranco: 😂

src/demos/calcite-shell.html
pr3tori4n: I ended up removing these in most pages (missed this one) since we have them on the main nav under 'advanced'. Lmk if you see value in keeping some inline links to the other shell demos.

support/demoPageSync.js
```diff
@@ -1,24 +1,35 @@
...
+const chokidar = require('chokidar');
 
 const projectRoot = normalize(`${__dirname}/../`);
-const sourceDir = `${projectRoot}src/`;
-const sourceDemoDir = `${sourceDir}demos/`;
-const destinationDir = `${projectRoot}www/`;
-const destinationDemoDir = `${destinationDir}demos/`;
-const sourceIndexFile = `${sourceDir}index.html`;
-const buildTriggeringFile = `${sourceDir}components/interfaces.ts`;
+const sourceDir = normalize(`${projectRoot}src/`);
```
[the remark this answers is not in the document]
pr3tori4n: I needed to normalize the directories used in the replace function, since the strings had different slashes (mix of forward and backward). I supposed we could remove it from `sourceIndexFile` and `buildTriggeringFile`. Want me to?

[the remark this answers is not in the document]
jcfranco: Gotcha. Nah, we can leave it.

pr3tori4n:
@asangma I've addressed your feedback and have some follow up questions.

src/demos/toggles.js
pr3tori4n: all yours bro

Issue #560, opened 2019-11-14
Bug: calcite-block disabled state doesn't 'hide' slider handle


AdelheidF:
### Summary

![image](https://user-images.githubusercontent.com/44481292/68901696-3edd6780-06eb-11ea-9e56-fd54c69514dd.png)

cc @asangma @jcfranco 

### Reproduction Steps

1.

jcfranco:
The scrim currently uses a z-index of 2, so there are cases where siblings in the same stacking context will show above it. I propose for now a simple CSS var for the scrim's z-index and if necessary, we can create a follow-up issue for something more involved to guarantee that it's always above. 

@asangma @pr3tori4n @driskull WDYT?

driskull:
@jcfranco This is the current structure of the scrim

```tsx
<div class={CSS.content} hidden={!hasContent || !open}>
   <slot />
  {loading || disabled ? <CalciteScrim loading={false}></CalciteScrim> : null}
</div>
```

Can we just do the following to get the stacking index relative to where the scrim is?

```tsx
<div class={CSS.content} hidden={!hasContent || !open}>
  <div style="position:relative; z-index:1">
     <slot />
  </div>
  <div>
  {loading || disabled ? <CalciteScrim loading={false}></CalciteScrim> : null}
</div>
```

driskull:
We could also enhance `<CalciteScrim>` to surround the content you want contained in the right z-index stacking. That might be even better.

jcfranco:
@AdelheidF Can you help verify this one?

AdelheidF:
Tested locally. Seems fine now.

<img src=https://user-images.githubusercontent.com/44481292/69205162-10ea8f80-0afe-11ea-88a8-cc8a557f3ad5.png width=200/>

PR #567, opened 2019-11-15
fix: calcite-block disabled state doesn't 'hide' slider handle #560

driskull:
**Related Issue:** #560

fix: calcite-block disabled state doesn't 'hide' slider handle #560

### DIFF
--- a/src/components/calcite-block/calcite-block.tsx
+++ b/src/components/calcite-block/calcite-block.tsx
@@ -189,18 +189,14 @@ export class CalciteBlock {
       </div>
     );
 
-    const slotNode = <slot />;
-
     return (
       <Host>
         <article aria-expanded={collapsible ? (open ? "true" : "false") : null} aria-busy={loading}>
           {headerNode}
           <div class={CSS.content} hidden={!hasContent || !open}>
-            {loading || disabled ? (
-              <CalciteScrim loading={false}>{slotNode}</CalciteScrim>
-            ) : (
-              slotNode
-            )}
+            <CalciteScrim loading={false} disabled={disabled}>
+              <slot />
+            </CalciteScrim>
           </div>
         </article>
       </Host>
--- a/src/components/calcite-block/calcite-block.tsx
+++ b/src/components/calcite-block/calcite-block.tsx
@@ -194,7 +194,7 @@ export class CalciteBlock {
         <article aria-expanded={collapsible ? (open ? "true" : "false") : null} aria-busy={loading}>
           {headerNode}
           <div class={CSS.content} hidden={!hasContent || !open}>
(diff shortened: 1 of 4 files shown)


driskull:
@jcfranco if this approach sounds good, I can update all the other files that use `CalciteScrim`

jcfranco:
This changes the design slightly by hiding all the content. cc @asangma 

Does the shuffling of the slot impact the block's size?

driskull:
> This changes the design slightly by hiding all the content

What do you mean?

> Does the shuffling of the slot impact the block's size?

It shouldn't. the div only has positioning of relative and z-index.

src/components/calcite-block/calcite-block.tsx
```diff
@@ -189,13 +189,18 @@ export class CalciteBlock {
...
+
     return (
       <Host>
         <article aria-expanded={collapsible ? (open ? "true" : "false") : null} aria-busy={loading}>
           {headerNode}
           <div class={CSS.content} hidden={!hasContent || !open}>
-            <slot />
-            {loading || disabled ? <CalciteScrim loading={false}></CalciteScrim> : null}
+            {loading || disabled ? (
+              <CalciteScrim loading={false}>{slotNode}</CalciteScrim>
```
driskull: @jcfranco I could always put the slot inside the scrim since the loader is only shown when its loading but then it would always have a position:relative and zIndex of 1 for the container. That might be ok though.

jcfranco:
Brain fart, never mind. I like this.

src/components/utils/CalciteScrim.tsx
```diff
@@ -16,10 +16,22 @@ const divStyle = {
...
-export const CalciteScrim: FunctionalComponent<ScrimProps> = ({ loading }) => (
-  <div style={divStyle}>{loading ? <calcite-loader is-active></calcite-loader> : null}</div>
-);
+const containerDivStyle = {
+  position: "relative",
+  zIndex: "1"
+};
+
+export const CalciteScrim: FunctionalComponent<ScrimProps> = ({ loading, disabled }, children) =>
+  disabled || loading
```
driskull: moved this into the functional component because it was repetitive throughout the other components. It won't return anything if its not disabled or loading.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -215,7 +209,7 @@ export class CalcitePanel {
           {this.renderContent()}
           {this.renderFooter()}
         </article>
-        {this.renderScrim()}
+        <CalciteScrim loading={loading} disabled={disabled}></CalciteScrim>
```
jcfranco: Can you self-close childless tags?

Issue #569, opened 2019-11-16
A11y: Action focus ring gets largely hidden


asangma:
### Summary
Action focus ring in ActionBar gets hidden.

### Actual & Expected Behavior
![action focus](https://user-images.githubusercontent.com/12503298/68984807-14f97300-07c7-11ea-91d8-13ccfa961583.gif)

### Reproduction Steps

1. focus an Action in a ActionBar.

asangma:
cc @arjavbadjatiya

asangma:
@kat10140 verfee!

kat10140:
Verfeeeeeeeeeeee

PR #570, opened 2019-11-16
fix(action): focus outline a11y

asangma:
**Related Issue:** #569

## Summary
Outline inset some to make sure focus is visible when action buttons are right next to each other.

@driskull @jcfranco  Hoping this can get into a patch for the beta. JB brought it to our attention.


asangma:
cc @arjavbadjatiya

asangma:
Merp!

Issue #333, opened 2019-10-03
Enhancement: PickList/ValueList Filter Label


pr3tori4n:
### Description
I've put the placeholder on an aria-label attribute for now. I'm opening this issue to tackle what sort of labeling we want to support for the internal filters. Do we want to add a prop that allows the pickList author to set the label? Do we need to add a slot for this to support HTML or will plain text work? 
### Acceptance Criteria 
~@asangma @kat10140 could you do some requirements gathering for this?~

- text label default value is going to be just 'filter' 
- remove text label from dom

### Helpful Details

**Which Component:** Filter

pr3tori4n:
@kat10140 please verify.

kat10140:
Verified

PR #574, opened 2019-11-20
removed text node from fitler label.

pr3tori4n:
**Related Issue:** #333


pr3tori4n:
@driskull There is a prop `textLabel`
https://github.com/Esri/calcite-app-components/blob/d7536f62c748b461ef7556af9eba0fb571060c0c/src/components/calcite-filter/calcite-filter.tsx#L30

driskull:
@pr3tori4n ok


fixme: should we update the doc description? It no longer appears there right?
```ts
/**
   * A text label that will appear next to the input field.
   */
```

nitpick: Could the text filter just be set by default and not have the fallback in render? This gives the user the ability to get rid of it if they really want to.

```tsx
  @Prop() textLabel = Text.filterLabel;
```

pr3tori4n:
> nitpick: Could the text filter just be set by default and not have the fallback in render? This gives the user the ability to get rid of it if they really want to.

On second thought no, there needs to be something there or else this component will fail a11y.

PR #577, opened 2019-11-20
Added styles and reworks for demos.

asangma:
**Related Issue:** #463

## Summary
Styles and reworked nav for demos.

### DIFF
--- a/src/components/calcite-demo-nav/calcite-demo-nav.scss
+++ b/src/components/calcite-demo-nav/calcite-demo-nav.scss
@@ -1,8 +1,12 @@
 :host {
-  position: relative;
+  top: 0;
+  left: 0;
+  right: 0;
+  position: fixed;
   display: block;
-  background-color: transparent;
-  width: auto;
+  background-color: #f3f3f380;
+  box-shadow: var(--calcite-app-shadow-1);
+  z-index: 999;
 }
 .hamburger {
   background-color: transparent;
@@ -29,16 +33,17 @@ ul {
     display: flex;
   }
   display: none;
-  flex-flow: column;
+  flex-flow: row wrap;
   box-shadow: var(--calcite-app-shadow-1);
   top: 100%;
   left: 0;
   z-index: 9999;
   position: absolute;
-  li {
+  > li {
     position: relative;
-    width: 12.5vw;
-    &:first-child {
+    width: 24vw;
+    &:first-child,
+    &:last-child {
       width: 100%;
     }
   }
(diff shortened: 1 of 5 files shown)


src/demos/demos.css
```diff
@@ -1,26 +1,33 @@
...
-    overflow-x: scroll;
-  }
+  color: var(--calcite-app-foreground-subtle);
+  max-width: 1024px;
+  min-width: 480px;
+  width: 70vw;
+  padding: var(--calcite-app-cap-spacing-half) var(--calcite-app-side-spacing-double) var(--calcite-app-cap-spacing-double);
+  margin: 0 auto var(--calcite-app-cap-spacing-double);
+  background-color: var(--calcite-app-background-content);
+  box-shadow: var(--calcite-app-shadow-1);
```
pr3tori4n: Can we darken this just a bit. It's very faint.

src/demos/demos.css
```diff
@@ -1,26 +1,33 @@
...
 
-h1, h2, h3, h4, h5 {
+h1,
+h2,
+h3,
+h4,
+h5 {
+  font-family: 'Courier New', Courier, monospace;
   font-weight: 300;
+  font-size: 14px;
```
pr3tori4n: What do you think about increasing the font size and uppercasing the h1 so that the component name sticks out a bit better?

pr3tori4n: On a second review, all of the headings are kind of getting lost on the page. I like the courier font but I think this font size may be a bit too small for effectively dividing the sections of the demo up.

src/demos/demos.css
asangma: Yo! I nuked the desktop nav on purpose. It mostly just caused a horizontal scroll.

re: darkening 
I actually made the bg color the same and just added a bit of box shadow. If anything, we can prolly just nix the box-shadow and not indicate the center column at all.

src/demos/demos.css
asangma: This was in response to @driskull comment that the headers get mixed in with the components. Their size, color, and font-family are all contrasting to the main content, i.e. the components.

Does that make sense?

asangma: Actually, the h1 was too little. Upped its size and weight.

src/demos/demos.css
asangma: I went ahead and nixed the box-shadow. 
So clean...
kleen

asangma: image

driskull: > Yo! I nuked the desktop nav on purpose. It mostly just caused a horizontal scroll.

Agreed. We outgrew it anyway.

driskull: Can the nav button be positioned as sticky? So its always on the top left even when scrolled down?

src/components/calcite-demo-nav/calcite-demo-nav.scss
```diff
@@ -1,23 +1,19 @@
-$largeScreen: 1025px;
-
 :host {
   position: relative;
```
driskull: position stickY????

src/demos/demos.css
```diff
@@ -1,26 +1,36 @@
 body {
   margin: 0;
-  background-color: #dfdfdf;
+  /* background-color: #dfdfdf; */
```
driskull: remove comment

src/demos/demos.css
asangma: It can, but it causes other issues with accessing nav items that fall below the scroll. I did some work to have the menu items be in a 3-up grid, but it was taking more time to make right than it's probably worth.

asangma: Especially with the advanced demos child menu.

asangma: OK...I take it back. 
Have a look now. @driskull Sticky!

driskull: Looks good.

The full window demos need the body margin and padding removed.

can we put a title in that navbar header "Calcite App Components"?

Also, the information architecture of the nav items is odd. Can we move the children of the advanced demos under each component they belong to and get rid of the advanced demos.

asangma: re: title. I'm cool with that. Will add.

re: margin on full width demos
Sorry...that was an oversight on my part.

re: advanced demos.
I like that idea. Can we consider it a later enhancement?

driskull: yes, later enhancement is fine.

pr3tori4n:
@asangma The typography demo's fonts are being overridden. This probably isn't intended right?

Issue #423, opened 2019-10-21
Bug: Create Test to make sure event passes item HTMLElement correctly.


driskull:
### Summary

Add a test for `calciteListItemChange` on `calcite-pick-list-item` to make sure that the `item` property is correctly defined in the event.

### Actual Behavior

Previously, the event did not pass the item like it was supposed to.

### Expected Behavior

The event has the item defined as a HTMLElement.

### Relevant Info 

https://github.com/Esri/calcite-app-components/commit/80e590081c87f4b5e1bd3830077e0134e4eb998a

pr3tori4n:
@driskull please verify.

PR #578, opened 2019-11-21
Lists: added tests to validate event data

pr3tori4n:
**Related Issue:** #423 

## Summary

Previous tests were only validating the event was called. This validates the event detail object for the two event types is what's expected. This should help avoid bugs being introduced where the event data is modified unintentionally.

### DIFF
--- a/src/components/calcite-pick-list-item/calcite-pick-list-item.e2e.ts
+++ b/src/components/calcite-pick-list-item/calcite-pick-list-item.e2e.ts
@@ -26,4 +26,29 @@ describe("calcite-pick-list-item", () => {
     await item.click();
     expect(await item.getProperty("selected")).toBe(false);
   });
+
+  it("should fire event calciteListItemChange when item is clicked", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(`<calcite-pick-list-item text-label="test" value="example"></calcite-pick-list-item>`);
+    const item = await page.find("calcite-pick-list-item");
+    await page.evaluate(() => {
+      document.addEventListener("calciteListItemChange", (event: CustomEvent) => {
+        (window as any).eventDetail = event.detail;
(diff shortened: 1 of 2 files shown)


src/components/calcite-pick-list-item/calcite-pick-list-item.e2e.ts
```diff
@@ -26,4 +26,29 @@ describe("calcite-pick-list-item", () => {
...
+    await item.click();
+
+    await page.waitForChanges();
+
+    const eventDetail: any = await page.evaluateHandle(() => {
+      return (window as any).eventDetail;
+    });
+    const properties = await eventDetail.getProperties();
+    expect(properties.get("item")).toBeDefined();
+    expect(properties.get("value")._remoteObject.value).toBe("example");
```
jcfranco: Noice! 🎸

src/components/calcite-pick-list-item/calcite-pick-list-item.e2e.ts
```diff
@@ -26,4 +26,29 @@ describe("calcite-pick-list-item", () => {
...
+      document.addEventListener("calciteListItemChange", (event: CustomEvent) => {
+        (window as any).eventDetail = event.detail;
+      });
+    });
+
+    await item.click();
+
+    await page.waitForChanges();
+
+    const eventDetail: any = await page.evaluateHandle(() => {
```
jcfranco: Can you type the detail object? This type could also be used in event handlers:

```tsx
// assuming ItemChangeDetail is defined somewhere
calciteListItemChangeHandler(
  this: CalcitePickList | CalciteValueList, 
  event: CustomEvent<ItemChangeDetail>
): void {
```

Aside: I just noticed that we can also type the EventEmitters, that could be useful too.

Issue #531, opened 2019-11-08
Tooling: Set up SCSS linting + auto-fixes


jcfranco:
## Summary

It would be useful to lint our styles and auto-fix whenever possible.

## Measure of Success

**Pain point:** Sass is not being linted and can lead to inconsistencies.

**Desired outcome:** Linting + auto-fixing on commit.

### Resources 

https://github.com/Esri/calcite-app-components/tree/jcfranco/set-up-style-linting

jcfranco:
Installed.

@driskull Can you help verify? There's also a VSCode plugin: https://github.com/shinnn/vscode-stylelint

driskull:
Works. extension is nice too.

![image](https://user-images.githubusercontent.com/1231455/69461966-9fdcef00-0d2c-11ea-8efe-00b37b632dbf.png)

PR #579, opened 2019-11-21
chore(linting): set up style linting

jcfranco:
**Related Issue:** #531

## Summary

This sets up `stylelint` for Sass linting with rules that catch [possible errors](https://stylelint.io/user-guide/rules#possible-errors) with minor customizations (see `.stylelint.json`). The majority possible error rules are not autofixable; I'll need help to determine which [additional](https://stylelint.io/user-guide/rules#limit-language-features) [ones](https://stylelint.io/user-guide/rules#stylistic-issues) to enable (there are several autofixable ones). 

Let me know if we want to enable [`no-descending-specificity`](https://stylelint.io/user-guide/rules/no-descending-specificity). I disabled it since it was reporting some errors and I wanted to ask here before fixing.

This also restores some linting scripts we had before. These come in handy when there's an update to any linter config and you want to do an overall linting pass. 

Also, linting caught a few errors, which are fixed in da13c42.


src/components/calcite-flow-item/calcite-flow-item.scss
```diff
@@ -32,7 +32,6 @@ calcite-panel {
   min-width: var(--calcite-app-menu-min-width);
   right: var(--calcite-app-menu-offset);
   visibility: visible;
-  display: flex;
```
driskull: awesome. already catching things :)

src/components/calcite-flow/calcite-flow.scss
```diff
@@ -45,6 +45,3 @@ $start-alpha: 0.5;
     }
   }
 }
-
-:host([loading]) {
```
driskull: nice

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -72,7 +72,6 @@ slot[name="header-content"]::slotted(.heading) {
 }
 
 .content-container {
-  background-color: var(--calcite-app-background-content);
```
driskull: yussss

Issue #551, opened 2019-11-13
Bug: Shift click unselection in calcite-pick-list doesn't work


driskull:
### Summary

Bug: Shift click unselection in calcite-pick-list doesn't work

### Actual Behavior

You can shift click to select all items up to the selected item but you can't do the same for unselecting them

### Expected Behavior

If You have all the items selected and shift-click on them, they should all become unselected

### Reproduction Steps

1. Select all items
2. Try to unselect all but the first item by shift clicking on the first item.

pr3tori4n:
@kat10140 please verify.

kat10140:
verified

Issue #492, opened 2019-11-04
Research: Properly typing values for JS Map Union Types


pr3tori4n:
### Background

`shared-list-logic.ts` shares a utility method between PickList and ValueList. The method gets an instance reference via `.bind(this)`. The method attemps to add to the state on the instance by adding items to a Javascript Map via the `.set` method. Both the map and the item being added to the map are union types.
`Map<string, HTMLCalcitePickListItemElement> | Map<string, HTMLCalciteValueListItemElement>`
`HTMLCalcitePickListItemElement | HTMLCalciteValueListItemElement`

The typescript compiler errors saying that unionType `HTMLCalcitePickListItemElement | HTMLCalciteValueListItemElement` is not assignable to type `HTMLCalcitePickListItemElement`.

The current workaround is to redefine the map to instead of being a union between two maps, it's a map that accepts a union type.

`const selectedValues = this.selectedValues as any as Map<string, HTMLCalcitePickListItemElement | HTMLCalciteValueListItemElement>`

#### Main Drawback to workaround
This assuages the compiler, but it is technically allowing an instance to occur where the instance could be of type `PickList` but the `selectedValues` map could be passed `ValueListItems`

#### Resources

[StackOverflow Question](https://stackoverflow.com/questions/58699962/typescript-how-to-share-a-bound-utility-function-between-two-classes-that-add-i)

### Desired Outcome

A more sophisticated solution that still passes compilation, but also avoids the drawback mentioned above. Avoiding the use of any or other means of "tricking" the compiler.

PR #581, opened 2019-11-21
shift deselect for lists 551

pr3tori4n:
**Related Issue:** #551 

## Summary

* Added logic to provide mutli select behavior when deselecting.
* Note: When revisiting this file I noticed a bunch of boilerplate there just to please typescript. I simplified this because it didn't seem to be providing much value besides satisfying the compiler. It also simplifies the logic I needed to add to `selectSiblings`
* re-used `selectSiblings` for deselect and added a flag for adding/removing since most of hte logic is shared for both cases. Didn't rename, but let me know if you have a new suggested function name.


src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -2,21 +2,21 @@ import { CalcitePickList } from "./calcite-pick-list";
 import { CalciteValueList } from "../calcite-value-list/calcite-value-list";
 import { debounce } from "lodash-es";
 
-type pickListItemMap = Map<string, HTMLCalcitePickListItemElement>;
-type valueListItemMap = Map<string, HTMLCalciteValueListItemElement>;
+// type pickListItemMap = Map<string, HTMLCalcitePickListItemElement>;
```
driskull: remove comments

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -79,7 +82,8 @@ export const sharedListMethods = {
       item.icon = this.getIconType();
       item.compact = this.compact;
       if (item.selected) {
-        updateSelectedValuesMap(this.selectedValues, item);
+        this.selectedValues.set(item.value, item as any);
+        // updateSelectedValuesMap(this.selectedValues, item);
```
driskull: remove comment

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -79,7 +82,8 @@ export const sharedListMethods = {
       item.icon = this.getIconType();
       item.compact = this.compact;
       if (item.selected) {
-        updateSelectedValuesMap(this.selectedValues, item);
+        this.selectedValues.set(item.value, item as any);
```
driskull: nitpick: Can we give it a type here so that it's not using any?

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -110,8 +114,13 @@ export const sharedListMethods = {
...
     items.slice(Math.min(start, end), Math.max(start, end)).forEach((currentItem: pickOrValueListItem) => {
-      currentItem.toggleSelected(true);
-      updateSelectedValuesMap(this.selectedValues, currentItem);
+      currentItem.toggleSelected(!deselect);
+      if (!deselect) {
+        this.selectedValues.set(currentItem.value, currentItem as any);
+      } else {
+        this.selectedValues.delete(currentItem as any);
+      }
+      // updateSelectedValuesMap(this.selectedValues, currentItem);
```
driskull: remove comment

src/components/calcite-pick-list/shared-list-logic.ts
```diff
@@ -110,8 +114,13 @@ export const sharedListMethods = {
       return currentItem.value === item.value;
     });
     items.slice(Math.min(start, end), Math.max(start, end)).forEach((currentItem: pickOrValueListItem) => {
-      currentItem.toggleSelected(true);
-      updateSelectedValuesMap(this.selectedValues, currentItem);
+      currentItem.toggleSelected(!deselect);
+      if (!deselect) {
+        this.selectedValues.set(currentItem.value, currentItem as any);
```
driskull: nitpick: Can we give it a type here so that it's not using any?

src/components/calcite-pick-list/shared-list-logic.ts
pr3tori4n: this is the alternative fix that allowed to remove the behemoth boilerplate function. That's the alternative, if you prefer that I can go back to that. IMO it hurts readability and is just a workaround to please the compiler.

driskull: I agree that it sucks. @jcfranco is there an easier typing alternative?

pr3tori4n:
@driskull @jcfranco removed commented code. Lmk if I can move forward with the `as any` or if there's a better way unbeknownst to me.

src/components/calcite-pick-list/shared-list-logic.ts
jcfranco: >  remove the behemoth boilerplate function.

IMO, the helper function isn't a behemoth. Please restore. I'd like us to be as type safe as possible even if it does mean adding some extra code. https://github.com/Esri/calcite-app-components/issues/492 may help simplify some of this.

src/components/calcite-pick-list/shared-list-tests.ts
```diff
@@ -88,6 +88,33 @@ export const tests = {
...
+        await item1.click();
+        await page.keyboard.down("Shift");
+        await item3.click();
+        await page.keyboard.up("Shift");
+
+        const numSelected = await page.evaluate((type) => {
+          const domList: HTMLCalcitePickListElement | HTMLCalciteValueListElement = document.querySelector(
+            `calcite-${type}-list`
+          );
+          return domList.getSelectedItems().then((result) => {
```
jcfranco: Nitpick: can you use `await` to reduce promise nesting?

src/components/calcite-pick-list/shared-list-logic.ts
pr3tori4n: Think I found a better solution. I was able to use the Union Type
`type pickOrValueListItem = HTMLCalcitePickListItemElement | HTMLCalciteValueListItemElement;` once I set `textHeading` to optional in pickList. Once that prop became optional they were 'alike' enough to satisfy the compiler, yay.

pr3tori4n: https://github.com/Esri/calcite-app-components/pull/581/files#diff-e6d4d9d31a2a4213f2b8bd1ebbecd98eR71
https://github.com/Esri/calcite-app-components/pull/581/files#diff-e6d4d9d31a2a4213f2b8bd1ebbecd98eR104

Issue #450, opened 2019-10-23
Enhancement: calcite-flow-item - add second header line


AdelheidF:
### Description

design:
![image](https://user-images.githubusercontent.com/44481292/67417721-48f1c780-f57e-11e9-94cc-90a79df90ed6.png)

### Helpful Details

**Which Component:**
**Example Use Case:**

AdelheidF:
Duplicate of https://github.com/Esri/calcite-app-components/issues/406

driskull:
https://github.com/Esri/calcite-app-components/pull/448

AdelheidF:
@asangma @mitc7862 

Do we still want to show field name and field type in the header of the new map viewer attribute selection pop-up? Currently it shows below the header.
Not sure how it treats long names.

![image](https://user-images.githubusercontent.com/44481292/71752018-6078e680-2e32-11ea-8c65-c49c88eac1b9.png)

mitc7862:
Yeah, That should be in the header. The component will need to accommodate that somehow.

driskull:
Next release it should be able to.

kat10140:
verified

AdelheidF:
@mitc7862 

This is what it looks like with a long field name. I now can't see the field type at all. A hover doesn't show anything.

<img src=https://user-images.githubusercontent.com/44481292/72484715-0ffe7300-37ba-11ea-9771-736ab80e3992.png width=250/>

I guess we need to keep the line with the field name and type inside the body anyway. In that case I'm not clear what's the use of having it in the header then too.

<img src=https://user-images.githubusercontent.com/44481292/72484850-9450f600-37ba-11ea-9594-a90a27bc9547.png width=250/>

mitc7862:
Yikes!

I consulted Alan and we agree that having the field type to the sub header isn't going to work. Can we the following?

![image](https://user-images.githubusercontent.com/12564905/72647321-2416b180-392d-11ea-97ac-5ca27ad3fc1b.png)

PR #583, opened 2019-11-22
feat(flow-item, panel): add summary and related styles

asangma:
**Related Issue:** #450

## Summary
Added the summary stuff to Panel and made it so that FlowItem passes through to Panel.
Looking forward to the review. 😬


src/components/calcite-panel/calcite-panel.scss
```diff
@@ -42,32 +43,52 @@
...
+.header-trailing-content {
+  display: flex;
+}
+
+.header-leading-content + .header-middle-content {
   padding-left: var(--calcite-app-side-spacing-half);
 }
 
-slot[name="header-content"]::slotted(.heading) {
+slot[name="header-content"]::slotted(.heading),
```
driskull: these seem pretty brittle. "A slot that has a class of summary"?

src/components/calcite-panel/calcite-panel.scss
```diff
@@ -42,32 +43,52 @@
...
+slot[name="header-summary"]::slotted(.summary) {
+  display: block;
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
   width: 100%;
 }
 
+slot[name="header-content"]::slotted(.heading) {
```
driskull: these seem pretty brittle. "A slot that has a class of heading"?

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -163,7 +173,10 @@ export class CalcitePanel {
     return canDisplayHeader ? (
       <header class={CSS.header}>
         {headerLeadingContentNode}
-        {headerContentNode}
+        <div class={CSS.headerMiddleContent}>
+          {headerContentNode}
+          {headerSummaryNode}
```
driskull: @asangma It seems odd that there is a Content and a Summary slot. Its not obvious which goes above the other. Maybe we just need another component for `calcite-panel-header` or `calcite-flow-item-header` where a heading/summary slots could exist?  Then a user can choose to put that inside or not. Otherwise, we're requiring a heading/summary only in the middle. If that is an OK requirement, maybe we just rename the slot to heading?

src/components/calcite-panel/calcite-panel.scss
driskull: So if a user doesn't supply the heading class this won't be styled correctly. Or if the class is renamed in the flow-item, this would break.

driskull:
@asangma can you resolve the conflicts?

.stylelintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "stylelint-config-recommended-scss",
+  "rules": {
+    "no-descending-specificity": null,
```
pr3tori4n: This rule seems helpful. There a pattern we're following that contradicts this?

pr3tori4n:
@asangma can you update the PR so that master is merged into the base branch, and the base branch is merged into your branch? I'm seeing several diffs for changes that have already been added to master in a separate PR. Hopefully once the branch is up to date these will go away so we can see only the new edits.

.stylelintrc.json
jcfranco: No, we can enable it. I had originally asked if we wanted to enable this before fixing the issues since it's not autofixable. 🛠

src/components/calcite-flow-item/calcite-flow-item.e2e.ts
```diff
@@ -203,13 +203,8 @@ describe("calcite-flow-item", () => {
 
   it("should be accessible", async () =>
     accessible(`
-<<<<<<< HEAD
       <calcite-flow-item heading="hello world" summary="test" menu-open show-back-button>
         <div slot="menu-actions">
```
driskull: @asangma can you put this back to `<div slot="${SLOTS.menuActions}">`

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -61,6 +61,11 @@ export class CalciteFlowItem {
    */
   @Prop() showBackButton = false;
 
+  /**
+   * Summary text. A description displayed underneath the heading.
+   */
+  @Prop() summary: string;
```
driskull: Can we put this as optional? `summary?:`?

src/components/calcite-panel/resources.ts
```diff
@@ -22,6 +22,7 @@ export const ICONS = {
 
 export const SLOTS = {
   headerContent: "header-content",
+  headerSummary: "header-summary",
```
driskull: @asangma can we remove this? I don't think its being used.

src/demos/calcite-flow-item.html
```diff
@@ -0,0 +1,86 @@
+<!DOCTYPE html>
```
driskull: @asangma these demo files should be removed. Only demos within their component folder should exist.

src/components/calcite-flow-item/calcite-flow-item.e2e.ts
asangma: Oops!

asangma: Updated.

src/components/calcite-flow-item/calcite-flow-item.tsx
asangma: Done.

src/components/calcite-panel/resources.ts
asangma: Done.

src/demos/calcite-flow-item.html
asangma: Done.

asangma:
Thanks @driskull !

Issue #590, opened 2019-11-25
Enhancement: Add property to ShellPanel that allows content to "float"


asangma:
### Description
Add a Prop to ShellPanel that will cause this component's `content` node to behave like a "floating panel" without affecting existing usage.

![image](https://user-images.githubusercontent.com/12503298/69577664-b466f480-0f83-11ea-8881-f102333a7464.png)

### Acceptance Criteria 
* Default behavior is as-is
* Prop to activate this appearance
* No change to current implementation
* include scale property for height

### Relevant Info 
This will benefit from the #262 .

#### Which Component
ShellPanel

### Example of unused space in the left panel.
![image](https://user-images.githubusercontent.com/12503298/69578017-73231480-0f84-11ea-8c2d-188c5f920140.png)

#### Example of panel only using as much space as neede.
Show more of the map when the content node can be shorter.
![image](https://user-images.githubusercontent.com/12503298/69579380-37d61500-0f87-11ea-8fde-a1fee49c41ad.png)

arjavbadjatiya:
@asangma  While this change will partly addresses the feedback we have received about the map area, I don't think it'll be enough. The feedback was typically for situations where multiple panels and popouts are open and reduce the visible area of the map. 
![Current](https://user-images.githubusercontent.com/29237960/69585743-63f89280-0f95-11ea-84fd-b9d0931a10c6.png)

   I spoke to @mitc7862  briefly when we had the map viewer usability issues sync about these ideas: 
- For the case of preserving map area when user is a few steps into a process with multiple panels open, we can utilize our **existing drill-in pattern and existing panels** to maintain the view of the map. 

How it could work:  
![Possible option](https://user-images.githubusercontent.com/29237960/69585745-6b1fa080-0f95-11ea-821f-b3f8fcba1b58.png)


And what we would need to do for it:
![Possible option described](https://user-images.githubusercontent.com/29237960/69585756-707ceb00-0f95-11ea-8dd7-5ab62b3e59bd.png)

- Additionally, our general UX goal should be to **minimize scrolling.** So in situations like the fill/outline panel it should be expanded as much as possible from the get go to house the components and cause minimal scrolling when some of the internal components are expanded (irrespective of whether it is a floating panel or drill-in)

- In cases where our goal is to **preserve visible map area**, we should rely on drill-in patterns so that only a fixed amount of space is taken by workflow panels. This should happen with the rich text editor (if possible), styling popouts etc. 

[...]

PR #591, opened 2019-11-25
feat(shell-panel, action): Add content-detached property to ShellPanel

asangma:
**Related Issue:** #590 

## Summary
* Adds the prop and related scale prop
  - maybe the scale prop needs to include "height"?
* Updates some pointer-events rules
  - could use some advice on this
  - intention is to allow interacting with e.g. a map behind the `calcite-shell-panel` node.
* Moves scoped Sass values into CSS vars.

cc @kevindoshier @AdelheidF @jbartley 

@kat10140 Could you test the pointer-events stuff out in a couple Windows browsers?


src/components/calcite-shell-panel/calcite-shell-panel.scss
```diff
@@ -1,24 +1,45 @@
...
+:host([float-content]) .content {
+  border: 1px solid var(--calcite-app-border);
+  border-radius: var(--calcite-app-border-radius);
+  box-shadow: var(--calcite-app-shadow-2);
+  margin: var(--calcite-app-cap-spacing-half) var(--calcite-app-side-spacing-half);
+  max-height: var(--calcite-app-shell-panel-max-height-medium);
+  height: auto;
+}
+
+:host([float-content][float-content-scale="s"]) .content {
```
driskull: @asangma can you test this in Edge? I'm not sure if the polyfill will handle it.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -24,6 +24,16 @@ export class CalciteShellPanel {
...
 
+  /**
+   * floatContent lets the content area appear like a floating panel and only sets a max-height.
+   */
+  @Prop({ reflect: true }) floatContent = false;
+
+  /**
+   * floatContentScale sets the max-height of the content area when floatContent is true.
+   */
+  @Prop({ reflect: true }) floatContentScale: "s" | "m" | "l" = "m";
```
driskull: Since we'll probably use `"s" | "m" | "l"` quite frequently, maybe we should have a type somewhere for `CalciteScale`?

```ts
type CalciteScale = "s" | "m" | "l";
```

src/components/calcite-shell-panel/calcite-shell-panel.scss
driskull: If it doesn't, might be better to have a `.content--modifier` class.

asangma: Sounds good.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
asangma: I like it!

asangma: Like maybe in *interfaces.ts*?

driskull: yep!

kat10140:
Screen Shot 2019-11-25 at 2 35 11 PM

In Edge, action bar pointer events seem to be working by itself, but in shell panel looks weird.

kat10140:
Pointer-events works in Edge, Firefox, Chrome on Windows 🎉🎉🎉 @asangma

asangma:
YEW @kat10140

src/components/calcite-shell-panel/calcite-shell-panel.tsx
asangma: Added a separate issue for this.

asangma:
@pr3tori4n 
>  Might be a better descriptor than float

I see where you're coming from. The styles and layout set the content in what we would call Level 1 where Level 0 is surface and Level 1 is slightly above.

asangma:
Currently, unable to accomplish a content-based height in the float mode because doing so conflicts with a flow-item's sticky header and footer.

So the scale prop sets the height to one of three sizes:
Small
image

Medium (default)
image

Large
image

src/components/calcite-shell-panel/calcite-shell-panel.scss
asangma: @driskull Done.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
asangma: #593

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -24,6 +26,16 @@ export class CalciteShellPanel {
    */
   @Prop({ reflect: true }) collapsed = false;
 
+  /**
+   * floatContent lets the content area appear like a floating panel and only sets a max-height.
+   */
+  @Prop({ reflect: true }) floatContent = false;
```
jcfranco: Is there a name for this that doesn't sound like an action/verb? On a related note, is there an alternate word we could use for `float` to avoid any possible confusion w/ CSS floats?

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -24,6 +26,16 @@ export class CalciteShellPanel {
...
+
+  /**
+   * floatContentScale sets the max-height of the content area when floatContent is true.
+   */
+  @Prop({ reflect: false }) floatContentScale: "s" | "m" | "l" = "m";
```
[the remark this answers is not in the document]
asangma: ha! @driskull Suggested the same thing! I added an issue: #593

src/components/calcite-shell-panel/calcite-shell-panel.tsx
asangma: @pr3tori4n Had the same concern and suggested `detached`. Maybe `contentDetached`?

asangma:
Hello team!
I believe I got the detached content to have dynamic height and max out based on its scale!
Please have another look at this.

Note, Safari does a little bit of weird stuff when there's a flow or anything with a sticky header/footer, but it's somewhat minor. TBH, I think this enhancement is worth that little bit of Safari funk.

Screen Shot 2019-12-06 at 8 55 53 AM

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -24,6 +26,16 @@ export class CalciteShellPanel {
    */
   @Prop({ reflect: true }) collapsed = false;
 
+  /**
+   * floatContent lets the content area appear like a floating panel and only sets a max-height.
+   */
+  @Prop({ reflect: true }) floatContent = false;
+
+  /**
+   * floatContentScale sets the max-height of the content area when floatContent is true.
```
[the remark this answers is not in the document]
asangma: Renamed to `contentDetached`

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -24,6 +26,16 @@ export class CalciteShellPanel {
    */
   @Prop({ reflect: true }) collapsed = false;
 
+  /**
+   * This property makes the content area appear like a "floating" panel.
+   */
+  @Prop({ reflect: true }) contentDetached = false;
```
driskull: Can we rename to just `detached` and `detachedScale` instead of `contentDetached` and `contentDetachedScale`?

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -52,10 +64,13 @@ export class CalciteShellPanel {
   // --------------------------------------------------------------------------
 
   render() {
-    const { collapsed, layout } = this;
+    const { collapsed, contentDetached, layout } = this;
 
     const contentNode = (
-      <div class={CSS.content} hidden={collapsed}>
+      <div
+        class={classnames(CSS.content, contentDetached ? CSS.contentDetached : null)}
```
driskull: Can you change to:

```ts
class={classnames(CSS.content, {[CSS.contentDetached]: contentDetached})}
```

asangma: Will do!

src/components/calcite-shell-panel/calcite-shell-panel.tsx
asangma: I like it.

asangma: Updated.

asangma:
@kat10140 Could use some help testing this.

src/components/calcite-shell-panel/calcite-shell-panel.tsx
```diff
@@ -24,6 +26,16 @@ export class CalciteShellPanel {
...
+   * This property makes the content area appear like a "floating" panel.
+   */
+  @Prop({ reflect: true }) detached = false;
+
+  /**
+   * This sets limits the height of the content area. It only applies when detached is true.
+   */
+  @Prop({ reflect: false }) detachedScale: "s" | "m" | "l" = "m";
+
   @Watch("collapsed")
```
jcfranco: Nitpick: This watcher should be moved right bellow its associated prop, no?

pr3tori4n: yea, ideally.

asangma:
@pr3tori4n Adding you as an assignee. Would you mind creating a snapshot of this for @kevindoshier? Thanks!

asangma:
Notes from build using this branch:

- [ ] flow needs width set to 100%
- [ ] flow-item needs width set to 100%
- [ ] added flow-item is not getting animation

src/assets/styles/_base.scss
```diff
@@ -18,6 +18,7 @@
   align-items: center;
   justify-content: space-between;
   flex-basis: var(--calcite-app-header-min-height);
+  min-height: var(--calcite-app-header-min-height);
```
pr3tori4n: is this and flex-basis both needed? Seems like one or the other would be redundant. Or are we using the same CSS variable for two different purposes? If it's the latter, can we rename the variable to prevent confusion?

src/components/calcite-flow-item/calcite-flow-item.scss
```diff
@@ -2,11 +2,11 @@
   background-color: var(--calcite-app-background-content);
   display: flex;
   height: 100%;
+  width: 100%;
```
pr3tori4n: display flex should act like a block container but apply flexbox model to its children. It shouldn't need width: 100% to stretch to fill its container if its behaving like a block.

src/components/calcite-flow/calcite-flow.scss
```diff
@@ -1,20 +1,22 @@
 $start-alpha: 0.5;
 
 :host {
+  align-items: stretch;
   display: flex;
-  height: 100%;
+  width: 100%;
```
pr3tori4n: same as above.

src/components/calcite-flow/calcite-flow.scss
```diff
@@ -1,20 +1,22 @@
...
   overflow: hidden;
   position: relative;
 
   .frame {
+    align-items: stretch;
     width: 100%;
```
pr3tori4n: same as above

src/components/calcite-shell-panel/calcite-shell-panel.scss
```diff
@@ -1,26 +1,49 @@
...
+  pointer-events: auto;
+}
+
+.content--detached {
+  border: 1px solid var(--calcite-app-border);
+  border-radius: var(--calcite-app-border-radius);
+  box-shadow: var(--calcite-app-shadow-0);
+  margin: var(--calcite-app-cap-spacing-half) var(--calcite-app-side-spacing-half) auto;
+  max-height: var(--calcite-app-shell-panel-max-height-medium);
   overflow: auto;
```
pr3tori4n: wouldn't we still want overflow: auto on the .content when its not detached?

src/components/calcite-shell-panel/calcite-shell-panel.scss
```diff
@@ -1,26 +1,49 @@
...
 }
 
+:host([detached-scale="s"]) .content--detached {
+  max-height: var(--calcite-app-shell-panel-max-height-small);
+}
+
+:host([detached-scale="l"]) .content--detached {
```
pr3tori4n: nitpick: This could be adjusted so there is a modifier on the content element to avoid the descendent selector.

src/components/calcite-flow-item/calcite-flow-item.scss
pr3tori4n: I opened a PR to address some of this stuff. This could follow up merging of this PR if desired. If so lmk and I'll adjust to approve. https://github.com/Esri/calcite-app-components/pull/669

Issue #587, opened 2019-11-22
Tooling: commonTests utils cause warning messages


jcfranco:
## Summary 

When running the E2E test suite, there are multiple warning messages showing up:

```
(node:7609) UnhandledPromiseRejectionWarning: Error: Failed to create target for page
    at assert (/Users/juan6600/dev/projects/calcite-app-components/node_modules/puppeteer/lib/helper.js:279:11)
    at Browser._createPageInContext (/Users/juan6600/dev/projects/calcite-app-components/node_modules/puppeteer/lib/Browser.js:176:5)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at JestEnvironment.newPuppeteerPage (/Users/juan6600/dev/projects/calcite-app-components/node_modules/@stencil/core/dist/testing/index.js:30917:26)
    at Object.newE2EPage (/Users/juan6600/dev/projects/calcite-app-components/node_modules/@stencil/core/dist/testing/index.js:39609:18)
    at simplePageSetup (/Users/juan6600/dev/projects/calcite-app-components/src/tests/commonTests.ts:29:16)
    at Object.accessible (/Users/juan6600/dev/projects/calcite-app-components/src/tests/commonTests.ts:36:16)
(node:7609) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:7609) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Culprit: The `accessible` helper wasn't properly set up in some tests. 🔍:shipit:

## Measure of Success 

No more warnings are shown when running the test suite.

jcfranco:
There were a lot of warnings showing up in the console where the `accessible` helper wasn't set up properly (tests needed to wait for the result before proceeding).

I've restored the helper in this branch (https://github.com/Esri/calcite-app-components/tree/jcfranco/587-wire-up-missing-accessible-tests), and there are two failing tests:

```
 FAIL  src/components/calcite-flow/calcite-flow.e2e.ts (10.401s)
  ● calcite-flow › should be accessible

    expect(received).toHaveNoViolations(expected)

    Expected the HTML found at $('calcite-flow-item:nth-child(3),h2') to have no violations:

[...]

PR #596, opened 2019-11-27
test(a11y): wire up missing accessible tests.

jcfranco:
**Related Issue:** #587 

## Summary

- [x] Restores `accessible` helper
- [x] Fixes failing tests

### DIFF
--- a/src/components/calcite-action/calcite-action.e2e.ts
+++ b/src/components/calcite-action/calcite-action.e2e.ts
@@ -100,8 +100,8 @@ describe("calcite-action", () => {
   });
 
   it("should be accessible", async () => {
-    accessible(`<calcite-action text="hello world" label="hi"></calcite-action>`);
+    await accessible(`<calcite-action text="hello world" label="hi"></calcite-action>`);
 
-    accessible(`<calcite-action text="hello world" label="hi" disabled text-display="visible"></calcite-action>`);
+    await accessible(`<calcite-action text="hello world" label="hi" disabled text-display="visible"></calcite-action>`);
   });
 });
--- a/src/components/calcite-action/calcite-action.tsx
+++ b/src/components/calcite-action/calcite-action.tsx
@@ -137,8 +137,8 @@ export class CalciteAction {
           title={labelFallback}
           aria-label={labelFallback}
           disabled={disabled}
-          aria-disabled={disabled}
-          aria-busy={loading}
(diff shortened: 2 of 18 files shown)


driskull:
@jcfranco fixed failing tests

src/components/calcite-action/calcite-action.e2e.ts
```diff
@@ -100,8 +100,8 @@ describe("calcite-action", () => {
   });
 
   it("should be accessible", async () => {
-    accessible(`<calcite-action text="hello world" label="hi"></calcite-action>`);
+    await accessible(`<calcite-action text="hello world" label="hi"></calcite-action>`);
```
pr3tori4n: would it be clearer for all the tests to use the `async await` pattern rather than the `(param1, param2, …, paramN) => expression` pattern?

src/components/calcite-action/calcite-action.tsx
```diff
@@ -103,24 +116,19 @@ export class CalciteAction {
...
         {calciteLoaderNode}
       </div>
     );
+  }
 
-    const calculatedTextDisplay = textEnabled ? "visible" : textDisplay;
-
-    const textContainerNode =
-      calculatedTextDisplay !== "hidden" ? (
-        <div key="text-container" class={CSS.textContainer}>
```
pr3tori4n: is the key necessary here?

src/components/calcite-action/calcite-action.tsx
```diff
@@ -84,12 +87,22 @@ export class CalciteAction {
...
   //
   // --------------------------------------------------------------------------
 
-  render() {
-    const { compact, disabled, loading, el, textEnabled, textDisplay, label, text } = this;
+  renderTextContainer(textDisplay: TextDisplay): VNode {
+    const { text } = this;
+
+    return textDisplay !== "hidden" ? (
+      <div key="text-container" class={CSS.textContainer}>
```
pr3tori4n: is the key necessary here?

driskull: Yus & not part of PR

src/components/calcite-action/calcite-action.e2e.ts
jcfranco: Maybe. I don't want to enforce one vs the other though. I like to keep it concise and expressive whenever possible, especially since IDEs make it easy.

Issue #601, opened 2019-12-03
Enhancement: Start using REMs instead of pixels


asangma:
### Description
Update typography to use rems.

### Acceptance Criteria 
REMs

### Relevant Info 
Modernity.

#### Which Component
ALL DEM

#### Example Use Case
THE INTERNET.

![image](https://user-images.githubusercontent.com/12503298/70098105-f6291800-15df-11ea-8d11-a3db8df6330d.png)

asangma:
@kat10140 VERFEE!

kat10140:
is calcite-app-cap-spacing typography? if so we still have px definitions.
also looks like in _type.scss we have `--calcite-app-base-font-size: 16px;`

kat10140:
@asangma

asangma:
`spacing` would fall outside of typography.

We are explicitly setting `base-font-size` to 16px based on that being what most browsers already set it to by default. So this would keep things consistent if a browser wants to get cray.

kat10140:
okee! verfied

PR #602, opened 2019-12-03
refactor: update typography and size to use rems

asangma:
**Related Issue:** #601

## Summary
Pixels no more. Woot rems.


src/assets/styles/_type.scss
```diff
@@ -1,18 +1,16 @@
 // Type
-$line-height: 1.3rem !default;
-$base-font-size: 14px !default;
-
 :root {
-  --calcite-app-line-height: #{$line-height};
+  --calcite-app-line-height: 1.125rem;
 
-  --calcite-app-base-font-size: #{$base-font-size};
+  --calcite-app-base-font-size: 16px;
```
pr3tori4n: I'd recommend setting the root font size to 10px, that way the rem multiplication math is really simple.

driskull: Metric system font sizing :P

asangma: That'd be conceptually clean.
Browsers often default to 16px as their base font, so we can ride on that.
Also, this is based on a loooooooong discussion around typography with the calcite-components and Calcite peeps.

asangma:
Sorry @driskull @jcfranco @pr3tori4n 
Had some changes come through during discussions with Calcite peeps.
Please have another look.

src/assets/styles/_type.scss
```diff
@@ -22,3 +20,20 @@ $base-font-size: 14px !default;
   font-size: var(--calcite-app-base-font-size);
   line-height: var(--calcite-app-line-height);
 }
+
+h1.heading {
+  font-size: var(--calcite-app-font-size-3);
+  font-weight: var(--calcite-app-font-weight);
```
driskull: question: Do we need different font weights?

src/assets/styles/_type.scss
asangma: As a note, the typography scaling for the majority of our type will be `0.125rem`.

src/assets/styles/_type.scss
asangma: Meaning for the headings? Or in general?

driskull: In general.

asangma: Yeah. We tried to make the design work with only the regular (400) font weight, but it wasn't enough contrast.

As a note, I am updating so that all headings get the demi weight, and am nixing the redundant `--calcite-app-font-weight-heading` var.

src/assets/styles/_type.scss
asangma: 👍

PR #603, opened 2019-12-04
docs: update 12 3 19

pr3tori4n:
## Summary

* I generated new docs with `npm run docs`
* I updated `stencil.docs.config.ts` to import the baseConfig and build on top of it to prevent the two from getting out of sync again.
* I added `*.config.ts*` to `tsconfig.json` so commits to the config files won't fail in the lint pre-commit.
* Moved `TextDisplay` type to `resources.ts` to fix build errors.


src/components/calcite-action/resources.ts
```diff
@@ -8,3 +8,5 @@ export const CSS = {
   slotContainerHidden: "slot-container--hidden",
   textContainer: "text-container"
 };
+
+export type TextDisplay = "hidden" | "visible" | "interactive";
```
jcfranco: For consistency, could you move this to `interfaces.ts`? We've been using it for shared interfaces and types.

tsconfig.json
```diff
@@ -13,5 +13,5 @@
     "jsx": "react",
     "jsxFactory": "h"
   },
-  "include": ["src", "node_modules/@esri/calcite-components/dist/types/components.d.ts"]
+  "include": ["src", "node_modules/@esri/calcite-components/dist/types/components.d.ts", "*.config.ts"]
```
jcfranco: Nitpick: Maybe keep project-specific files grouped together, then dependencies (`node_modules`).

Issue #586, opened 2019-11-22
Enhancement: Flow should remove the current panel automatically when back is clicked


driskull:
### Description

Enhancement: Flow should remove the current panel automatically when back is clicked

From: https://github.com/Esri/calcite-app-components/wiki/2019-Weekly-Meeting#2019-11-15

### Acceptance Criteria 

Add property `beforeBack`.

```tsx
@Prop() beforeBack: () => Promise<void> = () =>
```

#### Which Component
CalciteFlow

driskull:
This will be a breaking change, so I will remove all deprecations as well.

jcfranco:
Verified locally.

PR #608, opened 2019-12-06
fix!: Flow - Remove current panel automatically when back is clicked #586

driskull:
fix: Flow - Remove current panel automatically when back is clicked 

**Related Issue:** #586

## Summary

- **BREAKING CHANGES**!
- fix: Flow - Remove current panel automatically when back is clicked #586
- chore: Remove deprecated events, methods & properties.
- fix: Remove auto text capability in calcite-action. No longer need this as the FAB is happening in calcite-components.


driskull:
@jcfranco @pr3tori4n This is the breaking changes PR that changes the way the flow `back()` works. 

It also removes all the other deprecated events, properties, methods.

@asangma I also removed the `textDisplay` for auto showing text since that kind of logic is going to happen in the `calcite-components` FAB.

driskull:
@asangma When would we want to install this? We can do it after the next MapViewer release. When is that?

src/components/calcite-action/calcite-action.scss
```diff
@@ -140,9 +120,7 @@
   background-color: var(--calcite-app-background-active);
 }
 
-:host([indicator]) .button--text-visible-visible,
-:host([indicator]) .button--text-interactive:focus,
-:host([indicator]) .button--text-interactive:hover {
+:host([indicator]) .button--text-visible-visible {
```
jcfranco: Nitpick: unrelated to this PR, but is there another name we could use for this modifier to avoid the duplicate `visible`?

src/components/calcite-flow-item/calcite-flow-item.e2e.ts
```diff
@@ -180,15 +180,20 @@ describe("calcite-flow-item", () => {
...
+
+  it("Back method should remove element", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent("<calcite-flow-item></calcite-flow-item>");
+
+    let element = await page.find("calcite-flow-item");
 
-    const eventSpy = await page.spyOnEvent("calciteFlowItemBackClick", "window");
+    await element.callMethod("back");
```
jcfranco: API-wise, would it be useful for `back` to return the removed element?

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -29,6 +29,11 @@ export class CalciteFlowItem {
   //
   // --------------------------------------------------------------------------
 
+  /**
+   * Optionally perform a function to run before going back.
+   */
+  @Prop() beforeBack: () => Promise<void> = () => Promise.resolve();
```
jcfranco: Could you add a test for this?

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -76,23 +81,27 @@ export class CalciteFlowItem {
...
+  //  Public Methods
   //
   // --------------------------------------------------------------------------
 
-  @Element() el: HTMLCalciteFlowItemElement;
+  /**
+   * Removes the component.
+   */
+  @Method()
+  async back(): Promise<void> {
```
jcfranco: From an API-perspective, I would expect the back-handling logic to be on the parent and not the individual items. What's the main reasoning behind the move? Maybe I'm missing something.

driskull: @jcfranco

1.  I was thinking that each panel may have its own logic for when it is able to go back or not. Like a panel may have saved changes and another not need to worry about it. This way, a user can supply a beforeBack on each flow-item.
2. If we have `beforeBack` on the `flow-item` and `back` on the `flow`, there is a disconnect there. If you just have a `flow-item-panel`, the back button does nothing.
3. The click event on the `flow-item` calls the `back` method. If it were on the `flow`, the `flow-item` would have to get its parent and call back? Kind of related to #2 above.

Thoughts?

src/components/calcite-action/calcite-action.scss
driskull: I think it should actually be `button--text-visible` so likely a bug? @asangma

src/components/calcite-flow-item/calcite-flow-item.e2e.ts
driskull: yup

src/components/calcite-flow-item/calcite-flow-item.tsx
driskull: Will move `back()` back to the `flow` 👍

src/components/calcite-action/calcite-action.scss
driskull: Updated

src/components/calcite-flow-item/calcite-flow-item.tsx
driskull: Added tests

src/components/calcite-action/calcite-action.scss
```diff
@@ -66,26 +66,6 @@
   width: 100%;
 }
 
-.button--text-interactive {
```
asangma: Getting rid of this does clean things up, but we are losing the animation. Is there a way to keep the animation?

src/components/calcite-action/calcite-action.tsx
```diff
@@ -123,15 +118,13 @@ export class CalciteAction {
...
-    const { compact, disabled, loading, el, textEnabled, textDisplay, label, text } = this;
+    const { compact, disabled, loading, el, textEnabled, label, text } = this;
 
-    const calculatedTextDisplay = textEnabled ? "visible" : textDisplay;
     const labelFallback = label || text;
     const rtl = getElementDir(el) === "rtl";
 
     const buttonClasses = {
-      [CSS.buttonTextVisible]: calculatedTextDisplay === "visible",
-      [CSS.buttonTextInteractive]: calculatedTextDisplay === "interactive",
```
asangma: I'm thinking that textDisplay is more the appropriate property here since all actions should have text associated with them, yeah? And the property would be related to if and how the text is displayed.

driskull: @asangma I was thinking we would go back to just textEnabled since we aren't using the animation in any use case. What do you think? I can put the animation back.

src/components/calcite-action/calcite-action.scss
driskull: Yes, ill keep it.

asangma: Cool.

src/components/calcite-action/calcite-action.tsx
asangma: Aren't we using the animation when the actionbar expands?

driskull: yes but not on hover/focus. Thats what Im getting rid of.

src/components/calcite-flow-item/calcite-flow-item.tsx
driskull: Test added & fixed.

driskull:
@asangma @jcfranco can you please review again? I'd like to merge this end of the week after the release.

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -29,6 +29,11 @@ export class CalciteFlowItem {
   //
   // --------------------------------------------------------------------------
 
+  /**
+   * Optionally perform a function to run before going back.
```
jcfranco: Small suggestion: `When provided, this method will be called before it is removed from the parent flow.` or something like that. cc @kat10140

src/demos/advanced/flow-adding-items.html
```diff
@@ -103,14 +99,30 @@ <h2>Adding new Flow-items to Flow</h2>
...
 
-      flowNode.addEventListener("calciteFlowItemBackClick", function() {
-        flowNode.back();
-      });
-
       var addFlowButtonNode = document.getElementById("add-flow-item");
 
       addFlowButtonNode.addEventListener("click", function() {
         var newNode = document.createElement("calcite-flow-item");
+        /*
```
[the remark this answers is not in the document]
driskull: Ill un comment a portion of it. 👍

src/components/calcite-flow/calcite-flow.e2e.ts
```diff
@@ -25,13 +25,34 @@ describe("calcite-flow", () => {
...
+
+    await page.setContent("<calcite-flow><calcite-flow-item></calcite-flow-item></calcite-flow>");
+
+    await page.$eval("calcite-flow-item", (elm: HTMLCalciteFlowItemElement) => {
+      elm.beforeBack = this.beforeBack;
+    });
+
+    const flow = await page.find("calcite-flow");
+
+    await flow.callMethod("back");
```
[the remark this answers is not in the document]
driskull: Having trouble getting the item returned. Is it ok with just checking something is defined? The tests don't seem to work with the E2E element returning.

asangma:
@driskull 👀

src/components/calcite-flow/calcite-flow.tsx
```diff
@@ -26,6 +26,34 @@ export class CalciteFlow {
...
+  async back(): Promise<HTMLCalciteFlowItemElement> {
+    const lastItem = this.el.querySelector(
+      "calcite-flow-item:last-child"
+    ) as HTMLCalciteFlowItemElement;
+
+    if (!lastItem) {
+      return;
+    }
+
+    const beforeBack = lastItem.beforeBack ? lastItem.beforeBack : () => Promise.resolve();
```
pr3tori4n: Since the flow-item has a default already, is this fallback necessary?

src/components/calcite-flow/calcite-flow.tsx
```diff
@@ -26,6 +26,34 @@ export class CalciteFlow {
...
+
+    await beforeBack.call(lastItem);
+
+    lastItem.remove();
+
+    return lastItem;
```
pr3tori4n: We may want to return the promise here. That allows anyone calling back to be able to fall into a `.catch()` for error handling should the custom promise they passed error.
```
return lastItem.beforeBack().then(() => {
  lastItem.remove();
});
```

src/components/calcite-flow/calcite-flow.tsx
```diff
@@ -26,6 +26,34 @@ export class CalciteFlow {
...
+   * Removes the currently active `calcite-flow-item`.
+   */
+  @Method()
+  async back(): Promise<HTMLCalciteFlowItemElement> {
+    const lastItem = this.el.querySelector(
+      "calcite-flow-item:last-child"
+    ) as HTMLCalciteFlowItemElement;
+
+    if (!lastItem) {
+      return;
```
pr3tori4n: probably should return `Promise.resolve();` or `Promise.resolve(null);` since that's the expected return type.

src/components/calcite-flow/calcite-flow.tsx
driskull: This is an async function so a promise is always returned.

src/components/calcite-flow/calcite-flow.tsx
driskull: This is an async function so a promise is resolved. Are you suggesting to return null instead of undefined?

src/components/calcite-flow/calcite-flow.e2e.ts
```diff
@@ -25,13 +25,35 @@ describe("calcite-flow", () => {
...
+
+    await page.setContent("<calcite-flow><calcite-flow-item></calcite-flow-item></calcite-flow>");
+
+    await page.$eval("calcite-flow-item", (elm: HTMLCalciteFlowItemElement) => {
+      elm.beforeBack = this.beforeBack;
+    });
+
+    const flow = await page.find("calcite-flow");
+
+    const backValue = await flow.callMethod("back");
```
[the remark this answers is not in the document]
driskull: I just want to test the method on the flow here in the flow e2e test. The event is tested in the item test.

src/components/calcite-flow/calcite-flow.tsx
driskull: Its a nice to have since someone could set the property to null or something.

src/components/calcite-flow/calcite-flow.tsx
driskull: @pr3tori4n ^

src/components/calcite-flow/calcite-flow.tsx
pr3tori4n: Yes, but we aren't returning the beforeBack promise. If that Promise were to error, it would get swallowed by this code. I'm suggesting preserving the Promise chain.

```
const flow = document.querySelector("calcite-flow");
flow.back().catch(error => {
  // beforeBack must have errored, handle gracefully
}).then(() => {
  // resume operations
});

See the fiddle how in one case the catch works and in the other an exception is thrown.
https://jsfiddle.net/pr3tori4n/834auzyd/

File: src/components/calcite-flow/calcite-flow.tsx (hunk 12 above)
Comment (pr3tori4n) [reply src/components/calcite-flow/calcite-flow.tsx]: > This is an async function so a promise is resolved. 

Wasn't sure about this, so yea a new promise wrapper is redundant and not needed.

File: src/components/calcite-action/calcite-action.scss [hunk 14]
```diff
@@ -64,25 +70,10 @@
 
 .button--text-visible {
```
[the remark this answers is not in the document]
driskull: Fixed

src/components/calcite-flow/calcite-flow.tsx
driskull: Fixed. thanks!

PR #610, opened 2019-12-06
fix: demo pages in IE11

driskull:
**Related Issue:** None

fix: demo pages in IE11

- Only loads demopagereloader script if port 3333-3339
- Only loads toggles script if not IE (We can load this if its rewritten with IE support)
- Cleanup

### DIFF
--- a/src/demos/head.js
+++ b/src/demos/head.js
@@ -1,32 +1,57 @@
 (function() {
-  const root = window.location.pathname.split("demos").shift();
+  const CSS = ["demos/demos.css", "build/calcite-app.css", "vendor/@esri/calcite-components/calcite.css"];
+
+  const DEV_SCRIPTS = [
+    {
+      src: "demos/toggles.js"
+    },
+    {
+      src: "demos/demoPageReloader.js"
+    }
+  ];
+
+  const CALCITE_SCRIPTS = [
+    {
+      src: "build/calcite-app.esm.js",
+      type: "module"
+    },
+    {
+      src: "build/calcite-app.js",
+      noModule: true
+    },
+    {
+      src: "vendor/@esri/calcite-components/calcite.esm.js",
+      type: "module"
+    },
+    {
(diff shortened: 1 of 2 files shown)


src/demos/head.js
```diff
@@ -1,32 +1,58 @@
...
+    }
+  ];
+
+  if (location.port >= 3333 && location.port < 3340) {
+    SCRIPTS.push({
+      src: "demos/demoPageReloader.js"
+    });
+  }
+
+  const IS_IE = /*@cc_on!@*/ false || !!document.documentMode; // Internet Explorer 6-11
```
jcfranco: Can the user agent be used instead or why is `@cc_on!@` needed?

Nitpick: can the var be `IS_IE11` or why do we need to focus on older versions?

src/demos/head.js
```diff
@@ -1,32 +1,58 @@
...
+      src: "vendor/@esri/calcite-components/calcite.esm.js",
+      type: "module"
+    },
+    {
+      src: "vendor/@esri/calcite-components/calcite.js",
+      noModule: true
+    }
+  ];
+
+  if (location.port >= 3333 && location.port < 3340) {
```
jcfranco: Where is this port range coming from? Is it guaranteed?

driskull: This is what I've seen when starting the dev server. Do you have any ideas?

jcfranco: We _could_ assume if there's any port at all that the dev server is running and enable page reloading.

src/demos/head.js
```diff
@@ -1,32 +1,58 @@
...
+      src: "vendor/@esri/calcite-components/calcite.esm.js",
+      type: "module"
+    },
+    {
+      src: "vendor/@esri/calcite-components/calcite.js",
+      noModule: true
+    }
+  ];
+
+  if (location.port) {
```
jcfranco: Could you add a comment about this for future us?

Issue #573, opened 2019-11-19
Enhancement: Interim focus style


asangma:
### Description
Until #111 is full addressed, add a consistent adjustment to browser standard `outline`. Seems like most issues with default focus ring styles are the ring getting clipped.

Propose a mixin that insets the focus outline.

```
@mixin focusRingInset() {
&:focus {
    outline-offset: var(--calcite-app-outline-inset);
  }
}
```

`--calcite-app-outline-inset` is already in.

### Acceptance Criteria 
A good interim solution for focus ring issues.

### Relevant Info 
[W3 focus doc](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html)

#### Which Component
All.

#### Example Use Case
Keyboard navigation.

asangma:
@kat10140 VERVEY!

kat10140:
Verified!

kat10140:
![value list focus](https://user-images.githubusercontent.com/48069902/70483818-a5616580-1a9e-11ea-8aff-3c352c08a840.gif)

@asangma I missed this on compact mode on the value list- notice the smaller focus on compact items. Also cuts off a bit on the main meat of the dragon drop items

asangma:
VERF AGERN

kat10140:
lewks gewd

PR #615, opened 2019-12-09
feat: add mixin for interim focus style

asangma:
**Related Issue:** #573

## Summary
* Add a mixin for insetting the focus ring
* Include mixin in action, block, and pick-list-item


jcfranco:
focus

PR #621, opened 2019-12-10
fix(value-list-item): focus outline for handle

asangma:
**Related Issue:** #573 

## Summary
Focus ring for value-list-item handle.


kat10140:
yeyer!

PR #627, opened 2019-12-11
chore: replace "calcite-example" component with doc

driskull:
**Related Issue:** none

## Summary

This PR moves the actual "calcite-example" component to a readme boilerplate markdown file. The required snippets are all in a `.md` file and has the added benefit that we can add instructions if necessary.

- **Advantages:** It won't actually create a component `calcite-example` that is defined as an HTML Element.


jcfranco:
TBH, I'd rather do away with the example component and have the boilerplate doc be more high level and link to existing code where it needs to. Moving the boilerplate to a text file opens it up to inconsistencies, typos and possible bugs that won't be easily caught anymore.

driskull:
@jcfranco I'm on board with that. This was a simple step to get the example out because it was mixed in with real source code. We may even want the boilerplate to be a resource in a separate repo that is shared between `calcite-components`.

driskull:
Added `Disadvantages: Boilerplate example isn't debuggable or lintable until you copy it into a file.`

@jcfranco Should we have an issue for further changes?

jcfranco:
Awesome, thanks! Yeah, a follow-up issue would be good.

driskull:
https://github.com/Esri/calcite-app-components/issues/640

Issue #628, opened 2019-12-11
Docs: action-bar snippets


kat10140:
### Description 
snippets and usage folder for action-bar component 
sub issue for #619 

### Which Component
action-bar

Issue #619, opened 2019-12-10
Docs: Add usage .md files for each component


driskull:
### Which Component

- [x] calcite-action-bar * #628 
- [x] calcite-action (@driskull) #620
- [x] ~~calcite-action-group~~ // not needed
- [x] calcite-action-pad * #635 
- [x] calcite-block* #637 
- [ ] calcite-block-section
- [x] ~~calcite-demo-nav~~ // not needed
- [x] ~~calcite-example~~ // not needed
- [x] ~~calcite-filter~~ //not needed for now
- [x] calcite-flow* #695 
- [ ] calcite-flow-item
- [x] calcite-panel* #697 
- [x] calcite-pick-list* #699 
- [ ] calcite-pick-list-group
- [ ] calcite-pick-list-item
- [x] calcite-shell* #701 
- [x] calcite-shell-panel* #702 
- [x] calcite-tip* #705 
- [x] ~~calcite-tip-group~~ // not necessary
- [x] calcite-tip-manager* #703 
- [x] calcite-value-list* #704 
- [ ] calcite-value-list-item

@kat10140 if this is too much work, we can divy up the components. Let us know.

We can also prioritize which components need usage docs first.

### Resources 

1. Create a `basic.md` file in a `usage` folder inside each components folder.
2. Add HTML snippets on how the component will be used.
3. Add more advanced `.md` files as necessary for each component.
4. The name of the file will be translated into a header in the documentation for that component, so we should come up with a standard file naming convention. We can have things like `basic`, `advanced` or more specific examples and use cases for each component usage.

Example: https://github.com/Esri/calcite-components/blob/master/src/components/calcite-tooltip/usage/basic.md

Marked as high priority to support users

kat10140:
I put stars by the ones that I think are higher priority. I'll probably just go down the list.  Please feel free to modify the original comment errybody

kat10140:
@driskull i see you marked as installed. Do you think it's fine to just have the main components that are highlighted by storybook to have usage folders and snippets?

driskull:
Whoops my bad.

> Do you think it's fine to just have the main components that are highlighted by storybook to have usage folders and snippets?

Yeah I think so. We could always do the sub ones later if needed. Seems kinda redundant for the `calcite-action-group` and such.

kat10140:
Sounds good!

PR #629, opened 2019-12-11
docs: action bar snippets

kat10140:
**Related Issue:** #619 #628 

Do you think this is enough? Or should we be extra explicit and show RTL directions and try and cover all the bases for each comp?


driskull:
> should we be extra explicit and show RTL directions and try and cover all the bases for each comp?

I don't think we should cover RTL.

I think we should do the basics and cover all the main properties/attributes of the component. Some of them may be pretty basic and not need a snippet.

src/components/calcite-action-bar/usage/basic.md
```diff
@@ -0,0 +1,53 @@
...
+</calcite-action-bar>
+```
+
+#### With grouping
+
+Renders a group of `calcite-actions` contained in a `calcite-action-group`.
+
+```html
+<calcite-action-bar>
+  <calcite-action-group>
```
driskull: I would have two groups here and explain that groups will be divided.

src/components/calcite-action-bar/usage/basic.md
```diff
@@ -0,0 +1,53 @@
...
+    <calcite-action text="Add">
+      <!-- icon -->
+    </calcite-action>
+    <calcite-action text="Save">
+      <!-- icon -->
+    </calcite-action>
+    <calcite-action text="Layers">
```
driskull: Maybe layers in a separate group?

src/components/calcite-action-bar/usage/basic.md
```diff
@@ -0,0 +1,53 @@
...
+</calcite-action-bar>
+```
+
+#### Bottom Actions
+
+Renders `calcite-actions` that stick to the bottom of the action bar.
+
+```html
+<calcite-action-bar>
+  <div slot="bottom actions">
```
driskull: Maybe have an action at the top too? And just one at the bottom.

Maybe also explain that they will be above the expand/collapse?

src/components/calcite-action-bar/usage/basic.md
```diff
@@ -0,0 +1,53 @@
+#### Basic Action Bar (Top Actions)
+
+Renders `calcite-actions` that stick to the top of the bar.
```
driskull: Same as Franco's feedback from previous PR: `calcite-action`'s

driskull: Occurs multiple places.

kat10140:
dun!

kat10140:
You right 🤓

Issue #401, opened 2019-10-18
Enhancement: Use storybook for demos


driskull:
### Description

`calcite-components` is now using storybook for displaying demos. I think we should probably do the same.

https://esri.github.io/calcite-components/

### Acceptance Criteria 

- Use storybook for demos
- Remove demo nav

### Relevant Info 

See PR for calcite-components: https://github.com/Esri/calcite-components/pull/177

jcfranco:
@kat10140 Can we target this for Eberhart or Fern? I can start doing some setup work and I'll pull in more people when it comes to adding individual component stories.

kat10140:
Sure, sounds good @jcfranco 💫

jcfranco:
Installed storybook groundwork. Will create follow-up issue to complete stories.

driskull:
Groundwork looks good @jcfranco!

PR #631, opened 2019-12-11
docs(demos): wire up storybook

jcfranco:
**Related Issue:** #401 

## Summary

This sets up Storybook to generate public-facing demos. 

cc @asangma @kat10140 

### Notes

* TSX is not supported, so stories need to use either string templates (similar to e2e tests) or programmatically set up element (e.g., `document.create("calcite-foo");`). [`createComponentHTML`](https://github.com/Esri/calcite-app-components/blob/d63c882327af5c4cfa682a8e28233e55c4489b98/.storybook/utils.ts#L52-L53) can help out for template-based stories (feedback is much appreciated).
* Includes stories for `calcite-block` and `calcite-block-section` (placeholders for the rest). 
* `docs` folder should be replaced once we have all stories filled out (possibly outside of this PR). 
* The `docs:preview` script can be used to test locally without generating a docs build.
* Supporting components should be handled in parent story (see [block stories](https://github.com/Esri/calcite-app-components/blob/d63c882327af5c4cfa682a8e28233e55c4489b98/src/components/calcite-block/calcite-block.stories.ts) for example).
* Stories use [CSF format](https://storybook.js.org/docs/formats/component-story-format/) (recommended).
* `tsconfig.json` split up into multiple ones (base, default, storybook, tslint).
* Follows `calcite-components` storybook.
* This still allows generating docs the old way, but should be removed once all stories are complete (i.e., undo 6f9441b0c133675cb2cc95ae9360666e0c51ba5b).


.storybook/theme.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export default create({
+  base: "light",
+  brandTitle: "Calcite App Components",
+  brandUrl: "./",
+  brandImage: logo,
+  addonNotesTheme: {
+    code: {
+      whiteSpace: "normal !important"
```
driskull: question: Whats up with these `!important` styles?

.storybook/theme.ts
```diff
@@ -0,0 +1,23 @@
+import { ThemeVars, create } from "@storybook/theming";
+import logo from "./logo.svg";
+
+interface ExtendedThemeVars {
+  addonNotesTheme: any;
+}
+
+export default create({
+  base: "light",
+  brandTitle: "Calcite App Components",
```
driskull: nitpick: Maybe we could import this title from somewhere?

.storybook/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+  optionsKnob as options,
+  radios,
+  select,
+  text
+} from "@storybook/addon-knobs";
+
+export const darkBackground = [
+  {
+    name: "Dark",
+    value: "#202020",
```
driskull: nitpick: maybe we could import this color from calcite colors?

src/components/calcite-block/calcite-block.stories.ts
```diff
@@ -0,0 +1,118 @@
...
+      block: parseReadme(blockReadme),
+      section: parseReadme(sectionReadme)
+    },
+    backgrounds: darkBackground
+  }
+};
+
+const createBlockAttributes: () => Attributes = () => {
+  const group = "block";
+  const dirOptions = ["ltr", "rtl"];
```
driskull: question: Maybe we could have constants we import for `dirOptions` and `themeOptions` since they will be the same throughout stories?

src/components/calcite-block/calcite-block.stories.ts
```diff
@@ -0,0 +1,118 @@
...
+  const themeOptions = ["light", "dark"];
+
+  return [
+    {
```
driskull: question: In the future, it seems like we maybe could auto generate these from docs JSON generated by Stencil. I remember Paul maybe tried that in `calcite-components` and thought it wasn't worth it because its hard to filter some stuff out. Maybe we could have a filter map file to hide the props we don't want to see. Thoughts?

jcfranco: I think we could automate a base set of options, but stories will still need to be set up manually. We can do this in a follow-up issue.

src/components/calcite-block/calcite-block.stories.ts
jcfranco: 💥

.storybook/utils.ts
jcfranco: 💥

.storybook/theme.ts
jcfranco: I got this from `calcite-components`, but after a quick test it doesn't seem like anything is overriding the theme, so I think we can drop the `!important`s.

jcfranco: 💥

.storybook/theme.ts
driskull: Yep! The demo nav can go away after this PR merges anyway. 👍

src/components/calcite-block/calcite-block.stories.ts
driskull: Sounds good

Issue #599, opened 2019-12-03
Bug: Filter. X is showing when there is nothing in input


asangma:
### Summary
X should only show when there is something in the input.

### Actual Behavior
![image](https://user-images.githubusercontent.com/12503298/70082426-3fb53b00-15bf-11ea-8419-3a10c1f55724.png)

### Expected Behavior
![image](https://user-images.githubusercontent.com/12503298/70082457-4e035700-15bf-11ea-81be-072649eb5095.png)

And when there's text...
![image](https://user-images.githubusercontent.com/12503298/70082497-5eb3cd00-15bf-11ea-90b8-0e6613a6439e.png)

### Reproduction Steps

1.

pr3tori4n:
@kat10140 please verify.

kat10140:
verified

Issue #597, opened 2019-12-02
Enhancement: calcite-pick-list - change filter text


AdelheidF:
### Description

Is it possible to change the string in the filter?

Design
<img src=https://user-images.githubusercontent.com/44481292/67066856-1765a000-f129-11e9-8364-d8a9a752acc3.png width=250/>
Component:
<img src=https://user-images.githubusercontent.com/44481292/67066889-38c68c00-f129-11e9-808c-9d217f600c0c.png width=250/>

pr3tori4n:
@kat10140 please verify.

AdelheidF:
Works with 4.11.1.

<img src=https://user-images.githubusercontent.com/44481292/71122722-19ce8f80-2196-11ea-88ed-1c2dff954f3c.png width=250/>

kat10140:
Werkin for me!

PR #634, opened 2019-12-11
feat: filter updates 597 599

pr3tori4n:
**Related Issue:** #597 #599 

## Summary

* Clear button only appears when the filter has a value.
* pickList and valueList now have a `textFilterPlaceholder` prop used to override the default "Filter Results" placeholder string.


src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -42,6 +42,8 @@ export class CalciteFilter {
 
   @Element() el: HTMLCalciteFilterElement;
 
+  @State() empty = true;
```
driskull: nitpick: Instead of `empty`, maybe `active` for when it does have a value populated?

Issue #642, opened 2019-12-12
Bug: Giant panel headers and footers


asangma:
### Summary
![image](https://user-images.githubusercontent.com/12503298/70727215-25144d80-1cb4-11ea-8760-702f8588e880.png)

### Actual Behavior
Gigantor.

### Expected Behavior
![image](https://user-images.githubusercontent.com/12503298/70727253-378e8700-1cb4-11ea-89ec-2e79dc103b5c.png)

### Reproduction Steps

1. full window demo
1. collapse all blocks in seconary panel.

### Relevant Info 
@CodeBarker @kevindoshier

jbartley:
I assume this is related to what is on dev this morning.  

![image](https://user-images.githubusercontent.com/2277787/70731093-006fa400-1cbb-11ea-9f89-24c78ba5647b.png)

If it is we should get it fixed quickly.  

@kevindoshier @CodeBarker

driskull:
Yeah, they are a little big :D

asangma:
@driskull LOL!

@kevindoshier Once we release a patch, please let me know if the issue is solved.

driskull:
Patch released

asangma:
![image](https://user-images.githubusercontent.com/12503298/70733950-23508700-1cc0-11ea-9c22-3ab29a41f86f.png)

asangma:
@kevindoshier @AdelheidF is this still an issue with the updates?

kevindoshier:
No I think its fine now

PR #643, opened 2019-12-12
fix(panel): flex grow, shrink, basis for panel headers and footers

asangma:
**Related Issue:** #642

## Summary
Added better flex grow, shrink, basis for panel header and footer.

@driskull 
Could use a patch for this one.


asangma:
Thanks @driskull !

Issue #564, opened 2019-11-15
Tooling: Tweak build to process non Stencil-related TS files.


jcfranco:
## Summary 

Currently, TS files that are unrelated to components (e.g., demo/workflow support files) are not getting picked up by the Stencil build, preventing us from writing these in TS. 

## Measure of Success 

Having all non-component TS files get built alongside the Stencil build.

![3ggzs7](https://user-images.githubusercontent.com/197440/68961675-c37dc380-0787-11ea-80ef-a02b2ff74f21.jpg)

jcfranco:
Installed.

PR #649, opened 2019-12-13
chore(tooling): tsify all files

jcfranco:
**Related Issue:** #564 

## Summary

# TSifies™ all the things!

- TSifies remaining JS files
- Support files from NPM scripts are transformed and run w/ `ts-node`
  - tested all but `githubReleaseFromChangelog.ts`
- Demo files get their own small build when starting the dev server
  - Emits ES5 (for IE11 support)

### Question:

There are now several `tsconfig.json` files at the root, _should these be moved closer to their context_? e.g., move `tsconfig-storybook.json` to `./storybook/tsconfig.json`?
        
One drawback (at least in IntelliJ) to having them at the same level is that autocompletion in `tsconfig.json` gets disabled if renamed.


src/demos/head.ts
```diff
@@ -0,0 +1,32 @@
+(() => {
```
driskull: Fixme: This is an older version and missing some of the logic added to not load demo page reloader unless port is present and to not load the toggles for IE11.

https://github.com/Esri/calcite-app-components/blob/e99c18f00d98da1364e151b870dbf7b201fbaf8f/src/demos/head.js

jcfranco: Good catch. Will fix.

jcfranco: 💥

Issue #538, opened 2019-11-09
Enhancement: Figure out aria-role for valueList


pr3tori4n:
### Description
`aria-role="checkbox"` is on the Host for the list-items. But it'd be more appropriate on the icon. However value lists don't have an icon. Need to find another solution.
#### Desired New Behavior
none. This is more of a refactor. Functionality shouldn't change. A11y should still pass.
#### Desired Change to Look and Feel
none
### Relevant Info 
its invalid on the label element.
### Helpful Details
**Which Component:** `calcite-value-list-item, calcite-pick-list-item`

kat10140:
verified

PR #655, opened 2019-12-17
style: added role="menu" to value and pick list

pr3tori4n:
**Related Issue:** #538 

## Summary

* This improves their a11y.
* Moved from `listbox` role to `menu` role because listbox has requirements regarding direct children, which is complicated with nested shadow doms + light doms, specifically when the groups were used. The `menu` + `menuitemcheckbox` are not as restrictive, and add similar semantic value.


driskull:
This one is failing

---

expect(received).toHaveNoViolations(expected)
    Expected the HTML found at $('calcite-pick-list') to have no violations:
    <calcite-pick-list role="listbox" aria-disabled="false" aria-busy="false" mode="selection" text-filter-placeholder="Filter results" class="hydrated">
    Received:
    "ARIA input fields have an accessible name (aria-input-field-name)"
    Fix any of the following:
      aria-label attribute does not exist or is empty
      aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
      Element has no title attribute or the title attribute is empty
    You can find more information on this issue here: 
    https://dequeuniversity.com/rules/axe/3.4/aria-input-field-name?application=axeAPI

pr3tori4n:
> Should this be fix rather than style? IMO: it does improve the semantics.

Boils down to this imo: Do you think its worth being mentioned in the changelog?

pr3tori4n:
@driskull just fyi I updated the role value. Fixes the failing a11y tests.

driskull:
👍

Issue #637, opened 2019-12-11
Docs: block snippets


kat10140:
### Description 
snippets and usage folder for the block component
sub issue for #619

Which Component
calcite-block!

kat10140:
they there

PR #657, opened 2019-12-17
docs: calcite block snippets

kat10140:
**Related Issue:** #637 #619


src/components/calcite-block/usage/basic.md
```diff
@@ -0,0 +1,45 @@
...
+
+```html
+<calcite-block heading="Icon't believe it!">
+  <div slot="icon">🤯</div>
+</calcite-block>
+```
+
+#### Header and content (always open)
+
+Renders a header and content that remains open - no the option to collapse the block.
```
jcfranco: > - no the option 

This is missing something, right?

kat10140: yes, thank you! fixed :)

Issue #644, opened 2019-12-12
Docs: [Storybook] Complete component stories


jcfranco:
### Description 

#401 sets up storybook for demos and only one functional story (see resources below) as an example. The rest are placeholders that need to be filled out.

### Which Component

@driskull 

* [x] `calcite-action` 
* [x] `calcite-action-bar`
* [x] `calcite-action-pad`
* [x] `calcite-flow`
* [x] `calcite-panel`
* [x] `calcite-shell`

@pr3tori4n 

* [x] `calcite-pick-list`
* [x] `calcite-tip`
* [x] `calcite-tip-manager`
* [x] `calcite-value-list`

I can help out if needed. I'll just need info on the stories we want. The story-building utility might also need some tweaks to accommodate other use cases. 

Once all stories are complete, we should replace the old way of generating docs (i.e., undo 6f9441b).

### Resources 

https://storybook.js.org/docs/basics/writing-stories/
[`calcite-block.stories.ts`](https://github.com/Esri/calcite-app-components/blob/master/src/components/calcite-block/calcite-block.stories.ts)

driskull:
We should probably do some cross browser testing before swapping gh-pages to this.

jcfranco:
☝️ Did a quick test on all browsers (Chrome, Firefox, Safari, Edge) and everything seems to run fine.

driskull:
Sweet! lets kick off the PR to make this public :)

jcfranco:
☝️ 💥

@driskull @pr3tori4n Excellent job, you guys!

jcfranco:
Verified on https://esri.github.io/calcite-app-components 🎉

PR #662, opened 2019-12-17
docs: [Storybook] Flow & Panel component stories

driskull:
**Related Issue:** #644

## Summary

docs: [Storybook] Flow & Panel component stories

### DIFF
--- a/.storybook/resources.ts
+++ b/.storybook/resources.ts
@@ -41,11 +41,10 @@ export function createSVG(path: string, size: number): string {
 }
 
 export function createAction(
-  { text, label, textEnabled }: { text: string; label: string; textEnabled?: boolean },
+  { text, label, textEnabled, slot }: { text: string; label?: string; textEnabled?: boolean; slot?: string },
   path: string
 ): string {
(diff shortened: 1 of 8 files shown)


src/components/calcite-panel/calcite-panel.scss
```diff
@@ -73,6 +73,7 @@ slot[name="header-content"]::slotted(.heading) {
 }
 
 .header-trailing-content {
+  display: flex;
```
driskull: This is to display the close button and trailing slot together.

.storybook/resources.ts
```diff
@@ -40,6 +40,11 @@ export function createSVG(path: string, size: number): string {
 `;
 }
 
-export function createAction({ text, label }: { text: string; label: string }, path: string): string {
-  return `<calcite-action text="${text}" label="${label}">${createSVG(path, 16)}</calcite-action>`;
+export function createAction(
+  { text, label, textEnabled, slot }: { text: string; label?: string; textEnabled?: boolean; slot?: string },
```
jcfranco: Nitpick: might be worthwhile extracting this into a separate interface if it gets more arguments.

driskull: yeah, it sucks we can't just use JSX :(

Issue #565, opened 2019-11-15
Bug: Calcite-picklist not spacing list correctly - Safari


russroberts:
### Summary
It looks like the picklist is not spacing correctly causing rows to overlap. Only reproducible in Safari
### Actual Behavior
![image](https://user-images.githubusercontent.com/17597538/68967861-2c2c6680-07af-11ea-9f4b-aa38f97be773.png)

Reproduced this by opening up the Add Fields view in NMV with this map
https://devtesting.mapsdevext.arcgis.com/apps/mapviewer/index.html?webmap=775dd1692f72453d89036c67b647de77

MacOS, Safari 13.0.3

jcfranco:
+@kat10140 for triage.

asangma:
Looks like part of this can be solved by setting the `max-height` on **calcite-flow** instead of on **calcite-pick-list**.

There is still some issue with the height of header and footer of **calcite-panel**. We may have a better fix for this in the next release.

AdelheidF:
> Looks like part of this can be solved by setting the max-height on calcite-flow instead of on calcite-pick-list.

No that hides the Done and Cancel buttons of the flow-item for all browsers.

AdelheidF:
No change with "@esri/calcite-app-components@4.10.0" if I set max-width on the flow. Done and Cancel buttons are at the bottom of the page.

![image](https://user-images.githubusercontent.com/44481292/70668190-c0131600-1c27-11ea-8407-ed16c74e1db3.png)

<img src=https://user-images.githubusercontent.com/44481292/70668232-db7e2100-1c27-11ea-9d6f-ce752446663e.png width=200/>

AdelheidF:
Not working with "@esri/calcite-app-components@4.10.1" either.

asangma:
@AdelheidF @russroberts Probably fix for this in [4.11.1 release](https://github.com/Esri/calcite-app-components/releases/tag/v4.11.1).

AdelheidF:
Works now in Chrome.

<img src=https://user-images.githubusercontent.com/44481292/71124216-3c15dc80-2199-11ea-8738-64843998bc9d.png width=250/>

But I noticed a small CSS issue. Once you scroll the grey line shortens.

<img src=https://user-images.githubusercontent.com/44481292/71124318-71222f00-2199-11ea-8979-142ccdd8c1b9.png width=250/>

AdelheidF:
[...]

PR #664, opened 2019-12-18
fix(flow): remove height rule for flow's frame div

asangma:
**Related Issue:** #565

## Summary
Tested in local build of map viewer beta in Safari. Seemed to work.

cc @AdelheidF


asangma:
Would need a patch.

asangma:
Thanks @driskull

PR #670, opened 2019-12-19
refactor: action - make text prop required

pr3tori4n:
**Related Issue:** #586 

## Summary
Remove the label prop and mark text as required. The text prop is now either used to show the text inside the button, or added to attributes when text-enabled is not present to allow for a11y. This should help prevent scenarios from happening where the text and label (both optional props due to one or the other being expected) being omitted and the a11y of the action being poor.

Note: This may be considered a breaking change if folks were using icons + label for a11y. If the text prop was being used as the fallback, then it should work the same as before.


src/components/calcite-action/calcite-action.tsx
```diff
@@ -45,11 +45,6 @@ export class CalciteAction {
    */
   @Prop({ reflect: true }) indicator = false;
 
-  /**
-   * Label of the action, exposed on hover.
-   */
-  @Prop() label: string;
```
driskull: @pr3tori4n maybe we should keep this. It gives us more flexibility to set more of a hover description or some additional information to just the text being displayed. @asangma what do you think?

pr3tori4n:
Yea there's a downside no matter what we do. Let's discuss some options on Friday.

pr3tori4n:
@driskull I've added the label back and made text required. Please review.

src/components/calcite-action/calcite-action.e2e.ts
```diff
@@ -6,19 +6,19 @@ describe("calcite-action", () => {
 
   it("honors hidden attribute", async () => hidden("calcite-action"));
 
-  it("should not have text container", async () => {
+  it("should not have text container when text is not enabled", async () => {
     const page = await newE2EPage();
 
-    await page.setContent(`<calcite-action></calcite-action>`);
+    await page.setContent(`<calcite-action text="hello world"></calcite-action>`);
```
driskull: @pr3tori4n I think this one is failing the test.

pr3tori4n:
@driskull fixed the tests.

driskull:
Looks good now. Will merge after build

PR #672, opened 2019-12-19
chore: remove .only from test files

pr3tori4n:
# Summary

A few tests were merged that had .only attached. Fixing.


kat10140:
I do think you have to make a semantic pr title for it to let you merge

PR #677, opened 2019-12-19
chore: turn on storybook toolbar so docs can be toggled.

driskull:
**Related Issue:** none

chore: turn on storybook toolbar so docs can be toggled.


.storybook/config.ts
```diff
@@ -7,7 +7,7 @@ addParameters({
   backgrounds: [{ name: "Light", value: "#f8f8f8", default: true }],
   options: {
     theme,
-    isToolShown: false,
+    isToolShown: true,
```
jcfranco: Nitpick: we can drop this line since it's the default.

Issue #660, opened 2019-12-17
Docs: Footer slot for flow item and panel


kat10140:
### Description 
As it is-  " @slot footer-actions - A slot for adding `calcite-actions` to the footer. "
This slot should be for adding a calcite-button, not actions. 

### Which Component
flow item and panel

kat10140:
dun closing

PR #684, opened 2019-12-20
docs: fixed footer actions slot doc to use button #660"

kat10140:
**Related Issue:** #660


src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -18,7 +18,7 @@ const SLOTS = {
  * @slot header-content - A slot for adding content in the center of the header.
  * @slot header-leading-content - A slot for adding a `calcite-action` on the leading side of the header.
  * @slot header-trailing-content - A slot for adding a `calcite-action` on the trailing side of the header.
- * @slot footer - A slot for adding `calcite-actions` to the footer.
+ * @slot footer - A slot for adding a `calcite-button` to the footer.
```
driskull: should be `buttons`, they can have multiple

kat10140: 🤙🏼

Issue #691, opened 2019-12-23
Bug: shell-panel and panel flex problem


asangma:
### Summary
![image (1)](https://user-images.githubusercontent.com/12503298/71382611-8b8b4c80-258d-11ea-9e32-31814e6250d5.png)

### Reproduction Steps

1.

asangma:
@kat10140 verfee pleez

kat10140:
👍🏼

PR #692, opened 2019-12-23
fix(panel, shell-panel): flex-flow fix

asangma:
**Related Issue:** #691 

## Summary
- flex-flow rule for panel and shell-panel
- block spacing to take account of margin rendering in flex
- panel styles to ensure spacing stays accurate with blocks


asangma:
Thanks @driskull!

PR #693, opened 2019-12-23
chore: set bg on docs iframe

driskull:
**Related Issue:** none

## Summary

set bg on docs iframe

image


driskull:
Closing. doesn't work right

Issue #699, opened 2019-12-26
Docs: pick list snippets


kat10140:
### Description 
#619 
### Which Component
pick list

PR #700, opened 2019-12-26
docs: calcite pick list snippets

kat10140:
**Related Issue:** #619 #699

### DIFF
--- a/src/components/calcite-pick-list/usage/basic.md
+++ b/src/components/calcite-pick-list/usage/basic.md
@@ -94,7 +94,7 @@ Renders compact list items with the text description and a bit of padding remove
       <!-- icon -->
     </calcite-action>
   </calcite-pick-list-item>
-  <calcite-pick-list-item text-label="OBJECTID" text-description="" value="OBJECTID">
+  <calcite-pick-list-item text-label="OBJECTID" value="OBJECTID">
     <calcite-action slot="secondaryAction" id="action-test">
       <!-- icon -->
     </calcite-action>
--- a/src/components/calcite-pick-list/usage/basic.md
+++ b/src/components/calcite-pick-list/usage/basic.md
@@ -4,17 +4,17 @@ Renders a basic pick list with radio buttons and actions on the right side.
 
 ```html
 <calcite-pick-list>
-  <calcite-pick-list-item text-heading="T. Rex" text-description="Arm strength impaired" value="trex">
+  <calcite-pick-list-item text-label="T. Rex" text-description="Arm strength impaired" value="trex">
     <calcite-action slot="secondaryAction">
       <!-- icon -->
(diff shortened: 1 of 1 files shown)


src/components/calcite-pick-list/usage/basic.md
```diff
@@ -0,0 +1,103 @@
...
+    text-label="ID"
+    text-description="This is an id for this field and people use it for things."
+    value="ID"
+  >
+    <calcite-action slot="secondaryAction" id="action-test">
+      <!-- icon -->
+    </calcite-action>
+  </calcite-pick-list-item>
+  <calcite-pick-list-item text-label="OBJECTID" text-description="" value="OBJECTID">
+    <calcite-action slot="secondaryAction" id="action-test">
```
driskull: Fixme: We may want to change this slot name. All of our other slot names are hyphenated like `secondary-action`. Since we are doing breaking changes, this would be a good time to change it. I looked at all the other slot names and this is the only one that is camel cased.

We should nail down the consistency on this.

 @pr3tori4n @jcfranco

src/components/calcite-pick-list/usage/basic.md
```diff
@@ -0,0 +1,103 @@
...
+  <calcite-pick-list-item
+    text-label="ID"
+    text-description="This is an id for this field and people use it for things."
+    value="ID"
+  >
+    <calcite-action slot="secondaryAction" id="action-test">
+      <!-- icon -->
+    </calcite-action>
+  </calcite-pick-list-item>
+  <calcite-pick-list-item text-label="OBJECTID" text-description="" value="OBJECTID">
```
driskull: nitpick: Do we need to have the `text-description` if it is empty?

src/components/calcite-pick-list/usage/basic.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+#### Compact
+
+Renders compact list items with the text description and a bit of padding removed from the card.
+
+```html
+<calcite-pick-list compact>
+  <calcite-pick-list-item
+    text-label="ID"
```
driskull: question:  Should we rename these props from `text-label` and `text-description` to just `label` and `description`? Seems like the text part is unnecessary. We usually have the `text-` prefix for props that are used for UI strings that apply to every item but not for unique values.

Consistency thing again. @pr3tori4n @jcfranco

kat10140: I think that's a great idea

src/components/calcite-pick-list/usage/basic.md
driskull: @kat10140 just remove the `text-description=""` here and we can merge. We'll address the other comments in a follow up issue.

Issue #701, opened 2019-12-26
Docs: shell snippets


kat10140:
### Description 
#619 
### Which Component
shell

kat10140:
dun

PR #706, opened 2019-12-27
docs: calcite shell snippets

kat10140:
**Related Issue:** #701 #619

### DIFF
--- a/src/components/calcite-shell/usage/advanced.md
+++ b/src/components/calcite-shell/usage/advanced.md
@@ -1,0 +1,87 @@
+#### Advanced
+
+Renders a shell with leading and trailing floating panels, action pad, block, flow.
+
+```html
+<calcite-shell>
+  <calcite-shell-panel slot="primary-panel" layout="leading" detached>
+    <calcite-action-pad id="action-pad" placement="side" slot="action-pad" hidden>
+      <calcite-action-group>
+        <calcite-action text="Add">
+          <!-- icon -->
+        </calcite-action>
+      </calcite-action-group>
+    </calcite-action-pad>
+    <calcite-action-bar slot="action-bar" theme="dark">
+      <calcite-action-group>
+        <calcite-action text="Add">
+          <!-- icon -->
+        </calcite-action>
+        <calcite-action text="Save" disabled>
+          <!-- icon -->
+        </calcite-action>
+        <calcite-action text="Layers" active indicator>
+          <!-- icon -->
+        </calcite-action>
+      </calcite-action-group>
+      <calcite-action-group>
+        <calcite-action text="Add">
+          <!-- icon -->
+        </calcite-action>
+        <calcite-action text="Layers" indicator>
+          <!-- icon -->
+        </calcite-action>
+      </calcite-action-group>
+    </calcite-action-bar>
+    <calcite-block collapsible  heading="Primary Content" summary="This is the primary.">
+      <calcite-block-content>
+        <calcite-action text="Side ActionPad" text-enabled id="action-pad-button" indicator>
+          <!-- icon -->
+        </calcite-action>
+        <calcite-action text="Anchored Shell Floating Panel" text-enabled id="shell-floating-panel-button">
+          <!-- icon -->
+        </calcite-action>
+        <calcite-action text="Over Shell Floating Panel" text-enabled id="shell-floating-panel-over-button">
+          <!-- icon -->
+        </calcite-action>
+      </calcite-block-content>
+    </calcite-block>
+    <calcite-block collapsible  heading="Additional Block" summary="Baby shark doo doo doo doo.">
+      <calcite-block-content>
+          <p>Cool thing.</p>
+      </calcite-block-content>
+    </calcite-block>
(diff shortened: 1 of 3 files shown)


src/components/calcite-shell/usage/advanced.md
```diff
@@ -0,0 +1,115 @@
+#### Advanced
+
+Renders a shell with leading and trailing floating panels, action bar/pad, block, flow, tip manager, footer
+
+```html
+<calcite-shell>
+  <calcite-shell-panel slot="primary-panel" layout="leading" detached>
+    <calcite-action-pad id="action-pad" placement="side" slot="action-pad" hidden>
```
driskull: actionpad doesn't have `placement`. If its hidden, can we remove it from the code?

src/components/calcite-shell/usage/advanced.md
```diff
@@ -0,0 +1,115 @@
...
+      <calcite-action-group>
+        <calcite-action text="Add">
+          <!-- icon -->
```
driskull: Maybe we can use `<calcite-icon>` here now that we have that component to use? Would apply to other usage files so this can be a follow up issue to address.

src/components/calcite-shell/usage/advanced.md
```diff
@@ -0,0 +1,115 @@
...
+        </calcite-action>
+        <calcite-action text="Layers" indicator>
+          <!-- icon -->
+        </calcite-action>
+      </calcite-action-group>
+    </calcite-action-bar>
+    <calcite-block collapsible  heading="Primary Content" summary="This is the primary.">
+      <calcite-block-content>
+        <calcite-action text="Side ActionPad" text-enabled id="action-pad-button" indicator>
```
driskull: Maybe w can change the text from `side actionpad` to something more general so it doesn't appear like it should work already.

Same with `Anchored shell floating panel` and `Over shell floating panel`

src/components/calcite-shell/usage/advanced.md
```diff
@@ -0,0 +1,115 @@
...
+          <!-- icon -->
+        </calcite-action>
+        <calcite-action text="Anchored Shell Floating Panel" text-enabled id="shell-floating-panel-button">
```
driskull: Can we get rid of any use of `id` attribute? Applies to multiple places

src/components/calcite-shell/usage/advanced.md
```diff
@@ -0,0 +1,115 @@
...
+                <calcite-action slot="secondaryAction" id="action-test">
+                </calcite-action>
+              </calcite-value-list-item>
+            </calcite-value-list>
+          </calcite-block>
+        </calcite-flow-item>
+      </calcite-flow>
+  </calcite-shell-panel>
+  <calcite-tip-manager slot="tip-manager" id="tip-manager">
+    <calcite-tip heading="The Red Rocks and Blue Water" text-thumbnail="This is an image of nature.">
```
driskull: The `text-thumbnail` doesn't show up unless there is a `thumbnail` defined. Maybe we should add it for these?

Can you try copying these snippets to a HTML document and see if everything looks good and nothing is confusing for a user?

src/components/calcite-shell/usage/basic.md
```diff
@@ -0,0 +1,61 @@
...
+
+```html
+<calcite-shell>
+  <div slot="shell-header">
+    <header class="header">
+      <h2 class="heading">Shell Header: My App</h2>
+    </header>
+  </div>
+  <p>Shell Content</p>
+  <!-- insert map or image here -->
```
driskull: maybe `<!-- insert map or fillable content here -->`

driskull: Same in other locations

src/components/calcite-shell/usage/basic.md
```diff
@@ -0,0 +1,61 @@
+#### Basic
+
+Renders a basic shell with a header and a footer.
+
+```html
+<calcite-shell>
+  <div slot="shell-header">
+    <header class="header">
+      <h2 class="heading">Shell Header: My App</h2>
```
driskull: Can we remove use of `class=` in the snippets? applies multiple places.

src/components/calcite-shell/usage/advanced.md
kat10140: that one is hidden bc it pops up on click of the upper right icon in the shell panel. I can take it out all together?

kat10140: i think taking it out is easier. i just removed it. better not to clutter it up any more since it's already a lot of code

src/components/calcite-shell/usage/advanced.md
kat10140: sounds good. I just took out the thumbnails because they were place holders, ha! but just put them back in and checked the code and i think it looks good enough for a rough sample

kat10140:
@driskull can you check me again? Do you think it would be helpful to put comments in there at all or not since stencil easy for us to make it pretty understandable?

kat10140:
DUN 🥓

PR #713, opened 2019-12-31
fix: focus button styles

driskull:
**Related Issue:** none

## Summary

- Focus styles where there are hover styles.


src/components/calcite-pick-list-item/calcite-pick-list-item.scss
```diff
@@ -8,7 +8,8 @@
   animation: calcite-app-fade-in var(--calcite-app-animation-time) var(--calcite-app-easing-function);
 }
 
-:host(:hover) {
+:host(:hover),
+:host(:focus) {
```
asangma: nitpick: This will likely never get applied since :host never gets focus, i.e. focus is getting set on `label` and `button`. Doesn't break anything tho.

Issue #703, opened 2019-12-26
Docs: tip manager snippets


kat10140:
### Description 
#619 
### Which Component
tip manager!

kat10140:
verf

PR #714, opened 2019-12-31
docs: calcite tip manager snippets

kat10140:
**Related Issue:** #619 #703


src/components/calcite-tip-manager/usage/basic.md
```diff
@@ -0,0 +1,43 @@
+#### Basic
+
+Renders a tip manager using tip group as well as single tips.
```
driskull: How does `Renders a tip manager using a group of tips and single tips.` sound?

kat10140: sounds good! i'll change it

PR #720, opened 2019-12-31
refactor!: Modify Tip API & remove unused utilities

driskull:
**Related Issue:** None

## Summary

BREAKING CHANGE: Modify Tip API & remove unused utilities

- Remove local storage support in Tip.
- Remove unnecessary slots in Tip.
- Add tip slot for thumbnail.
- Remove utilities and functions that are not being used.
- Update demos


jcfranco:
I like the simplification, but could you add some info on what prompted this? Is it to place tip management concerns on the app?

driskull:
> could you add some info on what prompted this? Is it to place tip management concerns on the app?

Mostly what prompted this was to ensure we aren't handling app concerns in the future and since we are approaching a major breaking changes release, these changes could go into it.

kat10140:
@asangma looooooook

PR #725, opened 2020-01-02
fix: action display on edge.

driskull:
**Related Issue:** None

## Summary

Action width not displaying correctly on Edge. This fixes it.


driskull:
@asangma can you review this?

Issue #734, opened 2020-01-06
Bug: Global styles have classnames that are not scoped to `calcite-app-components`


driskull:
### Summary

The following SCSS files have classnames that are vague and not scoped to calcite-app-components:

#### `_base.scss`

```scss
header {
  margin: 0;
  display: flex;
  align-items: center;
  justify-content: space-between;
  color: var(--calcite-app-foreground);
  fill: var(--calcite-app-foreground);
}

.heading {
  padding: 0;
  margin: 0;
}

.header .heading {
  flex: 1 0 auto;
  padding: var(--calcite-app-cap-spacing-half) var(--calcite-app-side-spacing-half);
}
```

#### `_type.scss`

```scss
.heading {
  font-weight: var(--calcite-app-font-weight-demi);
}

h1.heading {
  font-size: var(--calcite-app-font-size-3);
}

h2.heading {
  font-size: var(--calcite-app-font-size-2);
}

h3.heading {
  font-size: var(--calcite-app-font-size-1);
}
h4.heading,
h5.heading {
  font-size: var(--calcite-app-font-size-0);
}
```

#### Root setting default font sizes

The following CSS is being set for anywhere that includes the global css file. We should probably not be setting these for any project that uses our components.

```scss
  font-family: var(--calcite-app-font-family);
  font-size: var(--calcite-app-base-font-size);
  line-height: var(--calcite-app-line-height);
```

### Expected Behavior

These styles should be scoped so that someone else doesn't conflict with the names of these styles. Styles such as `heading` are easy to conflict with.

### Relevant Info 

The built global styles are not scoped to our components in any way.

@jcfranco can you confirm this?

jcfranco:
Verified locally.

Issue #733, opened 2020-01-06
Bug: Modify global styles to remove host


driskull:
### Summary

From Stencil support:

**Question:** Are we able to use `:host{}` in the global style?
**Answer:** `:host` no, i don’t think that’d work, they should use :root or html instead
:host would mean like the component host element, but for a global style on the document i’m not sure it’d know what host is

### Actual Behavior

Host is used in global includes

### Expected Behavior

Host should not be used in global includes. 

### Relevant Info 

We should probably create a mixin we can import to each component.

Host is used in `_base.scss`, and `_color.scss`

@jcfranco does this sound correct?

driskull:
@jcfranco We should discuss what we want to do.

jcfranco:
Verified locally.

PR #736, opened 2020-01-07
fix: Modify global styles to remove host

driskull:
**Related Issue:** #733 & #734

## Summary

fix: Modify global styles to remove host

- Moves all the global styles into a mixin that is then applied on each component.


driskull:
@jcfranco @asangma maybe this could be the first step for cleaning up the CSS. Then the second step could remove the header and heading styles from base/type.

src/assets/styles/_mixins.scss
```diff
@@ -1,5 +1,115 @@
 // Mixins
 
+@mixin calciteLightThemeVars {
```
asangma: Sorry if we discussed this, but I am not clear on why these vars are in a mixin. Same with the DarkThemeVars.

driskull: The light vars are stated twice so its in a mixin for that

driskull:
Closing, the other PR is better.

Issue #682, opened 2019-12-20
chore: Use calcite-icon to clean up <svg> usage


jcfranco:
Now that `calcite-icon` is available, it should be used to clean up markup using the SVG tag + path data.

jcfranco:
@kat10140 Can you help verify this one? There shouldn't be any more inlined SVGs. 💪

kat10140:
verified!

PR #737, opened 2020-01-07
chore(demos): replace inlined SVGs with calcite-icon

jcfranco:
**Related Issue:** #682 

## Summary

Replaces inlined SVGs with `calcite-icon`.

@pr3tori4n @asangma Does the following look OK? Previously, pink circle icons were sized slightly smaller than `size=s`.

```patch
-  <svg width="16" height="16" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
-    <circle cx="7" cy="7" r="7" fill="#f689d8" />
-  </svg>
+  <calcite-icon icon="circle" scale="s" filled style="color: #f689d8;"></calcite-icon>
```


kat10140:
looks like we still have two demos with svgs. (add action after init & full window reversed) I will go in and make those changes

PR #741, opened 2020-01-07
refactor: styles.

driskull:
**Related Issue:** #734, #733

## Summary

This approach uses inheritance instead of so many mixins. Let me know what you think.

I was noticing a lot of repeated `:root{}` and repeated `:host{}` selectors so this cleans that up.


src/assets/styles/_component.scss
```diff
@@ -0,0 +1,61 @@
+:host {
+  background-color: var(--calcite-app-background);
+  box-sizing: border-box;
+  color: var(--calcite-app-foreground);
+  font-family: var(--calcite-app-font-family);
+  font-size: var(--calcite-app-base-font-size);
```
driskull: @asangma I'm not sure if this should be `--calcite-app-base-font-size` or `--calcite-app-font-size-0`. Previously, the font-size was being set on the root but that is probably not a good idea since we don't want to modify everyone else's styles.

src/assets/styles/_component.scss
```diff
@@ -0,0 +1,61 @@
+:host {
+  background-color: var(--calcite-app-background);
```
driskull: I'm not sure if we should set a default `background-color` since there are a few components that already overwrite it and set it to transparent.

src/assets/styles/_root.scss
```diff
@@ -0,0 +1,7 @@
+:root {
```
driskull: - Only one root selector now
- Everything in global css is just under `:root`

src/assets/styles/_component.scss
```diff
@@ -0,0 +1,61 @@
+:host {
```
driskull: We could consider this being something we extend so that components could easily overwrite the background and there would be no two `:host` selectors.

src/assets/styles/_component.scss
asangma: There should be [a relevant var](https://github.com/Esri/calcite-app-components/blob/master/src/assets/styles/_color.scss#L28) from _color.

src/assets/styles/_component.scss
asangma: Should be `font-size-0`.

src/assets/styles/_component.scss
driskull: yes, so we should remove this and let each component decide the background?

src/assets/styles/_component.scss
driskull: Then `--calcite-app-base-font-size` would not be used anywhere correct?

src/assets/styles/_component.scss
asangma: That makes sense.

asangma: Would it work as a `_host` partial? `_component-host`?

asangma: Just a thought.

src/assets/styles/_component.scss
driskull: @asangma ^

asangma: It used to be used in **_base** to set the size in pixels for the page, and components were meant to use `rems`. We're also using that for our **_layout** spacing.

Wonder what a good strategy would be going forward. 🤔

driskull: @asangma Should our component project be setting page styles though? Seems wrong. Seems like those base HTML styles are more for a css library.

asangma: We have been setting it on `:root`. We could avoid setting a pixel font size at all and letting the app do that. If not set at the app level, it'll pick up the browser default which is normally `16px` (what we were setting it to).

driskull: Yeah, I think we should let the app do it.

asangma: K. So we should be able to nix that `base-font-size` var.

pr3tori4n:
@driskull I think mixins would actually create the effect you're desiring better than placeholders and extends. If you look at the generated global file, you'll see that the `:root` selector gets duplicated for every call to @extend because of how extending works in scss (combining selectors together). mixins would copy the styles into the existing selector, and prevent redudant :root selectors.

src/scss/global.scss
```diff
@@ -0,0 +1,9 @@
+:root {
+  @extend %component-timing;
```
pr3tori4n: Adding CSS comments here will be a nice QoL for inspecting these vars in the debug tools in browser. Will help provide context.

driskull:
I'll change them to mixins. Maybe the host one i'll leave as a placeholder.

src/scss/mixins/variables/_animation.scss
```diff
@@ -0,0 +1,10 @@
+$animation-time: 250ms !default;
+
+@mixin componentAnimationVars() {
+  // Component Animation Variables
```
pr3tori4n: Need to make these CSS comments or it won't propagate to output CSS. Same with the others.

src/scss/placeholders/_host.scss
```diff
@@ -0,0 +1,11 @@
+%component-host {
+  box-sizing: border-box;
```
pr3tori4n: Can we add a comment here as well?

driskull:
Added comments to code

src/scss/mixins/variables/_animation.scss
```diff
@@ -0,0 +1,10 @@
+$animation-time: 250ms !default;
+
+@mixin componentAnimationVars() {
+  /* Component animation variables */
```
jcfranco: Nitpick: IMO the vars are expressive enough to not need these comments.

driskull: It's just a nice separator in the output though. I'll leave it for now.

Issue #740, opened 2020-01-07
A11y: calcite-panel - set focus on dismiss button


AdelheidF:
### Summary
The smartMapping Gallery component info option needs a way to set the focus on the close button of a calcite-panel.

cc @driskull 

### Reproduction Steps

1.

AdelheidF:
Focus box seems big, but the focus works,

![image](https://user-images.githubusercontent.com/44481292/72558644-f7dd3100-3857-11ea-8e88-1adc6f04b3b2.png)

PR #742, opened 2020-01-07
feat: setFocus on dismiss button method  #740 

driskull:
**Related Issue:** #740 

## Summary

feat: #740 setFocus on dismiss button method


src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -102,6 +114,17 @@ export class CalcitePanel {
...
   };
 
+  // --------------------------------------------------------------------------
+  //
+  //  Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method()
+  async setFocusDismissButton() {
```
pr3tori4n: maybe just call it `focusDismissButton`

src/demos/panel/basic.html
```diff
@@ -53,6 +53,9 @@ <h2>Dismissible Panel</h2>
           dismissiblePanel.addEventListener("calcitePanelDismissedChange", function() {
             console.log("calcitePanelDismiss event");
           });
+          setTimeout(function() {
+            dismissiblePanel.setDismissFocus();
```
pr3tori4n: method name doesn't match.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -102,6 +114,17 @@ export class CalcitePanel {
...
   };
 
+  // --------------------------------------------------------------------------
+  //
+  //  Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method()
+  async focusDismissButton() {
```
jcfranco: Code looks good, but I had a general question. Are we OK with having individual focus methods for component parts? This does set a precedent for that. 

Just as an idea, we could use `setFocus()` for the default focus behavior and allow an argument for a specific part.

```ts
component.setFocus(); // default focus
component.setFocus("some-part"); // specialized focus
```

This does become tricky when there is only one part and no default focus behavior (like in calcite-panel 😅).

jcfranco: The name is now inconsistent w/ `calcite-action`: https://github.com/Esri/calcite-app-components/blob/9f02abe53f0ce530b90a2b8a5adeb1a290653085/src/components/calcite-action/calcite-action.tsx#L90

driskull: I like this idea. even though it is a little tricky.

src/components/calcite-panel/calcite-panel.e2e.ts
```diff
@@ -52,15 +52,15 @@ describe("calcite-panel", () => {
...
       withPeerDependencies: true
     });
 
     const panel = await page.find("calcite-panel");
 
-    const focusDismissButton = panel.callMethod("focusDismissButton");
+    const setFocus = panel.callMethod("setFocus");
 
-    expect(focusDismissButton).toBeInstanceOf(Promise);
+    expect(setFocus).toBeInstanceOf(Promise);
```
jcfranco: I think this also needs to assert on the final focused element.

driskull: I don't think focus works very well in e2e tests. Tried it but couldn't get it to work.

src/components/calcite-panel/calcite-panel.tsx
```diff
@@ -102,6 +116,19 @@ export class CalcitePanel {
...
 
+  // --------------------------------------------------------------------------
+  //
+  //  Methods
+  //
+  // --------------------------------------------------------------------------
+
+  @Method()
+  async setFocus(focusId?: FocusId) {
+    if (focusId === "dismiss-button") {
```
pr3tori4n: not sure the purpose of this argument. If the only valid value is dismiss-button, why not just have it hardedcoded to target that element and remove the argument altogether?

src/components/calcite-panel/calcite-panel.e2e.ts
pr3tori4n: maybe try 
```const someId = page.evaluate( () => {
const panel = document.querySelector("calcite-panel");
  const activeElement = panel.shadowRoot.activeElement;
  return activeElement.id; //or tagname or some attribute you can use to validate its the right one.
});

src/components/calcite-panel/calcite-panel.tsx
driskull: Because we may have the ability to focus a panel in the future.

driskull: Update: added ability to focus on container.

Issue #483, opened 2019-10-30
New Component: DNDList/SortableList


kat10140:
### Description
We want to make a reusable drag and drop component, as we are implementing something similar (via sortable) in many other components.  Having it as its own component will provide consistency and makes drag and droppable "lists" easier to implement in the future. 

#### User Stories
As an app developer, I want to be able to easily implement drag and drop within items in a panel. 

### Acceptance Criteria 
- Drag and drop container
- Wrapper for DnD items

#### Desired Behavior
To be able to drag and drop items vertically that are organized in a list like fashion.  This can be actual lists, block sections, etc. 

#### Desired Look & Feel
Items to have distinct wrapper and drag icon. (Alan?)

kat10140:
reassigning to @pr3tori4n (seems like you have a little more time right now)

kat10140:
Audit which components can be used with this. @asangma @pr3tori4n

pr3tori4n:
Block Components is the only addition for now.

pr3tori4n:
hey @kat10140 could you setup a kickoff meeting for this early next week? I think this will probably end up falling into grob glob.

kat10140:
Goal:
- broad story: general use drag and drop 
  - keyboard navigation paired with sortable (Adds event handlers for arrow keys)
  - consistent UX for mouse and keyboard users
- block is our immediate use case 
- as for wireframes, look at value list for now
  - v1 is just a vertical list, no grid

QA test case to evaluate heavily nested components. 

Harry make custom component for handle? Maybs?
Drag enabled property (or some prop) will still be passed down. This is how we'll use the space functionality.

Check in with Dhrumil with what he did on popup configuration? (Dragging whole block vs just using the handle). 

@kat10140 put all this in the first comment

pr3tori4n:
@kat10140 please verify.

kat10140:
Verified!! Woohoo!

PR #747, opened 2020-01-09
feat: drag n drop component 483

pr3tori4n:
**Related Issue:** #483 

## Summary

* Adds new `calcite-sortable-list` component, which can accept a list of any DOM nodes that can be sorted, so long as they have a handle with the same class as what's provided to the list (defaults to .handle).
* Adds `calcite-handle` component, which is a button that has the drag icon. This element has an activated prop which toggles based on if it's selected via keyboard, and fires a custom event when the up and down arrow keys are pressed when focused on the element.
* The list listens to the custom events fired by the handle to know when to trigger keyboard sorting.
* The block component was modified  to add the `drag-handle` boolean prop, which adds a `calcite-handle` to the heading area when true. False by default.


src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Private Methods
+  //
+  // --------------------------------------------------------------------------
+
+  setUpDragAndDrop(): void {
+    this.sortable = Sortable.create(this.el, {
+      handle: `.${this.handleClass}`,
```
driskull: question: should we allow for any selector instead of just a class? maybe the default could be `.calcite-handle`?

src/components/calcite-handle/calcite-handle.tsx
```diff
@@ -0,0 +1,95 @@
...
+  //
+  //  Events
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * Emmitted when the the handle is activated and the up or down arrow key is pressed.
+   * @event calciteHandleNudge
+   */
+  @Event() calciteHandleNudge: EventEmitter;
```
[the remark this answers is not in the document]
pr3tori4n: hmmm...you're not really moving the handle right you're moving the parent. @franco thoughts? I'm def open to renaming tho its a meh name.

src/components/calcite-handle/calcite-handle.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  // --------------------------------------------------------------------------
+  //
+  //  Render Methods
+  //
+  // --------------------------------------------------------------------------
+
+  render() {
+    return (
+      <button
```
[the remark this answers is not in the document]
pr3tori4n: doesn't seem necessary. The :host selector is still applied to the tag. I think you only need it if you either need the parent wrapper or if you need to modify its attributes.

src/components/calcite-handle/calcite-handle.tsx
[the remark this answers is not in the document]
driskull: Maybe we don't need this event and the keydown logic can just be added on the parent?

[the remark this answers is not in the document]
pr3tori4n: It's cleaner with the event, otherwise the event handler gets a bit messy when trying to find the handler in the composedPath.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,171 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * The class on the handle elements.
+   */
+  @Prop() handleClass = "handle";
```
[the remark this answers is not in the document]
pr3tori4n: no it needs to target the button.

src/components/calcite-handle/calcite-handle.tsx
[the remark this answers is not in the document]
driskull: Leaving as up/down for now. We can add other directions in the future.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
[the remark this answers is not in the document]
pr3tori4n: no needs to be the button so it can be focused

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,171 @@
...
+    }
+    if (appendInstead) {
+      sortItem.parentElement.appendChild(sortItem);
+    } else {
+      sortItem.parentElement.insertBefore(sortItem, this.items[buddyIndex]);
+    }
+
+    this.items = Array.from(this.el.children);
+
+    event.detail.handle.activated = true;
```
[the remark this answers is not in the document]
driskull: Can you comment @pr3tori4n ?

src/components/calcite-handle/calcite-handle.tsx
```diff
@@ -0,0 +1,95 @@
...
+        class={{ [CSS.handle]: true, [CSS.handleActivated]: this.activated }}
```
driskull: question: are these classes applied to the host or in the shadowDom? If on the host, can we move this inside the shadow dom so they aren't present on the host for a user to modify.

pr3tori4n: shadow dom.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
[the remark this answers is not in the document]
pr3tori4n: Using event.detail is clean and simple. Because the event is coming from a web component, the event.target will always equal the calcite-handle, and you'd have to inpsect the event.composedPath to get a reference to the button in the shadow dom. This is more complex code-wise, and it's more brittle if the handle happens to be inside another dom element or nested inside another web component.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+  // --------------------------------------------------------------------------
+  //
+  //  Properties
+  //
+  // --------------------------------------------------------------------------
+
+  /**
+   * The class on the handle elements.
+   */
+  @Prop() handleSelector = ".handle";
```
driskull: So is the `.handle` going into the shadowDom and the drag and drop supports it? Just wondering why it isn't just `calcite-handle`.

pr3tori4n: the host can't receive focus, its basically a div. The button is the focus item and the cornerstone of the handle functionality. When using the keyboard, when sortable sorts the items, the parent of the handle will be removed from the dom and re-inserted. This causes it to lose focus so we have to re-apply focus. That focus needs to go on the button, not the host.

driskull: @pr3tori4n does Sortable find the `.handle` in the shadowDOM? Is that how this works?

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (appendInstead) {
+      sortItem.parentElement.appendChild(sortItem);
+    } else {
+      sortItem.parentElement.insertBefore(sortItem, this.items[buddyIndex]);
+    }
+
+    this.items = Array.from(this.el.children);
+
+    event.detail.handle.activated = true;
+    event.detail.handle.shadowRoot.querySelector(this.handleSelector).focus();
```
driskull: fixme: I think we should add a `setFocus()` method on the handle instead of going through the shadowRoot. We already do this for the `calcite-action` and `calcite-panel`

src/components/calcite-sortable-list/calcite-sortable-list.tsx
driskull: fixme: Lets change this to `calcite-handle` by default with the `setFocus()` method.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
driskull: @pr3tori4n this wouldn't work in Edge anyway

pr3tori4n: It should now that Edge is on Chromium shouldn't it?

driskull: > It should now that Edge is on Chromium shouldn't it?

We should support older edge users regardless.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,188 @@
...
+  //
+  //  Private Properties
+  //
+  // --------------------------------------------------------------------------
+
+  @Element() el: HTMLCalciteSortableListElement;
+
+  @State() handleActivated = false;
+
+  items: Element[] = [];
```
driskull: question: should this be HTMLElement? since its coming chiildren of HTMLElement.

src/components/calcite-sortable-list/calcite-sortable-list.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+  observer = new MutationObserver(() => {
+    this.cleanUpDragAndDrop();
+    this.items = Array.from(this.el.children);
+    this.setUpDragAndDrop();
+  });
+
+  sortable: Sortable;
+
+  handle: Node = null;
```
driskull: question: Is this being used anywhere? What is the Node type?

pr3tori4n: good catch this is cleanup

src/components/calcite-sortable-list/calcite-sortable-list.tsx
pr3tori4n: changing this causes a TS error on lines `this.items = Array.from(this.el.children);`

Issue #738, opened 2020-01-07
Bug: Calcite-Value-List: when the list item value changes in tsx the old selected list item value is not deleted


kumarGayu:
### Summary
Calcite-Value-List: for a single select when the list item value changes in tsx the old selected value is not deleted

### Actual Behavior
e.detail has 2 values. one was the older value and the other one is newer selected value.

### Expected Behavior
Ideally list change event should have one item selected in e.detail for a single select.

### Reproduction Steps

1. have a single select list with multiple values.
2. cahnge the list value description synamically through tsx prop update.
3. now go and select the other item
4. e.detail has 2 values.

pr3tori4n:
@kat10140 please verify.

kat10140:
verified

PR #750, opened 2020-01-10
feat: lists react to value prop changes in items

pr3tori4n:
**Related Issue:** #738 

## Summary

This addresses the bug Kumar raised in which if the value of an item changed after that item was selected, the parent list's Map was not updating its keys. This is due to the mutationObserver not reacting to attribute changes - which is necessary since attribute changes are too frequent.
So we added a watcher to the value prop and emit a custom event when it changes, that the list can listen to and respond accordingly.


src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -127,6 +131,13 @@ export class CalcitePickListItem {
    */
   @Event() calciteListItemPropsUpdated: EventEmitter;
 
+  /**
+   * Emitted whenever the the item's value property is modified.
+   * @event calciteListItemPropsUpdated
```
jcfranco: The event name doesn't match.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -98,6 +98,10 @@ export class CalcitePickListItem {
    */
   @Prop({ reflect: true }) value!: string;
 
+  @Watch("value") valueWatchHandler(newValue, oldValue) {
+    this.calciteListItemValueUpdated.emit({ oldValue, newValue });
```
driskull: Should this be named `calciteListItemValueChange`? Events shouldn't be in past tense and we've been using `change` for events.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
driskull: Should this be named calciteListItemPropsChange? Events shouldn't be in past tense and we've been using change for events.

pr3tori4n:
@driskull can we make the existing event name change in a separate PR?

driskull:
Yes, it can be a separate pr for `calciteListItemPropsUpdated` but its a breaking change so we can either do it now or deprecate later.

pr3tori4n:
@driskull I added a test. Please take a second look.

pr3tori4n:
> 👍
> Can you link the issue for renaming the other event to this review?

I don't believe there is an issue for it. Mind creating one?

src/demos/value-list/valueUpdate.html
```diff
@@ -0,0 +1,57 @@
+<!DOCTYPE html>
```
jcfranco: For consistency, the demo filename should be kebab-cased.

driskull:
https://github.com/Esri/calcite-app-components/issues/773

Issue #751, opened 2020-01-10
Bug: [Block] unintentional toggle when control is focused and Space/Enter is pressed


jcfranco:
### Summary

Bubbled keyboard events (Space/Enter) can cause the block to unintentionally toggle.

### Actual Behavior

Block is toggled when a Space/Enter keyboard event is dispatched from a control element.

### Expected Behavior

!:point_up:

### Reproduction Steps

1. Open the following [test case](https://codepen.io/jcfranco-the-scripter/pen/zYxjZvW).
2. Focus on the control
3. Press Space or Enter key
4. Note how the block is toggled (unexpected)

jcfranco:
Installed.

kat10140:
@jcfranco looks like it's still happening with the test case you provided. should i test this somewhere else?

jcfranco:
@kat10140 This fix will be in the next release, so you'll need to test locally for now.

kat10140:
@jcfranco this is happening locally and on the test case you provided.

jcfranco:
It works fine using the latest: https://codepen.io/jcfranco-the-scripter/pen/VwYOBPz

1. Focus on the control text (`focus me and hit SPACE/ENTER`)
2. Press Space or Enter key
3. Block is not toggled. ✅

kat10140:
facepalm. i only changed the version of the stylesheets

PR #752, opened 2020-01-11
fix(block): prevent unintentional toggling from bubbled control space/enter keys

jcfranco:
**Related Issue:** #751 

## Summary

This updates `calcite-block` to prevent the default behavior on any bubbled Space/Enter keys from a slotted control element avoiding unwanted toggling.


src/components/calcite-block/calcite-block.tsx
```diff
@@ -115,6 +115,18 @@ export class CalciteBlock {
...
   };
 
+  onHeaderKeyDownOrUp = (event: KeyboardEvent) => {
+    const { currentTarget, key, target } = event;
+
+    if (key === " " || key === "Enter") {
+      const shouldBlockBubbledControlEvent = currentTarget !== target;
+
+      if (shouldBlockBubbledControlEvent) {
+        event.preventDefault();
```
pr3tori4n: Did you mean to do event.stopPropagation()? This wouldn't stop the event from bubbling.

jcfranco: That's intentional. I still want the event to propagate, but not fire a click event on the parent button element.

pr3tori4n: Is this a situation where we have nested interactive elements? That's usually an anti-pattern. There a new way we could archictect this structure so we don't need this?

jcfranco: I can't think of a way to restructure this at the moment. I could stop event propagation, but this means users would no longer be able to capture bubbled key events. Maybe that's fine for now since they can listen at the nested control level?

> Is this a situation where we have nested interactive elements? That's usually an anti-pattern. 

For my own purposes, could you share a link?

pr3tori4n: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

Verify here: https://validator.w3.org/#validate_by_input. Paste the following.
```<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <title>Document</title>
</head>
<body>
    <button>
        <input type="text" />
    </button>
</body>
</html>```

pr3tori4n:
Couple options you have if you want to avoid refactoring right now. 1) You could replace the button element with a span with role="button". This removes the default behavior or space or enter keys triggering a click.

Option 2) You could move the keydown event listener to a div wrapping the input, which would stopPropagation.

jcfranco:
Went with option 1 since it seems to be the cleanest one as it preserves bubbling and doesn't set `defaultPrevented` on the bubbled event.

jcfranco:
> Does this work fine in Edge as well?

And... it does not. Fixing.

>  Do we want a refactoring issue?

👍

jcfranco:
Will resubmit with refactored, :sparkles:cleaner!:sparkles:, structure.

PR #762, opened 2020-01-15
fix(block): prevent unintentional toggling when interacting with a slotted control

jcfranco:
**Related Issue:** #751

## Summary

This restructures `calcite-block`'s control slot to avoid unintentional toggling.

### DIFF
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -13,9 +13,14 @@
 
 .header {
   flex-grow: 1;
+  justify-content: flex-start;
 }
 
 .header-container {
+  position: relative;
+  display: flex;
+  align-items: center;
+
   & > .header {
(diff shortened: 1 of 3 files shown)


src/components/calcite-block/calcite-block.scss
```diff
@@ -79,3 +83,16 @@ calcite-loader[inline] {
   padding: 0 var(--calcite-app-side-spacing-three-quarters) var(--calcite-app-cap-spacing-half);
   position: relative;
 }
+
+::slotted([slot="control"]) {
+  position: absolute;
```
driskull: Could the control be positioned flex instead of absolute over the header?

jcfranco: That messes up the header layout.

jcfranco: Sorry, I should've added more info. I tried something similar, but I need the control to be placed above the header toggle. Otherwise, it looks weird when focused:

Screen Shot 2020-01-16 at 9 01 05 AM

PR #766, opened 2020-01-16
fix: only modify textEnabled when expand functionality is enabled

driskull:
**Related Issue:** None

## Summary

fix: only modify textEnabled when expand functionality is enabled

### DIFF
--- a/src/components/utils/CalciteExpandToggle.tsx
+++ b/src/components/utils/CalciteExpandToggle.tsx
@@ -28,15 +28,12 @@ function getClosestShellLayout(el: HTMLElement): CalciteLayout | null {
 
 export function toggleChildActionText({
   parent,
-  expand,
   expanded
 }: {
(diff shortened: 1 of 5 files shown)


src/components/utils/CalciteExpandToggle.tsx
```diff
@@ -28,12 +28,15 @@ function getClosestShellLayout(el: HTMLElement): CalciteLayout | null {
...
   parent,
+  expand,
   expanded
 }: {
   parent: HTMLElement;
+  expand: boolean;
   expanded: boolean;
 }): void {
-  parent.querySelectorAll("calcite-action").forEach((action) => (action.textEnabled = expanded));
+  expand &&
```
jcfranco: Instead of modifying this utility, could you decide whether to use `toggleChildActionText` at its call site based on `expand` (or some other condition)?

Issue #771, opened 2020-01-17
Bug: height issues with calcite-flow/calcite-flow-item


AdelheidF:
### Summary
If I set height to 100% the blocks inside the panel stretch out.
<img src=https://user-images.githubusercontent.com/44481292/72633154-b65a8d80-390c-11ea-9408-a3d349fbecf6.png width=200/>

If I don't set the height the Close/Cancel buttons are no longer at the bottom of the panel.
<img src=https://user-images.githubusercontent.com/44481292/72633189-ce321180-390c-11ea-8769-bbe7255d5e9b.png width=200/>

@asangma 

### Reproduction Steps

1.

AdelheidF:
Looks good now.

asangma:
Thanks!

asangma:
Assigning to @kat10140

kat10140:
ye!

PR #772, opened 2020-01-17
fix: nesting and height issues

asangma:
**Related Issue:** #771 

## Summary
Don't call it a comeback...but the 100% heights are making a comeback.
When in non-detached mode, the flow-item footers were not sticking to the bottom.

Outside components being but into flow-item will still need to set their own heights (i.e. to 100%).

Would be nice to get this into a patch maybe?

### DIFF
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -1,7 +1,7 @@
 :host {
   @extend %component-host;
   display: flex;
-  flex: 1 0 auto;
+  flex: 0 0 auto;
   flex-direction: column;
   border-radius: var(--calcite-app-border-radius);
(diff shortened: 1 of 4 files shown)


asangma:
Thanks @driskull!

pr3tori4n:
I don't think this is the right way to go. Making Shell > Flow > Block height 100% seems very specific to the mapmaker. This is a brittle approach that makes significant assumptions about the integration pattern.

Outside of this one use case, I see these heights either being inert (due to not being inside a container that has height: 100% , or causing problems by being inside a container with a defined height when the desire is not for the component to stretch.

asangma:
> seems very specific to the mapmaker

This is specific to full page data viz apps...the main aim of these components.

Issue #774, opened 2020-01-17
Tooling: Add script to generate a changelog preview


jcfranco:
## Summary 

Currently, the changelog is only available during the publishing process. It would be useful to have a way to preview this without having to mess with publishing scripts.

## Measure of Success 

Add a script to `package.json` that generates a changelog preview.

jcfranco:
Installed.

kat10140:
verified!

PR #775, opened 2020-01-18
chore(tooling): add script to preview changelog

jcfranco:
**Related Issue:** #774 

## Summary

This adds a new NPM script `previewChangelog` that outputs the contents of a changelog generated from the last release to the most recent commit.

I noticed that tagging isn't required for this, so I may also be able to move the tagging step over to the `release` script. cc @driskull 

cc @kat10140 @asangma


package.json
```diff
@@ -28,13 +28,14 @@
...
     "preversion": "npm ci && npm test",
+    "previewChangelog": "standard-version --skip.commit --skip.tag --silent && cat CHANGELOG.md && git reset --hard --quiet",
     "release": "git push --follow-tags && npm publish && ts-node ./support/githubReleaseFromChangelog.ts",
     "start": "concurrently --kill-others --raw \"npm run demoPageSync\" \"tsc --project ./tsconfig-demos.json --watch\" \"stencil build --dev --watch --serve\"",
     "test": "stencil test --spec --e2e",
     "test:headed": "stencil test --spec --e2e --devtools",
     "test:watch": "stencil test --spec --e2e --watch",
     "test:watchAll": "stencil test --spec --e2e --watchAll",
-    "version": "npm run build && npm run docs && standard-version && git add ."
+    "version": "npm run build && npm run docs && standard-version --skip.bump --skip.commit --skip.tag && git add ."
```
pr3tori4n: suggestion: Maybe spilit `standard-version --skip.bump --skip.commit --skip.tag` into its own command to make this one easier to read.

.versionrc.json
```diff
@@ -1,9 +1,3 @@
 {
-  "changelogHeader": "# Changelog\n\nThis document maintains a list of released versions and changes introduced by those versions.\nThis project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)\n",
-  "skip": {
-    "bump": true,
-    "changelog": false,
-    "commit": true,
-    "tag": true
```
pr3tori4n: noticing both scripts are skipping commit and tag. Could we keep those two here under the global config?

jcfranco: I had the same thought, but sided on keeping 'skips' together instead of having to look in two places. Separating them out into commands like you suggested should clean this up, methinks.

package.json
jcfranco: I did a little digging and applied some `yargs` magic to override a top-level skip (there's no example like this in the `standard-version` doc)  to keep this clean.

Issue #543, opened 2019-11-11
A11y: Selected state of CalciteValueList is a little too subtle


driskull:
### Summary

Does the selected state of the value-list seem a little too subtle? At first, it wasn’t easy to tell which was selected or not.

### Actual Behavior

The selected visual indication can be mistaken for an outline or not recognized at all.

### Expected Behavior

More distinct representation between which items are selected or not selected.

### Reproduction Steps

1. View `demos/calcite-pick-list.html`
2. Select an item
3. Notice how it doesn't differ enough from an unselected state

### Relevant Info 

![image](https://user-images.githubusercontent.com/1231455/68626224-b911b000-048f-11ea-9dfa-5042e94af058.png)

### Working W3C Example/Tutorial 

N/A

asangma:
@driskull  Would you say that #615 would handle this?

driskull:
I think it would help. The selected state of an item is still just a 1px border. Is that enough for a user to understand the item is selected? Might want to test that.

driskull:
IMO, its still a bit difficult to identify if the 3rd one is selected or not.

![image](https://user-images.githubusercontent.com/1231455/70642423-09497280-1bf4-11ea-9eea-75cfede52213.png)

driskull:
@mitc7862 please advise.

driskull:
@asangma should the selected item's background color be different? Should there be a thick border-left to identify a selected item?

driskull:
It's better, but I don't think the "selected" state is still quite enough. I'll close the issue but I have a feeling this may come up from UX testing.

PR #776, opened 2020-01-21
fix(value-list-item): focus state fix

asangma:
**Related Issue:** #543

## Summary
- updated focus style
- updated selected style
- refactored style to be more on value-list-item and less so on pick-list-item

Note: this is a temporary update until the broader focus designer are in.


asangma:
@driskull Might be cool to get it into this release.

Issue #469, opened 2019-10-25
New Component: Wrapper for calcite-switch that conforms to panel layouts


asangma:
### Description
We regularly need the `calcite-switch` component to act as a "row" in our panel interfaces. Out of the box, `calcite-switch` will now layout as needed. 

### Acceptance Criteria 
* require no additional styling
* use `calcite-switch`
* events? @jcfranco could use your help on this.

#### Desired Look & Feel
![image](https://user-images.githubusercontent.com/12503298/67609906-20b3c580-f744-11e9-9818-17afa512fbe8.png)

asangma:
Per discussion from the last calcite-app-components meeting, this is a go.

jcfranco:
![Thunderbirds+are+go+_79f02300ba769e21bb09807c0edcfb2a](https://user-images.githubusercontent.com/197440/67811445-09cde580-fa5a-11e9-9743-abae7b2a9d2b.gif)

kat10140:
cc @asangma maybe move to calcite-components as an attribute?

asangma:
Moving to this https://github.com/Esri/calcite-components/pull/248

PR #777, opened 2020-01-21
feat: new switch-row component

asangma:
**Related Issue:** #469 

## Summary
New component which adds a layout container to a `calcite-switch`.

`scale` property from `calcite-switch` can be passed through via a property of the same name.

@jcfranco or @driskull Could use some help particularly with the **setFocus** method...and if I need that and stuff.

@kat10140 Thought it would be better to keep this in calcite-app-components since the layout is particular to panels.


macandcheese:
As a FYI the `calcite-label` component I'm working on for inputs will be able to achieve these layouts as well with "positioning" props for trailing, space between, etc. This isn't really styled yet but I'm hoping for it to support all of the following where clicking anywhere on the row will toggle the checkbox / switch / radio button when we add that: 

Screen Shot 2020-01-21 at 3 36 57 PM

asangma:
@macandcheese Oh dang. Cool man. Would be nice if we could use this in [block's section](https://esri.github.io/calcite-app-components/?path=/story/components-calcite-block--basic) with `toggleDisplay="switch"`.

driskull:
Yeah, lets let the label handle this. Otherwise we'll have to create a bunch of wrapping layout components.

asangma:
nuked

macandcheese:
I'll try and spend some more time on inputs/labels this week to get these in there and working, currently they are super WIP, or feel free to take a look on that open PR (clicking the label only focuses inputs for now, still need to make it work for input and checkbox).

macandcheese:
Just added (very quick) functionality to the open PR for input / label components: https://github.com/Esri/calcite-components/pull/248

Provides a few positioning options: `@Prop({ mutable: true, reflect: true }) appearance:
    | "inline"
    | "inline-space-between"
    | "default" = "default";` (default being not inline)

You can then just place your text wherever you'd like it to appear next to your checkbox/switch/(radio button which hasn't been built yet):

```
<calcite-label appearance="inline">
  <calcite-switch></calcite-switch>
  Text trailing inline
</calcite-label>

<calcite-label appearance="inline">
  Off
  <calcite-switch></calcite-switch>
  On
</calcite-label>

<calcite-label appearance="inline">
  Text leading inline
  <calcite-checkbox></calcite-checkbox>
</calcite-label>

<calcite-label appearance="inline">
  <calcite-checkbox></calcite-checkbox>
  Text trailing inline
</calcite-label>

<calcite-label appearance="inline-space-between">
  Text leading inline-space-between
  <calcite-switch></calcite-switch>
</calcite-label>

<calcite-label appearance="inline-space-between">
  <calcite-switch></calcite-switch>
  Text trailing inline-space-between
</calcite-label>
```

asangma:
Cool cool @macandcheese 🆒

Issue #763, opened 2020-01-15
Bug: [calcite-block] Broken toggle behavior in Edge


jcfranco:
### Summary

Toggling doesn't work as expected in Edge when you have a block with a control and content.

### Actual Behavior

With the aforementioned setup, blocks can be collapsed, but not expanded.

### Expected Behavior

Consistent toggling behavior.

### Reproduction Steps

Using the following snippet, try to toggle the block.

```html
<calcite-block heading="With control" summary="has inline input" open collapsible>
  <label slot="control">test <input placeholder="I'm a header control"/></label>
  <calcite-block-section text="Animals" open>
    <img alt="demo" src="https://placeimg.com/640/480/animals" />
  </calcite-block-section>
</calcite-block>
```

### Relevant Info 

Microsoft Edge 44.17763.831.0
Microsoft EdgeHTML 18.1776

jcfranco:
Installed.

kat10140:
verified!

PR #785, opened 2020-01-24
fix(block): fix block toggling in Edge

jcfranco:
**Related Issue:** #763 

## Summary

This restores block toggling in Edge. The issue was caused by the 'has content' logic, which caused the block to never open after being closed (`el.children` in Edge includes the elements in the render function, tripping up the logic).


driskull:
Murge?

jcfranco:
Murgled!

Issue #791, opened 2020-01-28
Bug: Block with new extra padding!


asangma:
### Summary
Extra padding in Block. 👀 

Seems to be coming from [here](https://github.com/Esri/calcite-app-components/blob/master/src/components/calcite-block/calcite-block.scss#L86).

### Actual Behavior
![Screen Shot 2020-01-28 at 3 46 22 PM](https://user-images.githubusercontent.com/12503298/73315676-93ba5580-41e5-11ea-8356-8d230b516f33.png)

### Expected Behavior
![image](https://user-images.githubusercontent.com/12503298/73315770-d54b0080-41e5-11ea-9057-9e40aee749c2.png)

### Reproduction Steps

1. Blocks!

### Relevant Info 
@jcfranco Looks like this was a fix for something in Edge? I'm not able to easily test in Edge. :(

asangma:
Not this is nice for stuff in `.content` but seems to be getting applied to stuff in the `.header-container`.

asangma:
@jcfranco I'm also seeing some updates to the header that seem to be causing issues when the `heading` gets long.

![image](https://user-images.githubusercontent.com/12503298/73371443-8e531e80-426a-11ea-997f-f404d46ff30c.png)

jcfranco:
Installed.

jcfranco:
>  I'm also seeing some updates to the header that seem to be causing issues when the heading gets long.

@asangma I'll create a separate issue for this one.

asangma:
Verified on demos and in MVB locally.

PR #793, opened 2020-01-29
fix(block): Avoid extra padding applied to slotted icon content

jcfranco:
**Related Issue:** #791 

## Summary

This removes unwanted padding on the icon slot. This was a regression from #763 (oversight on my part 😅).

### DIFF
--- a/.storybook/utils.ts
+++ b/.storybook/utils.ts
@@ -50,7 +50,8 @@ export interface SimpleAttribute {
   value: string | boolean | number;
 }
 
-export type Attributes = (KnobbedAttribute | SimpleAttribute)[];
+export type Attribute = KnobbedAttribute | SimpleAttribute;
+export type Attributes = Attribute[];
 
 export const createComponentHTML = (tagName: string, attributes: Attributes, contentHTML?: string) =>
   `<${tagName} ${attributes.map(({ name, value }) => `${name}="${value}"`).join(" ")}>${contentHTML}</${tagName}>`;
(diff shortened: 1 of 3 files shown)


asangma:
👀

asangma:
@jcfranco I guess I'm not clear on the need for the `::slotted(*)` selector. 🤔

jcfranco:
> I guess I'm not clear on the need for the ::slotted(*) selector. 🤔

That's the only way I could get the consistent padding for content (default slot) across browsers. Nesting `::slotted` anywhere else would cause Edge to ignore it. Let me know if you have a cleaner way.

asangma:
Could apply the padding to `.content`?

jcfranco:
That has a side effect when you have a block w/ no content (`.container` should have no padding when empty). E.g., block w/ heading + icon/control.

asangma:
I see. `:empty`?

jcfranco:
I'll try that out. I can't recall if I had issues with it. Stay tuned...

jcfranco:
WRT `:empty`, this won't work since it's never empty due to the slot. 

After thinking a little bit more on the extra padding, I think you’re right and we can always apply it to the content node. My changes were trying to address a problem only visible when you set the ___block to be open without any content___ (hence the visible empty content), which is an invalid use case.

jcfranco:
@asangma Can you review again? I'll tidy up the story file, but the rest should stay the same.

Issue #759, opened 2020-01-15
Enhancement: calcite-action - add scale option


AdelheidF:
### Description

Currently the calcite-action is using up a lot of width/height when used inside a 'text box'. It would be nice if there was an option to have a small size.

![image](https://user-images.githubusercontent.com/44481292/72474342-74f7a000-379d-11ea-97e5-84c4f9789d6f.png)

From the labeling config design:
![image](https://user-images.githubusercontent.com/44481292/72474498-bbe59580-379d-11ea-9ce7-f2f7c2fb0d59.png)

And could we use a small version then in pick-list so there's less blank space?, maybe `compact=true` works there too.
<img src=https://user-images.githubusercontent.com/44481292/72482459-9ca53300-37b2-11ea-8153-3ecade00a5ee.png width=200/><img src=https://user-images.githubusercontent.com/44481292/72482543-ed1c9080-37b2-11ea-9152-c0c6cc9d0102.png width=200/><img src=https://user-images.githubusercontent.com/44481292/72482579-0d4c4f80-37b3-11ea-9842-a6ec62222669.png width=200/>

cc @asangma

driskull:
@asangma what do you think? If we add scale, would we get rid of compact?

asangma:
@driskull Yeah...better to follow our other patterns and use **CalciteScale**.

asangma:
@AdelheidF this is installed but not yet released yet. Feel free to start adding the prop to any actions you deem in need of small-ness.

asangma:
@kat10140 Make for the verfee?

kat10140:
verified!

AdelheidF:
Looks good

![image](https://user-images.githubusercontent.com/44481292/73780674-c1eae880-4743-11ea-9015-6cd5f50a140b.png)

Will add it for map viewer beta 4

PR #796, opened 2020-01-30
fix(action): style updates for compact mode

asangma:
**Related Issue:** #759

## Summary
Additionally specific selectors for compact mode now that `scale` is a prop.

cc @kat10140


asangma:
Thanks @driskull!

Issue #753, opened 2020-01-11
Enhancement: ValueListItem - add removable property


asangma:
### Description
Add a property that automatically adds an Action with an X icon at the trailing end of the item and perhaps emits a "delete" event when clicked.

### Acceptance Criteria 
Existing implementation using Action in the current slot should still work.

### Relevant Info 
Current use case is for Charting where the ValueListItem needs both a graphic and and X.
![image](https://user-images.githubusercontent.com/12503298/72195453-38a3f880-33c7-11ea-8577-d40ace54f948.png)

#### Which Component
PickList and PickListItem

#### Example Use Case
See above screen.

cc @kumarGayu

pr3tori4n:
@asangma please verify.

asangma:
@pr3tori4n Looks like the layout stuff is in there nicely, but I am not seeing any `removable` property or associated event.

pr3tori4n:
@asangma What Adelheide is asking for is already possible by passing in more than one action using the `secondary-action` slot. This pattern is really flexible in that it will allow any users of the pick list to put whatever action(s) they need (using the most appropriate icon). It also means we won't be a bottleneck in needing to add specific options paired with actions as enhancements.

I advocate we continue with that pattern
```<calcite-pick-list-item>
  <calcite-action slot="secondary-action" id="colorPicker" onClick={...} />
  <calcite-action slot="secondary-action" id="delete" onClick={...} />
</calcite-pick-list-item>```

We hadn't tested having more than one action slot until now so a small CSS change was needed in order for them to display properly. That's been merged so I think the use case is now being served.

asangma:
It's good that we've addressed Adelheide's need. 👍 
This was a request coming from @kumarGayu.
This functionality is akin to the selected functionality and is pretty core to how value-list-items are used. Looking at the use cases currently in production, the remove functionality is common.

Attributes for SmartMapping
![image](https://user-images.githubusercontent.com/12503298/73965577-4c108980-48c9-11ea-9fb4-45b2e9fd656e.png)

Attributes for Popup Config
![image](https://user-images.githubusercontent.com/12503298/73965635-60ed1d00-48c9-11ea-9caa-bc5037b8bff9.png)

And we will see a lot more of this with charting, analysis, etc.

asangma:
@pr3tori4n Should we make an issue for Adelheid's request that is specific to the multiple secondary-action layout concerns?

pr3tori4n:
[...]

PR #798, opened 2020-01-30
feat: multiple actions appear side-by-side

pr3tori4n:
**Related Issue:** #753 

## Summary

What Adelheide is asking for is already possible by passing in more than one action using the same slot. The only thing missing was displaying them properly. Today they appear one atop of the other. With this small CSS change, they will appear side by side.

QUESTION: Matt brought up making the slot name plural `secondary-action(s)`. I don't mind the update. All slots by nature can accept multiple slotted elements - so this may speak to our slot naming conventions a bit.


src/demos/value-list/doubleAction.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
```
jcfranco: This filename's casing is inconsistent with existing demos.

asangma:
👀

pr3tori4n:
@asangma yea can you add styles for the other cases?

src/demos/value-list/doubleAction.html
jcfranco: @pr3tori4n This was not resolved before merging. Can you fix it?

Issue #780, opened 2020-01-23
Enhancement: Calcite-Filter should not query shadowRoot


driskull:
### Description

Enhancement: Calcite-Filter should not query shadowRoot

- [calcite-filter](https://github.com/Esri/calcite-app-components/blob/078682d73524d433bbf5a282ad855f291456fd02/src/components/calcite-filter/calcite-filter.tsx#L100)

```tsx
clear = (): void => {
    this.el.shadowRoot.querySelector("input").value = "";
    this.empty = true;
    this.calciteFilterChange.emit(this.data);
  };
```

@pr3tori4n does this work in Edge?

It would be better if it instead stored the node as a variable to reference it and clear its value.
```tsx
<input type="text" ref={(el) => this.textInput = el as HTMLInputElement} />
```

### Acceptance Criteria 

Shadowroot isn't queried for a node and the node is instead stored as a variable

#### Which Component

CalciteFilter

driskull:
Seems to work in edge. It would still be better to have a reference to the element instead of querying for it.

pr3tori4n:
@driskull please verify.

PR #799, opened 2020-01-31
fix: Replace shadowroot querying with textInput reference.

pr3tori4n:
**Related Issue:** #780 

## Summary

Replace shadowroot querying with textInput reference.


driskull:
@pr3tori4n is this mergeable? Can we get it merged before release if so?

driskull:
@pr3tori4n can we assign PRs to milestones too? That way we can easily know what milestone the PR went into without having to dig through commit log.

PR #810, opened 2020-02-05
fix: Action - Remove padding when used without an icon

driskull:
Fix display for text without icon. Previously, there was some padding to the left of the text for no reason.

**Related Issue:** None

## Summary

@asangma I noticed that the use of an action without an icon produced padding unnecessarily.

image


src/components/calcite-action/calcite-action.tsx
```diff
@@ -119,6 +119,8 @@ export class CalciteAction {
   renderIconContainer(): VNode {
     const { loading } = this;
 
+    const hasIcon = !!this.el.children.length;
```
jcfranco: Did Stencil fix this recently? I remember I had an issue with similar logic in Edge.

From https://github.com/Esri/calcite-app-components/pull/785:

> This restores block toggling in Edge. The issue was caused by the 'has content' logic, which caused the block to never open after being closed (`el.children` in Edge includes the elements in the render function, tripping up the logic).

driskull: dangit. Do you know a better way to figure out if the element has content?

jcfranco: Not at the moment. 😢

driskull:
cc @kevindoshier. Noticed this on mapviewer

PR #812, opened 2020-02-08
feat: deprecate layout in favor of position

driskull:
**Related Issue:** None

## Summary

- Deprecate use of `layout` property.
- Add `position` property which can be `start` or `end`.

https://github.com/Esri/calcite-components/pull/325#discussion_r376087970

### DIFF
--- a/src/components/calcite-action-bar/calcite-action-bar.tsx
+++ b/src/components/calcite-action-bar/calcite-action-bar.tsx
@@ -1,6 +1,6 @@
 import { Component, Element, Event, EventEmitter, Host, Prop, Watch, h } from "@stencil/core";
 
-import { CalciteLayout, CalciteTheme } from "../interfaces";
+import { CalciteLayout, CalcitePosition, CalciteTheme } from "../interfaces";
 
 import { CalciteExpandToggle, toggleChildActionText } from "../utils/CalciteExpandToggle";
 
@@ -60,8 +60,14 @@ export class CalciteActionBar {
   /**
    * Arrangement of the component. Leading and trailing are different depending on if the direction is LTR or RTL. For example, "leading"
    * in a LTR app will appear on the left.
+   * @deprecated()
    */
   @Prop({ reflect: true }) layout: CalciteLayout;
+
+  /**
+   * Arranges the component depending on the elements 'dir' property.
+   */
+  @Prop({ reflect: true }) position: CalcitePosition;
 
   /**
    * Used to set the component's color scheme.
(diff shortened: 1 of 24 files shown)


src/components/calcite-action-bar/calcite-action-bar.tsx
```diff
@@ -58,9 +58,7 @@ export class CalciteActionBar {
   @Prop() textCollapse = "Collapse";
 
   /**
-   * Arrangement of the component. Leading and trailing are different depending on if the direction is LTR or RTL. For example, "leading"
-   * in a LTR app will appear on the left.
-   * @deprecated()
+   * @deprecated Use `position` instead.
```
jcfranco: WDYT of still keeping the old doc while calling out the deprecation notice?

```
/** 
 * @deprecated since <version> - use <alternative> instead.
 * 
 * <prop doc>
 */
```

cc @kat10140

macandcheese:
FWIW we are still using `layout` for upcoming stepper - "horizontal/vertical" orientation.

driskull:
I think your use of `layout` makes more sense. Ours is being used more for start/end position.

src/components/calcite-action-bar/calcite-action-bar.tsx
```diff
@@ -119,15 +125,20 @@ export class CalciteActionBar {
   // --------------------------------------------------------------------------
 
   renderBottomActionGroup() {
-    const { expanded, expand, textExpand, textCollapse, el, layout, toggleExpand } = this;
+    const { expanded, expand, textExpand, textCollapse, el, layout, position, toggleExpand } = this;
+    const positionFallback: CalcitePosition = layout
```
jcfranco: Nitpick: `position` should have higher precedence in the odd case both are provided.

src/components/calcite-shell-panel/calcite-shell-panel.e2e.ts
```diff
@@ -78,11 +74,11 @@ describe("calcite-shell-panel", () => {
     expect(eventSpy).toHaveReceivedEvent();
   });
 
-  it("leading layout property should have action slot first ", async () => {
+  it("leading position property should have action slot first ", async () => {
```
jcfranco: Nitpick: we could keep the tests until the prop is removed.

src/components/calcite-action-bar/calcite-action-bar.tsx
kat10140: 👍🏼

Issue #784, opened 2020-01-24
Enhancement: Create utility method to find out if a slot exists


driskull:
### Description

`this.el.querySelector([slot=${SLOTS.secondaryAction}])` is very common now it seems.

Maybe we should have a utility for this like `hasSlot(el: HTMLElement, slotName:string): boolean;`?

### Acceptance Criteria 

utility method to find out if a slot exists

### Relevant Info 

https://github.com/Esri/calcite-app-components/pull/783#discussion_r370507954

#### Which Component

All

#### Example Use Case

Above

jcfranco:
I would like to propose the following:

```tsx
getSlotted<T extends HTMLElement = HTMLElement>(
  element: HTMLElement, 
  slotName: string, 
  options?: { all: boolean }
): T | NodeListOf<T>;
```

This has the flexibility of getting the slotted element(s) as well as making it easy to check for their existence:

```tsx
const hasSomeSlotted = !!getSlotted(el, SLOTS.some);
const slottedItem = getSlotted(el, SLOTS.item);
const slottedItems = getSlotted(el, SLOTS.item, { all: true });
```

The signature could be simplified by taking a flag instead of options too.

jcfranco:
Installed!

jcfranco:
@driskull I'll create a new refactor issue to start using the util.

PR #814, opened 2020-02-11
refactor(dom): Add util for finding slotted elements

jcfranco:
**Related Issue:** #784 

## Summary

After writing the tests, it became evident that the API could be simpler (i.e., `all` flag). Would it be preferable to have 2 utils `getSlotted` & `getSlottedAll` or just accept a 3rd boolean (instead of an object) for `all`?


driskull:
I like the 3rd boolean option rather than two methods. Or we just always do all and they can test for an empty array?

jcfranco:
> Or we just always do all and they can test for an empty array?

I kind of like this better. I'll make some changes.

driskull:
> I kind of like this better. I'll make some changes.

Sounds good!

src/components/utils/dom.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        document.body.firstChild.remove();
+      }
+    });
+
+    it("returns elements with matching slot name", () => {
+      expect(getSlotted(document.body, slotNameMultiple)).toHaveLength(2);
+      expect(getSlotted(document.body, slotNameSingle)).toHaveLength(1);
+    });
+
+    it("returns null when no results", () => expect(getSlotted(document.body, "non-existent-slot")).toHaveLength(0));
```
driskull: nitpick: rename to "returns empty array when no results"?

src/components/utils/dom.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+      }
+
+      customElements.define("slot-test", SlotTest);
+    });
+
+    beforeEach(() => {
+      document.body.innerHTML = `
+      <slot-test>
+        <h2 slot=${slotNameMultiple}>😃</h2>
```
driskull: :)

src/components/utils/dom.spec.ts
jcfranco: D-oh. Good catch.

PR #820, opened 2020-02-12
feat(pick-list): add removable prop

pr3tori4n:
**Related Issue:** #753 

## Summary

add removable prop to pick list


src/components/calcite-pick-list-item/calcite-pick-list-item.e2e.ts
```diff
@@ -61,4 +63,17 @@ describe("calcite-pick-list-item", () => {
...
   });
+
+  it("displays the remove button when removable is true", async () => {
+    const page = await newE2EPage();
+
+    await page.setContent(
+      `<calcite-pick-list-item text-label="test" value="example" removable></calcite-pick-list-item>`
+    );
+
+    page.debugger();
```
jcfranco: Leftover?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -207,11 +222,23 @@ export class CalcitePickListItem {
...
+    );
+  }
+
   renderSecondaryAction() {
     const hasSecondaryAction = this.el.querySelector(`[slot=${SLOTS.secondaryAction}]`);
-    return hasSecondaryAction ? (
+    return hasSecondaryAction || this.removable ? (
       <div class={CSS.action}>
         <slot name={SLOTS.secondaryAction} />
+        {this.renderRemoveAction()}
```
jcfranco: Nitpick: Would moving this into the slot as default content help against users providing a secondary action + removable?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -58,6 +58,11 @@ export class CalcitePickListItem {
     this.calciteListItemPropsChange.emit();
   }
 
+  /**
+   * Set this to true to display a remove action that removes the item from the list.
+   */
+  @Prop({ reflect: true }) removable = false;
```
driskull: Should we make this an optional prop so that users know they don't have to set it?

```
@Prop({ reflect: true }) removable?: boolean;
```

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -207,11 +222,23 @@ export class CalcitePickListItem {
     );
   }
 
+  renderRemoveAction() {
+    if (!this.removable) {
+      return null;
+    }
+    return (
+      <calcite-action class={CSS.remove} text="remove" onClick={this.removeClickHandler}>
```
driskull: Can we add a prop for the `remove` text` and store the value in the resources file as a TEXT const?

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -207,11 +222,23 @@ export class CalcitePickListItem {
...
+        <calcite-icon scale="s" icon="x" />
```
driskull: Can we store the icon in an `ICONS` const in the resources file? Like `ICONS.close`

src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -49,6 +49,11 @@ export class CalciteValueListItem {
    */
   @Prop() metadata?: object;
 
+  /**
+   * Set this to true to display a remove action that removes the item from the list.
+   */
+  @Prop({ reflect: true }) removable = false;
```
driskull: Should we make this an optional prop so that users know they don't have to set it?

```
@Prop({ reflect: true }) removable?: boolean;
```

asangma: +1

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
asangma: Please add `scale="s"` to this Action

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
pr3tori4n: not sure I follow. You mean add this manually to the lightDOM instead of to the shadow dom? Editing the light dom seems risky as we might accidentally override user defined slotted content.

pr3tori4n:
Adding @driskull as an assignee in case this goes beyond today.

jcfranco:
@driskull @asangma Could you review again? 🙇 https://github.com/Esri/calcite-app-components/pull/820/commits/26b2a67460a488be6fba2283688bafb007a5f91a has the main changes (adding test and updating demo page to leverage emitted event for item removal).

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -54,17 +54,22 @@ export class CalcitePickListItem {
...
 
-  @Watch("metadata") metadataWatchHandler() {
+  @Watch("metadata") metadataWatchHandler(): void {
     this.calciteListItemPropsChange.emit();
   }
 
+  /**
+   * Set this to true to display a remove action that removes the item from the list.
+   */
+  @Prop({ reflect: true }) removable? = false;
```
driskull: I think this shoud be `@Prop({ reflect: true }) removable?: boolean;`

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
```diff
@@ -200,18 +227,36 @@ export class CalcitePickListItem {
...
+    }
+
+    return (
+      <calcite-action
+        scale="s"
+        class={CSS.remove}
+        text={this.textRemove}
+        onClick={this.removeClickHandler}
+      >
+        <calcite-icon scale="s" icon={ICONS.remove} />
```
driskull: Now that the icon is part of the action, you can just use the icon prop on the action and remove this node.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
jcfranco: This wouldn't be correct though since the default value would be `undefined` when querying the prop on the instance.

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
jcfranco: 💥

jcfranco: Thanks for the tip!

src/components/calcite-pick-list-item/calcite-pick-list-item.tsx
driskull: hmm then I don't think the `?` will do anything here. It only seems to work when a prop is undefined at the get go.

jcfranco: We can tackle that separately. I think there are other optional props that will need updating.

Issue #795, opened 2020-01-29
Enhancement: calcite-pick-list minimalist style


AdelheidF:
### Description

Would you consider a calcite-pick-list style that looks something like this?, notice the minimalist spacing/padding. Label might be limited to one line with optional ellipses.

To be used inside the Filter component for the new map viewer. 'Search fields' would only show for more than 12 values or so.

![image](https://user-images.githubusercontent.com/44481292/73391517-fca8d880-428c-11ea-920b-cdde9cc3cd45.png)

Currently it's this which is too big for this use case.
![image](https://user-images.githubusercontent.com/44481292/73391688-47c2eb80-428d-11ea-87ba-65ee0086f24f.png)

cc @asangma

asangma:
@AdelheidF Looks like if we use the new `scale="s"` prop on the Action, this should help. We'd still need to update pick-list-item to only render the `summary` node when it has content.

AdelheidF:
I removed the caclite-action here. Not sure where this padding comes from, but I don't see summary being renderered.

![image](https://user-images.githubusercontent.com/44481292/73477057-61753900-4348-11ea-8c8c-57e37ae5a493.png)

AdelheidF:
I'm using scale=s in the smartMapping Gallery pick-list here. This is how it looks. (left side)
But where I need a small pick-list is here in the layer filter component where I'm using my own list right now. (right side). 
If you imagine moving that pick-list into that space available there you'll only see 3.5 values. You have to scroll for the rest. 

![image](https://user-images.githubusercontent.com/44481292/73888459-870bb200-4822-11ea-8ed1-fe671d7f3253.png)![image](https://user-images.githubusercontent.com/44481292/73888527-a86c9e00-4822-11ea-9076-53f3e27c5993.png)

AdelheidF:
This is the current look.

![image](https://user-images.githubusercontent.com/44481292/75077040-71e78200-54b6-11ea-914a-30d065df1a77.png)

PR #821, opened 2020-02-13
feat: styles for a tighter calcite-filter and related styles in pick-list and value-list

asangma:
**Related Issue:** #795

## Summary
- slightly less padding and margin
- fixed bottom "border" UI bug
- added a `menu-action` in demo.


asangma:
cc @AdelheidF

Issue #818, opened 2020-02-11
Enhancement: Rename text props we provide a default for to have `intl-` prefix


driskull:
### Description

Enhancement: Rename text props we provide a default for to have `intl-` prefix

### Acceptance Criteria 

All properties that we provide default text in English be prefixed with `intl-` so that we can document the purpose of these properties and have them named consistently.

### Relevant Info 

Also consider having a bundle file where we can reuse some of the common English strings for the defaults.

#### Which Component

All

#### Example Use Case

- `text-close` => `intl-close`
- `text-back` => `intl-back`

### calcite-action-bar

- [x]  textExpand => intlExpand
- [x]  textCollapse => intlCollapse

### calcite-action-pad

- [x]  textExpand => intlExpand
- [x]  textCollapse => intlCollapse

### calcite-block

- [x]  textExpand => intlExpand
- [x]  textCollapse => intlCollapse

### calcite-block-section

- [x]  textExpand => intlExpand
- [x]  textCollapse => intlCollapse

### calcite-filter

- [x]  textLabel => label
- [x]  textPlaceholder => placeholder

### calcite-flow-item

- [x]  textBack => intlBack
- [x]  textClose => intlClose
- [x]  textOpen => intlOpen

### calcite-panel

- [x]  textClose => intlClose

### calcite-tip

- [x]  textClose => intlClose

### calcite-tip-manager

- [x]  textClose => intlClose
- [x]  textNext => intlNext
- [x]  textPrevious => intlPrevious

driskull:
@jcfranco @asangma do we want to do this as a breaking change or as deprecations?

jcfranco:
I vote to deprecate and remove later.

driskull:
## Props to change

### calcite-action-bar

- textExpand => localeExpand
- textCollapse => localeCollapse

### calcite-action-pad

- textExpand => localeExpand
- textCollapse => localeCollapse

### calcite-block

- textExpand => localeExpand
- textCollapse => localeCollapse

### calcite-block-section

- textExpand => localeExpand
- textCollapse => localeCollapse

### calcite-filter

- textLabel => label
- textPlaceholder => placeholder

### calcite-flow-item

- textBack => localeBack
- textClose => localeClose
- textOpen => localeOpen

### calcite-panel

- textClose => localeClose

### calcite-tip

- textClose => localeClose

### calcite-tip-manager

- textClose => localeClose
- textNext => localeNext
- textPrevious => localePrevious


[...]

PR #823, opened 2020-02-14
feat(action-pad,action-bar): Add intl string properties.

driskull:
**Related Issue:** #818

## Summary

Enhancement: Rename text props we provide a default for to have locale- prefix


driskull:
Renamed to `intl-`.

driskull:

Issue #841, opened 2020-02-20
Enhancement: calcite-flow-item needs an onscroll event 


AdelheidF:
### Description
My component needs to close a popover if the user scrolls the content of a flow-item.
The document.click event does not catch this event and neither does the onscroll event directly on the flow-item or my content div of the flow-item.

mitc7862:
@AdelheidF Would it close anytime the user scrolls or only when the position of the popover is off-screen?

driskull:
@AdelheidF is the Popover inside of the content that is scrolling? If so, it should be able to hide it.

Another option would be using these attributes that get added to the Popover...

- data-popper-reference-hidden: This attribute gets applied to the popper when the reference element is fully clipped and hidden from view, which causes the popper to appear to be attached to nothing.
- data-popper-escaped: This attribute gets applied when the popper escapes the reference element's boundary (and so it appears detached).

```scss
/* Hide the popper when the reference is hidden */
calcite-popover[data-popper-reference-hidden] {
  visibility: hidden;
  pointer-events: none;
}
```

AdelheidF:
> Would it close anytime the user scrolls or only when the position of the popover is off-screen?

> - Clicking outside will close the popover

It sounds like any action outside of the popover should close it. So that's what it will do.

AdelheidF:
> is the Popover inside of the content that is scrolling?

I would like to know when the user touches any point in the yellow area. The document.click event doesn't catch this.

<img src=https://user-images.githubusercontent.com/44481292/74963623-e4793480-53c6-11ea-9972-2ee54f4a3644.png width=250/>

AdelheidF:
Works well, thanks.

PR #842, opened 2020-02-20
feat(flow-item,panel): Add an onscroll event for FlowItem and Panel

driskull:
**Related Issue:** #841

## Summary

Enhancement: calcite-flow-item needs an onscroll event

- Update demos
- Add scroll events to panel and flow-item


driskull:
Added issue for test.

PR #843, opened 2020-02-21
feat(flow-item,panel): Add intl string properties.

driskull:
**Related Issue:** #818

## Summary

Enhancement: Rename text props we provide a default for to have `intl-` prefix #818


driskull:

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -70,17 +70,35 @@ export class CalciteFlowItem {
   /**
    * 'Back' text string.
    */
-  @Prop() textBack = TEXT.back;
+  @Prop() intlBack?: string;
```
jcfranco: Nitpick: can these be placed in alphabetical order?

driskull:
Yeah, lets note that they should have a value.

PR #844, opened 2020-02-21
feat(tip,tip-manager): Add intl string properties.

driskull:
**Related Issue:** #818

## Summary

Enhancement: Rename text props we provide a default for to have locale- prefix


src/components/intl/en-us.ts
```diff
@@ -0,0 +1,11 @@
+export default {
```
jcfranco: Is there a plan to support other bundles? If not, could we keep this as as common resource (`TEXT`)?

src/components/intl/en-us.ts
```diff
@@ -0,0 +1,11 @@
...
+  back: "Back",
+  close: "Close",
+  open: "Open",
+  expand: "Expand",
+  collapse: "Collapse",
+  next: "Next",
+  previous: "Previous",
+  tipGroupTitle: "Tips",
```
jcfranco: I'm wondering if there would be any benefits, build-wise, in having individual bundles vs a single one.

driskull: I think the benefit is not having the text "Close" defined a million times throughout component resources. But I see the issue if a user just wants to change one of them. Yes, we should research this more.

src/components/intl/en-us.ts
driskull: Yeah maybe we can name this file differently to let users know we don't plan on doing any localization ourselves.

driskull:

Issue #846, opened 2020-02-24
Refactor: Add styles to calcite-panel for 'start' and 'end'


kat10140:
### Description
Our `calcite-panel` styles are still linked to `leading` and `trailing` properties on the component. After our meeting with the calcite-components team, we've decided to go with `start` and `end` for our props to keep it consistent across the board.  We need to adjust our styles to fit.

Note: for now we are KEEPING leading and trailing until they are fully removed. 

Move Alan magick ovuh to make werk agin!

#### Which Component
calcite-panel

asangma:
Yeyer! Looks like the start and end styles are in there, and I'll need to add the deprecated leading and trailing (back) in.

kat10140:
🤙🏼

asangma:
@kat10140 MURGLED.

kat10140:
@asangma i see leading and trailing but no start end?

asangma:
VERFEED

PR #847, opened 2020-02-24
fix(shell-panel): style for deprecated leading and trailing attributes

asangma:
**Related Issue:** #846

## Summary
Added back styles for deprecated attributes.


asangma:
Thanks @driskull.

PR #852, opened 2020-02-27
feat(block, block-section, filter): add intl props

jcfranco:
**Related Issue:** #818 

## Summary

This adds `intl<name>` props to replace their applicable `text<name>` equivalents.

Also adds new `intlClose` property to filter since it had no text prop.

### DIFF
--- a/src/components/calcite-block-section/calcite-block-section.tsx
+++ b/src/components/calcite-block-section/calcite-block-section.tsx
@@ -22,6 +22,16 @@ export class CalciteBlockSection {
   // --------------------------------------------------------------------------
 
   /**
+   * Tooltip used for the toggle when expanded.
+   */
+  @Prop() intlCollapse = TEXT.collapse;
+
+  /**
+   * Tooltip used for the toggle when collapsed.
+   */
+  @Prop() intlExpand = TEXT.expand;
+
+  /**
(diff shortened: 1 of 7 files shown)


src/components/calcite-block-section/calcite-block-section.tsx
```diff
@@ -21,6 +21,16 @@ export class CalciteBlockSection {
   //
   // --------------------------------------------------------------------------
 
+  /**
+   * Tooltip used for the toggle when expanded.
+   */
+  @Prop() intlCollapse = TEXT.collapse;
```
driskull: For the other PRs, I added the `TEXT.collapse` as a fallback in the render so even if they set `intlCollapse` to an empty string, there would be some text there. It also makes this prop optional by adding `?` Do we want to follow that approach?

jcfranco: I've updated the PR to be consistent with your changes. 

Sidebar, I thought marking props as optional/required props had an effect in typings or the doc, but I don't see that. Do you see see differently?

jcfranco: Actually, [this PR](https://github.com/Esri/calcite-app-components/pull/853) has the changes I mentioned. I somehow omitted them here.

PR #854, opened 2020-02-27
feat(filter): add placeholder property

jcfranco:
**Related Issue:** N/A

## Summary

This adds `placeholder` to `calcite-filter`. This prop deprecates `textPlaceholder` since it does not have a default value and therefore does not need the `text` prefix.

### DIFF
--- a/src/components/calcite-filter/calcite-filter.tsx
+++ b/src/components/calcite-filter/calcite-filter.tsx
@@ -136,7 +136,7 @@ export class CalciteFilter {
           <input
             type="text"
             value=""
-            placeholder={this.textPlaceholder || this.placeholder}
+            placeholder={this.placeholder || this.textPlaceholder}
             onInput={this.inputHandler}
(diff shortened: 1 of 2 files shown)


jcfranco:
lol forgot to push. #notanightowl

src/components/calcite-filter/calcite-filter.tsx
```diff
@@ -129,7 +136,7 @@ export class CalciteFilter {
           <input
             type="text"
             value=""
-            placeholder={this.textPlaceholder}
+            placeholder={this.textPlaceholder || this.placeholder}
```
driskull: shouldn't `placeholder` take precedence?

jcfranco: Ah, yes. Good catch. Will fix.

jcfranco: 💥

Issue #857, opened 2020-02-27
Bug: ActionBar in ShellPanel - collapse icon is wrong on the trailing/end side


asangma:
### Summary
<img width="298" alt="Screen Shot 2020-02-27 at 1 19 22 PM" src="https://user-images.githubusercontent.com/12503298/75487739-d8134f80-5963-11ea-826c-1e718b6afb94.png">

### Reproduction Steps

1. See `/story/components-calcite-shell--basic`

### Relevant Info 

@kat10140 @driskull @jcfranco Would be good get this in and release soon.

asangma:
@kat10140 Mind verfeeing?

kat10140:
VERF!

PR #858, opened 2020-02-27
fix(action-bar): removed default return from getCalcitePosition

asangma:
**Related Issue:** #857

## Summary
Removed default return value from `getCalcitePosition`.


asangma:
@driskull Updated.

Issue #862, opened 2020-02-28
Bug: border width not consistent witih calcite-block-section toggle mode


kat10140:
### Description
Per Bjorn: 

Minor, but looks a little sloppy:
![cd19e880-597e-11ea-8ecb-b16e3d8341e3](https://user-images.githubusercontent.com/48069902/75579578-03a54100-5a1b-11ea-9fdd-4ca133f2bc92.png)

### Repro Steps

* https://qa-testing.mapsqa.arcgis.com/apps/mapviewer/index.html?webmap=bc20b4b7c7724365b09addf472d208d2
* click a layer
* Click Style options

### Relevant Info (e.g. Browser, OS, Version, Regression)

* Only tested in Chrome Canary

Original issue:
https://devtopia.esri.com/WebGIS/arcgis-webviewer-app/issues/1319

jcfranco:
Installed. @kat10140 plz verrrfeeee. 🙇 🎳 :bowtie:

kat10140:
Verified!

PR #863, opened 2020-02-28
fix(block-section): increased top and bottom margins on block-section label to fix focus ring overlap

asangma:
**Related Issue:** #862

## Summary
Slight margin tweak for focus state.

cc @bsvensson


jcfranco:
> fix(block-section): margin tweaks for focus state

For the changelog, could you add a little more detail to the commit message to convey what it fixes?

asangma:
@jcfranco how about
> fix(block-section): increased top and bottom margins on block-section label to fix focus ring overlap

jcfranco:
☝️ That's excellent!

PR #867, opened 2020-03-02
fix(flow-item): update to only show fab-container when fab slot is used

asangma:
**Related Issue:** See 1322 in webviewer repo.

## Summary
- Added a check for the `fab` slot in flow-item
- `fab-container` will only show if the slot is used.


asangma:
@driskull #iswydt

PR #881, opened 2020-03-17
build(lint): set up ESLint

jcfranco:
**Related Issue:** N/A

## Summary

This sets up ESLint as our main linter. 

- All linting errors/warnings found by this switch have been fixed.
- TSLint is being kept to for rules that don't have an equivalent (these are all warnings)
  - [`component-member-order`](https://github.com/natemoo-re/tslint-stencil/blob/master/docs/component-member-order.md) (`tslint-stencil`)
  - [`lifecycle-order`](https://github.com/natemoo-re/tslint-stencil/blob/master/docs/lifecycle-order.md) (`tslint-stencil`)
  - [`prefer-conditional-expression`](https://palantir.github.io/tslint/rules/prefer-conditional-expression/) (`tstlint`)


jcfranco:
It does enforce types on everything, which I like, but I do see your point. I'll play with the config after merging to see if it can warn only when __no__ type can be inferred.

Issue #873, opened 2020-03-04
Enhancement: calcite-pick-list and calcite-value-list with down/up arrow support


jenn0003:
### Description
Currently on dev: String + "is" returns strings with a radio button. User does not see the button until they click on an option. 

Once clicked, the radio button appears. If the user clicks the down arrow, the scroll bar moves down. The expectation, and currently on production, is that the down arrow moves the radio selection to the next string.

Currently on dev, the user clicks the first string, then clicks TAB to go to the next string then SPACE BAR to select it. 

It would also be nice to auto scroll the main filter scroll to the bottom of each expression. For example, add one expression then add a second expression. Select a string field, then IS, and now a longer selection of strings appear. At this moment, it would be convenient for the user if the filter scroll automatically scrolled down to the bottom so that the list and the "add expression" button is shown.
### Acceptance Criteria 
1. Radio selection is activated by down arrow for Filter "IS" and "IS NOT"
2. Button appears for all strings. Right now it is all blank so there is no indication that that is an option.
3. Auto-scroll down when multiple expressions are added.

### Relevant Info 
![image](https://user-images.githubusercontent.com/10080810/75923303-a09c1b80-5e19-11ea-884d-07e1ea5ffe45.png)

#### Which Component
Filter

#### Example Use Case
Any layer with domains set. Here is a [public schools layer](https://urbanobservatory.maps.arcgis.com/home/item.html?id=22b7807fc3dc4d089020ba269f7dc766&view=table#data) that has the State field domains that works well with the "is" and "is not" filter. Will need to bring into dev to test the above scenario.

AdelheidF:
created a separate issue for the scrolling: https://devtopia.esri.com/WebGIS/arcgis-components/issues/650

driskull:
@jcfranco is taking this.

jcfranco:
Installed. @jenn0003 @AdelheidF Could you help verify this when released?

jenn0003:
@jcfranco sure thing.  To clarify, we'll test the up and down arrow on the filter after the release?

jcfranco:
Yes to the testing part. Not sure if this is targeted for beta 5 or after.

jcfranco:
@jenn0003 This is on devext now. Would you mind testing it out? Thanks!

jenn0003:
The up/down arrow option for filtering works well on dev. Looks great!

jcfranco:
Awesome, thanks for verifying @jenn0003!

PR #887, opened 2020-03-20
feat(value-list, pick-list): support ↑↓ arrow keys for navigation and value selection when multiple mode is disabled

driskull:
**Related Issue:** #873

## Summary

This updates the `pick/value-list` components to support item navigation with the up/down arrow keys. When `multiple` is disabled, moving an item will also affect item selection (akin to radio groups).

### Notable changes

* Added `setFocus` to lists and items.
* The wrapped `"calciteListItemChange"` event used by the `value-list` is modified before it continues bubbling to adjust the item instance accordingly since it was always a `pick-list-item` instance.


driskull:
@jcfranco this is working but needs the following still:

- Tests
- Only the selected item should be able to be tabbed to when `multiple` is disabled.
  - If no item is selected, the first item should be focusable.
  - This is how native radio elements work. Only the selected radio can be focused and using the arrow keys goes to the next item and selects it.

jcfranco:
@driskull Can you review? 🙇

jcfranco:
> the only concern I have is that maybe the control slot should receive focus for pick list?
> For example, I can't keyboard to the "X" icon without selecting that item first.

Could you walk me through that? When you tab into a list with removable items, you'd want the X to be focused and then ↑	↓ would move between other item's Xs?

driskull:
> When you tab into a list with removable items, you'd want the X to be focused and then ↑ ↓ would move between other item's Xs?

Yeah, I'm not sure what would be ideal. Maybe we could proceed with how you have it and then we can consult with Arjav?

jcfranco:
Sounds good. 👌

jcfranco:
Approving on behalf of @driskull. 🤜:boom: 🤛

PR #890, opened 2020-03-25
fix(action): Add slotChildNodesFix to stencil config extras

driskull:
**Related Issue:** None

## Summary

```ts
/**
     * For browsers that do not support shadow dom (IE11 and Edge 18 and below), slot is polyfilled
     * to simulate the same behavior. However, the host element's `childNodes` and `children`
     * getters are not patched to only show the child nodes and elements of the default slot.
     * Defaults to `false`.
     */
    slotChildNodesFix?: boolean;
```

@jcfranco do you remember which component we were trying to find out if the default slot is populated?

### DIFF
--- a/src/components/calcite-action/calcite-action.tsx
+++ b/src/components/calcite-action/calcite-action.tsx
@@ -130,7 +130,7 @@ export class CalciteAction {
     const calciteLoaderNode = loading ? <calcite-loader is-active inline /> : null;
     const calciteIconNode = icon ? <calcite-icon icon={icon} scale={iconScale} /> : null;
     const iconNode = calciteLoaderNode || calciteIconNode;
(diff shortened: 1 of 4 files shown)


driskull:
I think it's the action. I updated PR.

jcfranco:
> do you remember which component we were trying to find out if the default slot is populated?

I know block was one of them.

driskull:
@macandcheese @paulcpederson you may want to turn on this extra if you need to figure out if a default slot has content inside it.

driskull:
@jcfranco do you want to update the block in this PR or a different one?

jcfranco:
It's no longer relevant for the block (see https://github.com/Esri/calcite-app-components/pull/793#issuecomment-580369056).

PR #891, opened 2020-03-25
chore: Bump versions

driskull:
chore(bump): Bump versions


.gitignore
```diff
@@ -20,6 +20,7 @@ __docs-temp__/*
 *.bak
 *.orig
 .stencil/
+.storybook/
```
jcfranco: This shouldn't be ignored.

driskull: @jcfranco should we check in these files then?

Issue #849, opened 2020-02-25
Bug: [Block] control slot layout issues


jcfranco:
### Summary

The control slot is positioned absolutely to avoid having it in the button element (prevent it from receiving keyboard mouse events), leading to odd layout when the block width is smaller than the control.

<img width="213" alt="Screen Shot 2020-02-25 at 2 33 31 PM" src="https://user-images.githubusercontent.com/197440/75293848-ecc4db80-57db-11ea-8c76-a925df17fe82.png">

https://github.com/Esri/calcite-app-components/pull/762 introduced this change (mainly to address [this comment](https://github.com/Esri/calcite-app-components/pull/752#discussion_r366544260)).

### Actual Behavior

The slot content is not part of the block layout.

### Expected Behavior

!:point_up:

### Reproduction Steps

This can be seen in https://esri.github.io/calcite-app-components/?path=/story/components-calcite-block--with-header-control.

asangma:
Block likes to use its container for its width. I think the example in the storybook creates a case that is unlikely in the wild. Setting width on the container shows what would more likely happen in an app.
<img width="811" alt="Screen Shot 2020-03-19 at 9 52 32 AM" src="https://user-images.githubusercontent.com/12503298/77093043-ad973e00-69c7-11ea-8296-28f86594338b.png">

As such I'm reluctant to do something like put a min-width on Block, but let me know if you think that's preferable.

jcfranco:
Agreed 💯%!

In that case, should we just update the story to make it more realistic?

asangma:
Ya.

jcfranco:
Installed.

jcfranco:
@kat10140 Can you help verify this one?

kat10140:
verified!

PR #895, opened 2020-03-26
fix(block): use CSS grid to improve header layout

jcfranco:
**Related Issue:** #849

## Summary

This updates Block to use CSS grid for its layout to avoid an issue where a control may flow outside its container.

### DIFF
--- a/src/components/calcite-block/calcite-block.scss
+++ b/src/components/calcite-block/calcite-block.scss
@@ -17,8 +17,8 @@
 }
 
 .header-container {
-  position: relative;
-  display: flex;
+  display: grid;
+  grid: auto auto / auto minmax(0, 1fr);
   align-items: center;
 
   & > .header {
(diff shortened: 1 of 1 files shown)


src/components/calcite-block/calcite-block.scss
```diff
@@ -17,8 +17,9 @@
 }
 
 .header-container {
-  position: relative;
-  display: flex;
+  display: grid;
+  grid-template: auto / auto auto 1fr;
```
driskull: This won't work in IE11, but I think that's ok :)

jcfranco: There's partial support for it, but I haven't tested at all on IE11.

jcfranco:
> as long as @asangma approves :)

Whoops, I got trigger happy. 😄 @asangma can you review this when you get a chance?

Issue #896, opened 2020-03-27
Enhancement: [Block] Drop open/close icon


jcfranco:
### Description

The open/close icon is no longer needed in the updated block design.

### Acceptance Criteria 

No more caret! 🥕 

### Relevant Info 

N/A

#### Which Component

Block

#### Example Use Case

N/A

cc @asangma

jcfranco:
Installed.

kat10140:
verified!

PR #897, opened 2020-03-27
feat(block): drop open/close icon

jcfranco:
**Related Issue:** #896 

## Summary

Removes block header caret per updated block design. 👋🥕

### DIFF
--- a/src/demos/block/basic.html
+++ b/src/demos/block/basic.html
@@ -71,8 +71,7 @@
 
         <h2>With drag handle</h2>
 
-        <calcite-block heading="Heading" summary="summary" drag-handle="true">          <label slot="control">test <input placeholder="I'm a header control"/></label>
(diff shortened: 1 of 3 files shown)


src/demos/block/basic.html
```diff
@@ -78,7 +71,8 @@ <h2>Header + content (sections w/ switch)</h2>
 
         <h2>With drag handle</h2>
 
-        <calcite-block heading="Heading" summary="summary" drag-handle="true"></calcite-block>
+        <calcite-block heading="Heading" summary="summary" drag-handle="true">          <label slot="control">test <input placeholder="I'm a header control"/></label>
```
driskull: nitpick: can we remove the weird spacing here?

jcfranco: Even better, I'll remove it!

PR #898, opened 2020-03-30
docs(fab): fix storybook in edge

driskull:
**Related Issue:** None

## Summary

docs(fab): fix storybook in edge


driskull:
calcite-fab was showing up twice in Edge. This removes the duplicate.

PR #904, opened 2020-04-01
feat(action-bar,action-pad): Add tooltipExpand property

driskull:
**Related Issue:** None

## Summary

- Add a `expandTooltip` property to `calcite-action-bar` and `calcite-action-pad` to show a `HTMLCalciteTooltipElement` when the action is collapsed.


src/components/calcite-action-bar/calcite-action-bar.tsx
```diff
@@ -55,6 +55,11 @@ export class CalciteActionBar {
     this.calciteActionBarToggle.emit();
   }
 
+  /**
+   * Used to set the a tooltip for the expand toggle.
```
jcfranco, line 59: Nitpick: ` set the a tooltip` -> ` set a tooltip`

driskull: 🤦‍♂

driskull:
> Does this mean we have to add a similar prop for any component setting the title attr?

@jcfranco I don't think we have to for every component but I do see the map viewer team asking for more options to include tooltips for shadowed elements so we should have a plan for this incase.

> If this is a pattern we'll be applying across components, should we consider using tooltip as the prefix? This would be consistent w/ intl props.

Yeah we should consider it. Would `tooltipExpand` be ok?

driskull:
I'll work on a couple basic tests and stories.

driskull:
@jcfranco added usage file but can't figure out how to add story with a prop that accepts a dom node.

PR #913, opened 2020-04-02
fix(action): Support slotting icon after initialization

driskull:
**Related Issue:** None

## Summary

fix(action): Support slotting icon after initialization

### DIFF
--- a/src/components/calcite-action/calcite-action.tsx
+++ b/src/components/calcite-action/calcite-action.tsx
@@ -9,7 +9,7 @@ import { CSS } from "./resources";
 import { CSS_UTILITY } from "../utils/resources";
 
 import { getElementDir } from "../utils/dom";
-import { VNode } from "@stencil/core/internal";
+import { VNode, State } from "@stencil/core/internal";
 
 /**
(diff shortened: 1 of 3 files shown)


src/components/calcite-action/calcite-action.tsx
```diff
@@ -91,7 +91,29 @@ export class CalciteAction {
 
   @Element() el: HTMLCalciteActionElement;
 
-  private buttonEl: HTMLButtonElement;
+  @State() hasChildren = false;
+
+  buttonEl: HTMLButtonElement;
+
+  observer = new MutationObserver(() => this.setHasChildren());
```
jcfranco, line 98: Would using `forceUpdate` after a mutation change help to simplify?

driskull: I thought about that but wasn't sure if it was as efficient as using a state variable. Do you know?

jcfranco: Haven't really played with it, but I'd think it's similar. Every time there's a mutation change you update state to rerender. Using `forceUpdate` seems like it would skip the setting state part. Worth testing.

PR #926, opened 2020-04-14
docs(action,fab): fix edge legacy stories.

driskull:
**Related Issue:** None

## Summary
docs: fix edge legacy stories.

removes the `undefined` text from showing up in edge legacy stories.


driskull:
@jcfranco If I don't add it then `undefined` shows up inside the component.

PR #927, opened 2020-04-14
fix(action): Remove text flickering in Edge legacy.

driskull:
**Related Issue:** None

## Summary

fix(action): Remove text flickering in Edge legacy.


driskull:
ForceUpdate would cause the text to flicker on and off. Seems like forceUpdate doesn't do an efficient render I'm guessing.

PR #929, opened 2020-04-14
fix(flow-item): Stack error in Edge legacy

driskull:
**Related Issue:** None

## Summary

fix(flow-item): Stack error in Edge legacy


src/components/calcite-flow-item/resources.ts
```diff
@@ -12,7 +12,8 @@ export const CSS = {
   menuButton: "menu-button",
   menu: "menu",
   menuOpen: "menu--open",
-  fabContainer: "fab-container"
+  fabContainer: "fab-container",
+  defaultSlotContainer: "default-slot-container"
```
jcfranco, line 16: It doesn't seem like this class is being used.

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -368,12 +372,20 @@ export class CalciteFlowItem {
...
-      <div class={CSS.fabContainer} slot={PANEL_SLOTS.fab}>
+      <div key="fab-container" class={CSS.fabContainer} slot={PANEL_SLOTS.fab}>
         <slot name={SLOTS.fab} />
       </div>
     ) : null;
   }
 
+  renderDefaultSlot(): VNode {
+    return (
+      <div key="default-slot-container" class={CSS.defaultSlotContainer}>
```
jcfranco, line 383: Just curious if `key` is needed here since this node is not rendered conditionally.

jcfranco:
Was this only an issue in Edge legacy? Seems like this would've caused rendering trouble in other browsers because of the missing `key`s. Maybe it isn't common for a reused component to change part-toggling props.

driskull:
@jcfranco I can remove the keys, its the div around the slot that fixed it.

src/components/calcite-flow-item/resources.ts
driskull: Yeah but Its better than a div without anything.

driskull:
This might be another issue for Stencil. Putting a default slot inside another nested component could be the issue but would need to investigate more.

src/components/calcite-flow-item/resources.ts
jcfranco: Strong disagree. Let's not add anything that isn't being used.

src/components/calcite-flow-item/calcite-flow-item.tsx
```diff
@@ -374,6 +374,15 @@ export class CalciteFlowItem {
     ) : null;
   }
 
+  renderDefaultSlot(): VNode {
+    return (
+      // default slot requires a container for edge legacy.
```
jcfranco, line 379: Nitpick: could you add more context about this being a workaround?

driskull:
If I set `slotChildNodesFix: false` then this is no longer necessary. I'll open a Stencil issue instead of merging this right now.

driskull:
closing for now. Will reopen if necessary.

Issue #924, opened 2020-04-10
Docs: document supported browsers


bsvensson:
### Description

I couldn't find any info about supported browsers. It would be nice to have that clearly documented like https://github.com/Esri/calcite-components/#browser-support

### Which Component

README.md

bsvensson:
Thanks @driskull 

https://github.com/Esri/calcite-app-components#browser-support looks good, but what about IE11? If it doesn't support it, should it say so?  The current write-up leaves me wondering... :)

driskull:
@bsvensson should list every browser or just go with if its not on the list then its not supported?

I think its ok that if its not on the list then its not supported. Do we really need to list Opera, opera mini? etc.

bsvensson:
Fine with me :)

Closing.

PR #931, opened 2020-04-14
docs: Add text for browser support.

driskull:
**Related Issue:** #924

## Summary

@jcfranco  should we be more specific? like....

> Browsers that are not listed below may work as expected but are not officially supported.


jcfranco:
Let's leave it as is. IMO, it's pretty clear.

Issue #940, opened 2020-04-23
Enhancement: Storybook and demo fix for picklist and valuelist


asangma:
### Description
In PickList and ValueList demos, the circle icon is often mistaken for a radio button.

### Acceptance Criteria 
Kill the radio star.

#### Which Component
Pick and Value List demos.

asangma:
![Screen Shot 2020-04-23 at 4 23 22 PM](https://user-images.githubusercontent.com/12503298/80159197-3c771780-857f-11ea-9f65-dd7cf45e3fca.png)

asangma:
![Screen Shot 2020-04-23 at 4 27 47 PM](https://user-images.githubusercontent.com/12503298/80159296-7b0cd200-857f-11ea-98f4-3d612b888a08.png)
![Screen Shot 2020-04-23 at 4 28 07 PM](https://user-images.githubusercontent.com/12503298/80159304-7ea05900-857f-11ea-9f04-b80d26380bbc.png)

asangma:
But maybe not banana.

kat10140:
lewks gewd

asangma:
Thanks @kat10140 !

PR #943, opened 2020-04-23
feat: use less confusing icons in pick-list and value-list

asangma:
**Related Issue:** #940

## Summary
Icons in the component demos and storybooks updated to look less like radio buttons.
Also cleared out inline styling.


asangma:
Thanks mah doges. Will MURGHLE after Travis is happy.

Issue #945, opened 2020-04-27
Enhancement: Implement focus mixin to match calcite-components


asangma:
### Description
calcite-components has standardized :focus styles and animations. Update relevant styles to match.

### Acceptance Criteria 
Aligned styles.

#### Which Component
Probably all.

asangma:
I believe the PR handled #866 as well.

kat10140:
![Screen Shot 2020-04-28 at 4 08 50 PM](https://user-images.githubusercontent.com/48069902/80546547-c69df200-896a-11ea-9ba1-1759eefadf52.png)
Tip manager is the only one that looks kinda weird
@asangma

asangma:
Cool. I'll have a look.

kat10140:
yew!

PR #946, opened 2020-04-28
feat: replace focus ring mixin and added new mixin where needed

asangma:
**Related Issue:** #945 

## Summary
* Add new :focus mixin to match calcite-components.
* Replace old mixin with new
* Added new mixin where needed

cc @macandcheese let me know if there's a way to pull those mixins in from calcite-components. I'm currently reproducing them.

### DIFF
--- a/src/components/calcite-action/calcite-action.scss
+++ b/src/components/calcite-action/calcite-action.scss
@@ -24,7 +24,7 @@
   transition: color 125ms ease-in-out, fill 125ms ease-in-out, background-color 125ms ease-in-out;
   text-align: unset;
   position: relative;
-
+  @include focus-style-base();
   &:hover,
   &:focus {
     background-color: var(--calcite-app-background-hover);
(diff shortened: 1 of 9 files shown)


macandcheese:
They'll eventually be added to 'calcite-base' - we may also add a mixin that uses box shadow to allow for rounded items to have a consistent focus ring (the close button portion of a chip, rounded button, etc.)

asangma:
Thanks @macandcheese. Once it's in base, I'll refactor.

@driskull I can test, but it should be all right since we're styling the standard `outline`. @macandcheese Did y'all test at all in Edge legacy?

asangma:
Also, I don't have access to be able to test at the moment.

asangma:
@driskull Spoke with Adam, and it sounds like they haven't seen any issues. Would you prefer that we test this before moving forward?

driskull:
We can move forward I think. We should test at some point though.

asangma:
Sounds good. MURGLIN'.

Issue #952, opened 2020-05-01
Enhancement: FAB sticky in panel container with h-scroll


asangma:
### Description
When using the `fab` slot in Panel and there is horizontally scrolling content, the fab gets scrolled away. 

We'll probs see this in the new CalciteShellCenterRow component.

BAD
![fab-issue](https://user-images.githubusercontent.com/12503298/80844000-803ed200-8bba-11ea-86dd-457acbc46bc4.gif)

GOOD
![fab-good](https://user-images.githubusercontent.com/12503298/80844015-892fa380-8bba-11ea-8145-c70b2a6d1874.gif)

#### Which Component
Panel

cc @kat10140

asangma:
@kat10140 Pease see the `shell/demo-app-advanced-center-row.html` demo to test this.

kat10140:
yay! vurf

PR #953, opened 2020-05-01
feat(panel): tweak for FAB slot when h-scroll

asangma:
**Related Issue:** #952

## Summary
Little tweak to handle when FAB is in a Panel inside ShellCenterRow and stuff goes sideways. Sideways like scrolling horizontally.

cc @kat10140


asangma:
Thanks @driskull. Will MURG after Travis is done.

Issue #957, opened 2020-05-06
Bug: calcite-block-section - CSS issue with RTL


AdelheidF:
### Summary
When in RTL mode there is padding on the left of the toggle button that is not there in LTR. This means my toggle buttons don't align nicely.

LTR:
<img src=https://user-images.githubusercontent.com/44481292/81136290-02f3c400-8f10-11ea-978b-7bf46629d8c3.png width=250/>

RTL:
<img src=https://user-images.githubusercontent.com/44481292/81136321-16069400-8f10-11ea-9d7b-cee06af6a2c2.png width=250/>

cc @asangma 

### Reproduction Steps

1.

asangma:
👀

kat10140:
verf!

PR #962, opened 2020-05-06
fix(block-section): added rtl check and css

asangma:
**Related Issue:** #957

## Summary
Add `calcite--rtl` to BlockSection.
This CSS class on Block wasn't affecting BlockSection.

### DIFF
--- a/src/demos/block/basic.html
+++ b/src/demos/block/basic.html
@@ -91,7 +91,7 @@
             </calcite-block-section>
           </calcite-block>
 
-          <h3>With sections</h3>
+          <h3>Header + content (sections w/ switch)</h3>
           
           <calcite-block heading="Heading" summary="summary" open collapsible>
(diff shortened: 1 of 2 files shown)


asangma:
cc @AdelheidF

src/demos/block/basic.html
```diff
@@ -90,6 +90,18 @@ <h3>Header + content (collapsible)</h3>
               <img alt="demo" src="https://placeimg.com/640/480/nature" />
             </calcite-block-section>
           </calcite-block>
+
+          <h3>With sections</h3>
```
jcfranco, line 94: Should this be  `<h2>Header + content (sections w/ switch)</h2>`?

asangma: oh yeah.

asangma: Updated.

PR #963, opened 2020-05-06
docs: Add a11y to storybook

driskull:
**Related Issue:** None

## Summary

Add a11y addon to storybook to easily test

image

We would need to import it to all stories and add our aXe configuration.

https://github.com/storybookjs/storybook/tree/master/addons/a11y#parameters

What are your thoughts @jcfranco ?

https://github.com/storybookjs/storybook/tree/master/addons/a11y

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,5 +1,6 @@
 module.exports = {
   addons: [
+    "@storybook/addon-a11y/register",
     "@storybook/addon-backgrounds/register",
     "@storybook/addon-knobs/register",
     "@storybook/addon-notes/register-panel"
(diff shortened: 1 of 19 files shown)


driskull:
Kind of nice because i'm finding a few things easily...

image

cc @jcfranco

jcfranco:
IMO, think this should be part of testing and not in stories.

driskull:
@jcfranco I moved the `a11y` tab to be the last in the order, would that be ok?

I think it is useful to be able to test all the different knobs/properties and ensure they are being accessible. I was able to find a few things because of this. Creating test apps to to do this is a bit more work.

jcfranco:
👌 Per our meeting, we'll add the a11y tab in an effort to make it easy to find things the e2e tests may be missing.

jcfranco:
> and add our aXe configuration.

@driskull We use the default configuration and it seems like the addon does too. Can you confirm?

driskull:
@jcfranco yes it uses the default aXe configuration.

PR #968, opened 2020-05-08
fix(tip-manager): changed focus style to avoid hidden ring

asangma:
**Related Issue:** #945 

## Summary
Updated the focus ring to be outset.

cc @kat10140


asangma:
Thanks @driskull

PR #990, opened 2020-05-20
chore(standard-version): use changelogHeader replacement

jcfranco:
**Related Issue:** N/A

## Summary

This addresses the following warning:

```
[standard-version]: --changelogHeader will be removed in the next major release. Use --header.
```


jcfranco:
I'll update the header separately.

PR #1004, opened 2020-05-29
fix(action): Prevent disabled buttons from being clicked or tabbed into

driskull:
**Related Issue:** None

## Summary

fix(action): Prevent disabled buttons from being clicked or tabbed into

### DIFF
--- a/src/components/calcite-action/calcite-action.scss
+++ b/src/components/calcite-action/calcite-action.scss
@@ -4,6 +4,10 @@
   background-color: var(--calcite-app-background-clear);
 }
 
+:host([disabled]) {
+  pointer-events: none;
+}
+
 @import "../../scss/includes/_component";
 
 .button {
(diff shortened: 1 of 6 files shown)


asangma:
Shouldn't browsers default to this behavior with buttons that have a `disabled` attribute?

driskull:
> Shouldn't browsers default to this behavior with buttons elements that have a disabled attribute?

Yes, but when users set an onClick on the element it isn't set on the `button`. So this is needed.

asangma:
Ah...got it.

src/components/calcite-action/calcite-action.tsx
```diff
@@ -182,6 +182,7 @@ export class CalciteAction {
     return (
       <Host>
         <button
+          tabIndex={disabled ? -1 : null}
```
jcfranco, line 185: `disabled` should take care of the tab index, no?

src/components/calcite-action/calcite-action.scss
```diff
@@ -4,6 +4,10 @@
   background-color: var(--calcite-app-background-clear);
 }
 
+:host([disabled]) {
```
jcfranco, line 7: Can you add a test for this?

Issue #986, opened 2020-05-19
Bug: value-list focus


kat10140:
### Summary
Drag and dropping is acting really finicky with keyboard navigation

### Actual Behavior
Issue # 1 - activating the drag handle and moving the item down and back up to it's original place will then deactivate the item. 
1 - 2 on a similar note, activating the drag and taking the item to the bottom of the list takes the item back to the top and deactivates the drag handle.

![value list bug1](https://user-images.githubusercontent.com/48069902/83292676-5f5ca300-a19f-11ea-9a7b-b3ed468f4e0d.gif)

![value list bug 1-2](https://user-images.githubusercontent.com/48069902/83293039-fe819a80-a19f-11ea-8d13-c1e8b0daa60a.gif)

Issue # 2 - activating the drag handle after selection of an item moves down the item one slot, then the item's drag handle is deactivated and the item is deselected.

![value list bug2](https://user-images.githubusercontent.com/48069902/83292748-7bf8db00-a19f-11ea-98d9-3cff406a3e16.gif)

### Expected Behavior
tabbing and focus to work properly

### Reproduction Steps

1. start up demos in chrome, select drag handle and move down and back. 
2. refresh. select the item, then activate the drag handle and move down once to notice the deselection

jcfranco:
@kat10140 I can't repro with the provided steps. Can you add more info?

asangma:
I'm also seeing this in Safari.
@jcfranco Note, it mostly seems to be in the drag enabled value-list.

jcfranco:
Here's what I've noticed based on my understanding of how the component should work:

### Chrome (81.0.4044.138)

#### Mouse

Immediately dragging an item via the handle works (focus stays on the active item) ✅

#### Keyboard

Tabbing into an item gets focused and selected

  1. Pressing tab focuses the handle ✅
      a. Pressing Space enables drag n' drop mode
      b. Pressing ↕ arrows move item accordingly
      c. After item is moved drag n' drop mode stops ❌
  2. Pressing Tab again focuses the item content ✅, pressing it once more focuses the list's sibling element, if any ✅(reverse order works too)
  3. Pressing ↕ arrows move selection accordingly ✅

### Safari (13.1)

#### Mouse

Immediately dragging an item up or down works ONLY if the handle is focused first (focus stays on the active item) ❌

#### Keyboard

Same as Chrome, except for 1b. where drag n' drop mode stays enabled until Space is pressed again ✅

### Other deets

[...]

PR #1012, opened 2020-06-05
fix(value-list): ensure keyboard-enabled DND stays active when moving items around

jcfranco:
**Related Issue:** #986 

## Summary

This ensures the handle element is focused right after browser rendering.

Additionally, this also fixes some ESLint errors and adds more assertions to test keyboard item-moving.

### DIFF
--- a/src/components/calcite-value-list-item/calcite-value-list-item.tsx
+++ b/src/components/calcite-value-list-item/calcite-value-list-item.tsx
@@ -42,7 +42,7 @@ export class CalciteValueListItem {
   /**
    * Used to provide additional metadata to an item, primarily used when the parent list has a filter.
    */
-  @Prop() metadata?: object;
+  @Prop() metadata?: Record<string, unknown>;
 
   /**
    * Set this to true to display a remove action that removes the item from the list.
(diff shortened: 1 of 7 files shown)


src/components/calcite-value-list-item/calcite-value-list-item.tsx
```diff
@@ -115,7 +115,7 @@ export class CalciteValueListItem {
   //
   // --------------------------------------------------------------------------
 
-  getPickListRef = (el): HTMLCalcitePickListItemElement => (this.pickListItem = el);
+  getPickListRef = (el: HTMLCalcitePickListItemElement) => (this.pickListItem = el);
```
driskull, line 118: Did this change the return type?

src/components/calcite-value-list/calcite-value-list.tsx
```diff
@@ -179,28 +178,23 @@ export class CalciteValueList<
     if (!this.dragEnabled) {
       return;
     }
-    this.sortables.push(
-      Sortable.create(this.el, {
-        group: this.guid,
```
driskull, line 184: nice. yeah this wasn't necessary.

src/components/calcite-value-list-item/calcite-value-list-item.tsx
jcfranco: It should be able to infer the type, but I'll revert this change since linting still passes (I could've sworn it was failing earlier).

Issue #1031, opened 2020-07-10
Bug: calcite-shell-center-row, if no actionBar specified get an error


kevindoshier:
### Summary
calcite-shell-center-row was updated to allow an actionBar to be added, awesome, but if I dont use an action bar this line throws an error

https://github.com/Esri/calcite-app-components/blob/master/src/components/calcite-shell-center-row/calcite-shell-center-row.tsx#L72

seen with 6.1.0

### Expected Behavior

Allow calcite-center-row without an actionBar

### Reproduction Steps

1.

kevindoshier:
Thanks, looks good 👍

PR #1032, opened 2020-07-10
fix(shell-center-row): Error when no action-bar is provided

driskull:
**Related Issue:** #1031

## Summary

fix(shell-center-row): Error when no action-bar is provided


driskull:
@jcfranco can you review the test? I'm not sure if this is the best way to detect the error but I could not find another solution. 

If we do want to go with this, it seems like a good candidate to be something we test for automatically for each component.

src/components/calcite-shell-center-row/calcite-shell-center-row.e2e.ts
```diff
@@ -24,6 +24,19 @@ describe("calcite-shell-center-row", () => {
       }
     ]));
 
+  it("should not error when there is no action-bar", async () => {
```
driskull, line 27: @jcfranco could we make this test reusable like the `renders` test? I don't think any components should throw a console.error.

jcfranco: Looks like there's already a way to fail on console error:

```ts
it("should not error when there is no action-bar", async () =>
  await newE2EPage({
    html: "<calcite-shell-center-row></calcite-shell-center-row>",
    failOnConsoleError: true
  }));
```

driskull: Oh nice. Ill test it out. Can we turn it on by default?

jcfranco: Hmm... I don't see a way to enable it globally.

driskull: Can we maybe add this to the renders test?

Issue #1044, opened 2020-07-23
slot-container height of 0


AdelheidF:
### Summary

Not sure what component this goes to. The table no longer shows in the center.
Happens with 6.1.1, did not happen with 6.1.0.

http://devext.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/4yjifSiIG17X0gW4/arcgis/rest/services/FatalAccidents2017/FeatureServer/0
Click 'Show table' from layer menu.

<img src=https://user-images.githubusercontent.com/44481292/88345479-dc694c00-ccfa-11ea-9601-ccb147c9a101.png width=300/>

The div with class slot-container has a height of 0.
<img src=https://user-images.githubusercontent.com/44481292/88345334-9dd39180-ccfa-11ea-967e-137c1c934fae.png width=400/>

If I use the debugging tools and add a height of 100% it looks fine.
<img src=https://user-images.githubusercontent.com/44481292/88345407-be9be700-ccfa-11ea-9338-f7a38741985b.png width=400/>

<img src=https://user-images.githubusercontent.com/44481292/88345436-cb203f80-ccfa-11ea-857d-d25509e4ff14.png width=300/>

cc @zaramatheson

### Reproduction Steps

1.

asangma:
Cool cool. I got it.

asangma:
@AdelheidF Lemme know if you can verify.

AdelheidF:
Verified locally using beta.0. Waiting for a release for devext.

asangma:
Thanks @AdelheidF !

Issue #1000, opened 2020-05-27
Bug: calcite-fab - issue with z-index


AdelheidF:
### Summary

z-indexed content inside components can conflict with things like the sticky fab.
Find a generic solution as we'll probably see other widgets that utilize z-index, e.g. a slider with multiple handles.

### Actual Behavior
<img src=https://user-images.githubusercontent.com/44481292/82962442-14f7de00-9f75-11ea-88d4-1ada0796806d.png width=200/>

### Expected Behavior
No such conflicts.

### Reproduction Steps

1. add a z-indexed thing to a panel
1. add a calcite-fab in the panel's fab slot.
1. panel should have enough content above the z-indexed thing to make the panel scroll and to place the z-indexed thing under the fab when it's acting sticky.
1. make it so the z-indexed thing is under the fab slot

asangma:
@jcfranco @driskull @kat10140 Would y'all agree that setting the z-index to 5 is a good approach?

driskull:
Can we try setting `position:relative` on the container for the slot and a zindex if needed? We don't want to have to compete with whatever they put in the slot. We should just have the fab always be above it.

AdelheidF:
The container of the fab is a flow-item. If I set position:relative and z-index doesn't change anything.

http://devext.arcgis.com/apps/mapviewer/index.html?url=https://services1.arcgis.com/4yjifSiIG17X0gW4/arcgis/rest/services/FatalAccidents2017/FeatureServer/0
Click on Filter and add 2-3 expressions.

driskull:
@AdelheidF I'm talking about setting the position of the fab container within the component. We should try setting position relative on that and handling any zIndex issues so we don't have to worry about user zIndexes.

asangma:
@driskull Hmm...we're using `position: sticky` for the fab slot.

driskull:
yeah, just make sure that the container it's inside has a position relative

driskull:
We handled this same scenario in the scrim component

asangma:
Got it.

asangma:
Applying position: relative to various things is maybe not working. For now though, Jonathan will be removing z-index from the slider handle for when there's only one handle. This will take care of the specific problem, but we may see other z-index issues in the future. 

@AdelheidF I'm going to edit the issue to be general to any z-indexed thing. Cool?

driskull:
@asangma adding this to the container around the slot should fix it.

[...]

PR #1048, opened 2020-07-31
fix(panel): remove slot-container and associated styles

asangma:
Should fix this:
https://github.com/Esri/calcite-components/issues/797

Should not cause this again:
https://github.com/Esri/calcite-app-components/issues/1000

Should not cause this again:
https://github.com/Esri/calcite-app-components/issues/1044

@driskull tested this with a stand-in histogram slider and with a div with similar styling. Neither applied the z-index fix, and it seemed to be working.

Hoping that @AdelheidF can test this using the next.

### DIFF
--- a/src/components/calcite-panel/calcite-panel.scss
+++ b/src/components/calcite-panel/calcite-panel.scss
@@ -124,9 +124,3 @@ calcite-scrim {
   left: 0;
   right: 0;
 }
-
-.slot-container {
-  height: 100%;
-  position: relative;
-  z-index: 0;
-}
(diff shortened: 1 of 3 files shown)


asangma:
Want me to make sure this doesn't explode things when Adelheid tests it before adding it to calcite-components?

asangma:
Adelheid tested this in her test apps and it looks like this handles all the things:

* fab v slider z-index appears to be fixed
* fab position in panel is fixed
* table not showing up.

According to Adelheid, generally if it's good in her test apps it's good in MVB.

I will add this to the porting branch in calcite-components.

PR #1060, opened 2020-08-26
feat(pick-list-group, pick-list-item, tip-group, tip-manager, value-list-item): add replacement text props

jcfranco:
**Related Issue:** N/A

## Summary

Deprecates `text<Name>` props in favor of `<name>` for consistency. 

~The only remaining one is `textFilterPlaceholder`. Not sure if this should be `filterPlaceholder` or `intlFilterPlaceholder` instead.~ Went with `filterPlaceholder`.


driskull (approved):
Looks good!

eriklharper (approved):
👍

